Issue #4, opened 2015-12-06
Incorrect link on the community page


mscherer:
On https://beta-ovirtweb.rhcloud.com/community/about/community-guidelines/ 

there is some example link that are not transformed to html a tag by middleman. See "To search mailing lists"?

garrett:
I guess it's being thrown because of the `:` in the URL? Technically, it is supposed to be encoded as `%3A`... but people shouldn't have to worry about that.

So this Markdown:

```markdown
[example](https://www.google.fr/search?q=site:lists.ovirt.org+Engine+plugin)
```

Would be come this, with URL encoded string:

```markdown
[example](https://www.google.fr/search?q=site%3Alists.ovirt.org%20Engine%20plugin)
```

But, IMHO, Markdown and browsers _should_ handle things like that transparently.

garrett:
Also, isn't it a bit odd that we're asking people to search the various resources using Google site searches like that (as Bugzilla does have a search engine)? And, there's no wiki, so that part should be removed... the site has its own search too now.

It's also bad that we're asking people to search on a different site for multiple different oVirt resources. If we were to have people search offsite for all the various resources at once, it should be prefixed with something like this, perhaps:

```
site:lists.ovirt.org OR site:ovirt.org OR (site:bugzilla.redhat.com ovirt)
```

Of course, that's a little redundant with ovirt.org being in there, with the site's own searching.

garrett:
The above PR merge doesn't fix the issue with URLs with colons, but it does make that page better and replaces the links. (Basically, this was solved by rethinking the block instead of fixing the technical bug.)

thatdocslady:
@garrett so what are we doing to simplify the search function? If I have anything to say about it, I vote to keep the number of search platforms as low as possible, but I don't know how you can use the ovirt site to search in bugzilla =/

garrett:
The search function for the website only searches the website. There's no possible way to have it search anything else — it has to build an index, and it isn't going to download mailing list archives and bugzilla.

--

As far as the search form on that page, it searches both the mailing lists and bugzilla by default, but one can choose something else. I could extend that to search the website via Google too, but then it would get really noisy hits and nobody would see the issues outline in the mailing lists or bugzilla. (I've tried this.)

You can see it in action @ https://beta-ovirtweb.rhcloud.com/community/about/community-guidelines/

The embedded search form on that page, defaults search both mailing lists and bugzilla, but you can choose one or the other.

--

[...]

PR #7, opened 2015-12-07
search lists & bugzilla from suggestions page

garrett:
As Markdown supports some embedded HTML, I've added a form that kicks a query over to Google as previously only suggested in plaintext.

I dug around and figured out how to append the limiting text to a provided search string during a Google query so that one could simply type a simple string to search. In addition, I have this appended instead of prepended so that the search is always visible within Google as well.

### DIFF
--- a/source/community/about/community-guidelines.html.md
+++ b/source/community/about/community-guidelines.html.md
@@ -23,9 +23,23 @@ We want to keep the oVirt Community a great place to participate, but we need yo
 
 A few seconds on Google can help you find an answer to a question, or a past discussion on a subject, which can help us reduce duplication of effort and any associated frustration. Your effort will be appreciated.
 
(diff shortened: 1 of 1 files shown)


garrett:
This change is intended to address issue #4.

garrett:
While submitting this pull request, I just thought of the possibility of having a dropdown label that defaults to choose between searching "lists & bugzilla", just "lists", or just "bugzilla".

Looking into this now. (If it's quick and easy to implement, I'll do a force push to this feature branch so that there'll be a single commit in the PR with the change.)

garrett:
Okay; I've added a dropdown (and tested a bunch of stuff along the way, like search order, and if it makes sense to have site search in there too, or if it just adds noise — and it just adds noise, sadly).

PR #53, opened 2016-02-10
add redhat-rpm-config as dependency.

rafaelmartins:
this is needed at least for clean fedora 23 installs.

Signed-off-by: Rafael Martins <rmartins@redhat.com>


sandrobonazzola:
+1

PR #71, opened 2016-02-19
Fix release notes link

ekohl:


thatdocslady:
@garrett how do we handle updates to haml/yaml files? Is there anything extra we should run/check before we merge? Normally for md files the risk is low, but not sure how to ensure safe merge in this case.

garrett:
You should check it locally, to ensure the formatting looks correct.

However, Travis checks that the site builds. If there's an actual syntax error in a HAML or YAML file, the build will fail and the checks will _not_ pass. Therefore, always make sure all changes to any non-Markdown file go through the pull request process, so that the site won't break*.

_\* And this is the literal form of break. Wonkiness or incorrectness can still show up. There's no way to prevent **that** in a machine automated way. At least the PR + CI process will check for actual breakage. :smiley:_

---

**TL;DR**: Pull requests call Travis CI, which ensures the site builds (but it doesn't check the content).

thatdocslady:
@ekohl plz see above comment, can you confirm that you ran this fix locally and that it works? Unfortunately we don't have pre-merge staging to view the site output yet...

ekohl:
@thatdocslady I have now, but I simply used the edit in github as linked at the bottom of the page. I recommend you add a CONTRIBUTING.md to the repository so users who submit a pull request know they should manually verify it.

thatdocslady:
@ekohl thanks for the suggestion, will add contribution guidelines link :)

thatdocslady:
@ekohl can you please check if the new contribution guidelines file appears when you try to open a new pull request? I can't seem to find it and I have a feeling it's because I have commit rights ;)

ekohl:
@thatdocslady it does. Even though you have commit rights, you should be able to open a PR.
pr-screenshot

PR #74, opened 2016-02-20
Update ovirtappliance.html.md

mikedep333:
I would link to the "Current status" section of ovirtappliance.html.md, but it doesn't look like I can do that.


garrett:
You can link to `#current-status`, and it would jump to the heading of the same name on the current page. (If it's from another page, you can just add the URL before the `#`.)

thatdocslady:
@garrett I'm going to merge this, could you please help with linking like you said and submit a separate PR? Thx!

PR #79, opened 2016-02-21
fixing category for main infra doc page

eedri:


eedri:
fixing infra category

PR #80, opened 2016-02-21
Moving IPv6 page from node folder as it isn't node related.

yevgenyz:


garrett:
Shouldn't this be moved to `/develop/release-management/features/network/ipv6-support.html.md` ?

thatdocslady:
@garrett @sandrobonazzola - we should look into how feature pages are organized, i.e. do we want them in version-specific folders, in folders according to project/product area (like Garrett suggested), or something else? Sandro, this is mainly your call because release manager :)

sandrobonazzola:
@thatdocslady I think it's better organized by project/product area like @garrett suggested, but I'd like to keep a reference (category?) to the version introducing the feature.

yevgenyz:
Could we have tags on the pages, so they could be found by any of their different aspects and those could be updated (like adding a version once we know that). If we do that their actual folder isn't that important.

thatdocslady:
@garrett what did we end up deciding wrt categories/tags? is there any way to utilize them for version numbers, or maybe @sandrobonazzola should add a "proposed in version" section to the feature template instead?

yevgenyz:
@garrett the pull request was updated, so the file is in /develop/release-management/features/network as you proposed.

garrett:
Catagories/tags/version could exist in the Frontmatter metadata at the top of each file. There would have to be code written to utilize it, but you can standardize on any field first too.

garrett:
Every feature page is rendered using this layout: https://github.com/oVirt/ovirt-site/blob/master/source/layouts/feature.html.haml (as specified in [config.rb](https://github.com/oVirt/ovirt-site/blob/master/config.rb#L122-L124)). If you want to expose the to-be-added metadata from the files, edit the feature layout.

You can get to the data by accessing `current_page.data`. In the layout page, I made a shortcut called `pdata` ("page data") for part of the YAML. You can [see it in action in the layout file](https://github.com/oVirt/ovirt-site/blob/master/source/layouts/feature.html.haml#L163-L203).

garrett:
BTW: Naturally, when editing that file, you'll most definitely want to run the server locally. Do not change the HAML files via GitHub's editor — this is _especially_ true for the layout files. (Simple word changes in the `index.html.haml` file are okay in other words, but that's more an exception.)

dankenigsberg:
@garrett, is there an action item on @yevgenyz here? If so, please elaborate.

PR #105, opened 2016-02-28
Moved feature template page to the right location

mykaul:


sandrobonazzola:
looks ok to me. @thatdocslady any other template around in different position?

thatdocslady:
@sandrobonazzola I don't know, will have to search for it. I'll open an issue to do that so that we don't forget :)

PR #110, opened 2016-03-01
Fixed what's new formatting for 3.6.3 release notes

sandrobonazzola:


mykaul:
There are no bullets for the new features - do we want to add them?

sandrobonazzola:
I've no strong feelings about it. We never used bullets there but if you think it's better I'm fine adding them.

sandrobonazzola:
We can add bullets in a new patch, merging this for now.

PR #113, opened 2016-03-01
Update recommended browsers

gregsheremeta:
Update recommended browsers to remove Chrome, add "ESR" to Firefox.

### DIFF
--- a/source/download/index.html.haml
+++ b/source/download/index.html.haml
@@ -37,8 +37,7 @@ wiki_last_updated: 2015-01-21
 
           ### Recommended browsers
 
-          -   Latest Mozilla Firefox
-          -   Latest Google Chrome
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
+1

mykaul:
@gregsheremeta  - why did we decide to remove Chrome? Is it much worse than Firefox? Since we cannot recommend a version?

Do we want to provide a link to ESR (https://www.mozilla.org/en-US/firefox/organizations/) page?

gregsheremeta:
@mykaul Chrome has never been officially supported. We're fixing a BZ by removing its mention.

Re: providing a link to ESR, that's fine with me but would be up to Oved or Yaniv Dary or you.

mykaul:
@gregsheremeta  - I'd add the link to what ESR is - not many know.

Issue #99, opened 2016-02-24
Previous site pages are unreachable


sandrobonazzola:
Even if from the new website the previous website page are referenced, they're not reachable anymore so if something gone wrong with the conversion you can't reach the correct one to compare with.

thatdocslady:
Can you provide some examples?

sandrobonazzola:
http://www.ovirt.org/download/ovirt-live/ has a link to http://ovirt.org/OVirt_Live which just redirect to http://www.ovirt.org/download/ovirt-live/ so you can't check if something is missing.

Same for http://www.ovirt.org/release/3.6.3/ which has a link to http://ovirt.org/OVirt_3.6.3_Release_Notes which is Not found :(

thatdocslady:
@garrett @jasonbrooks can one of you help here?

jasonbrooks:
It looks like all those old wiki page links need to get `old.` appended. So, `http://ovirt.org/OVirt_Live` should become `http://old.ovirt.org/OVirt_Live`, an so on...

sandrobonazzola:
Thanks, fixed.

PR #119, opened 2016-03-02
prepend old. to original wiki links

jasonbrooks:
to address issue #99


sandrobonazzola:
+1, @thatdocslady can you review as well and merge?

PR #120, opened 2016-03-03
Update Image Upload sequence diagram

gpadgett:
Prior diagram was lost in translation from the old wiki; restore it with
the latest version.


gpadgett:
Ping?
(For review/verification, the original page with more recent image is at http://old.ovirt.org/Features/Image_Upload; the image in question is http://old.ovirt.org/File:Image_Upload-HTTP_Sequence_Diagram.png)

PR #130, opened 2016-03-10
tracker JS uses the same resource now (http/https)

### LINKED ISSUE #87: Mixed non-ssl resources on the ssl page
Hi,
So after enabling the correct cert, when loading the ssl-enabled page firefox complains about a couple resources not being fetched from secure connections, the resources are:

   http://www.redhat.com/j/elqNow/elqCfg.js
   http://www.redhat.com/j/elqNow/elqImg.js

The https counterparts already work, so probably it's just a matter of changing a hardcoded url somewhere

garrett:
This should fix #87.

### DIFF
--- a/source/layouts/_production_only.haml
+++ b/source/layouts/_production_only.haml
@@ -1,5 +1,5 @@
 -# Red Hat tracking
 
 / eloqua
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
Thanks

PR #133, opened 2016-03-15
3.6.z release management - Fix broken links

sandrobonazzola:


mykaul:
Still getting 404 on some.

http://www.ovirt.org/develop/release-management/releases/3.6/release-management/ did not survive well the migration from Wiki format.

sandrobonazzola:
The one with 404 points to page that have been lost in the migration, or at least I wasn't able to find them in the new website. I'll try to get them from the old site and re-create them in different patches.

PR #145, opened 2016-03-21
fix some numbering restarts

rstory:
non-indented text in the middle of a numbered list resets numbering. this
patch fixes some easy instances, but trickier ones remain (e.g. all the ones with 'examples'.)


mykaul:
I think the change is good, but frankly - the getting started is quite outdated - talking about Fedora 19 and releases 3.4 and 3.5. I wonder if we should just re-write it?
@thatdocslady  - thoughts?

sandrobonazzola:
+1 for the re-writing

sandrobonazzola:
while we wait for a re-writing of the guide, accepting the patch fixing the numbering.

PR #169, opened 2016-04-07
localinstall is deprecated in yum

fbacchella:
In yum's man page in RHEL6 (so quite old release):

       localinstall
              Is used to install a set of local rpm  files.  If  required  the
              enabled  repositories will be used to resolve dependencies. Note
              that the install command will do a local  install,  if  given  a
              filename. This option is maintained for legacy reasons only.


sandrobonazzola:
Can you please rebase this pull request?

sandrobonazzola:
:+1:

PR #171, opened 2016-04-08
Remove Spice Plugin, since it is unsupported since 4.0

mareklibra:
Blocks https://bugzilla.redhat.com/show_bug.cgi?id=1316560


sandrobonazzola:
:-1: oVirt 4.0 has not been yet released and 3.6 (current supported version) still support spice xpi plugins. I think that the change here should be a deprecation warning for 3.6 telling user to try to avoid its usage in 3.6 since in 4.0 it won't be available.

sandrobonazzola:
:+1:

PR #172, opened 2016-04-10
Update quickstart with latest versions.

xrobau:
This simply adds 3.6 to the quickstart documentation.

### DIFF
--- a/source/documentation/quickstart/quickstart-guide.html.md
+++ b/source/documentation/quickstart/quickstart-guide.html.md
@@ -34,6 +34,7 @@ The following requirements are typical for small- to medium-sized installations.
     -   [oVirt 3.3](http://resources.ovirt.org/pub/yum-repo/ovirt-release33.rpm)
     -   [oVirt 3.4](http://resources.ovirt.org/pub/yum-repo/ovirt-release34.rpm).
     -   [oVirt 3.5](http://resources.ovirt.org/pub/yum-repo/ovirt-release35.rpm).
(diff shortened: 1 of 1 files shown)


source/documentation/quickstart/quickstart-guide.html.md
```diff
@@ -34,6 +34,7 @@ The following requirements are typical for small- to medium-sized installations.
     -   [oVirt 3.3](http://resources.ovirt.org/pub/yum-repo/ovirt-release33.rpm)
     -   [oVirt 3.4](http://resources.ovirt.org/pub/yum-repo/ovirt-release34.rpm).
     -   [oVirt 3.5](http://resources.ovirt.org/pub/yum-repo/ovirt-release35.rpm).
+    -   [oVirt 3.6](http://resources.ovirt.org/pub/yum-repo/ovirt-release36.rpm).
```
mykaul: Why is this link different than the links below (the 'http://plain...' ones)?

mykaul:
Personally, I'd remove all but 3.6 (possibly we can leave 3.5). There are also outdated references there (to Fedora 19, for example).

xrobau:
I agree, I think it needs a good cleanup, but I don't think *I'm* qualified enough to do so, as I've been using it for less than 24 hours. Those were just some obvious mistakes that needed fixing.

sandrobonazzola:
:+1: thanks for the patch. I agree with @mykaul that the whole document needs a massive rewrite, but I think that this patch is ok.

PR #194, opened 2016-04-21
Update InClusterUpgrade feature documentation

rmohr:


source/feature/inclusterupgrade.html.md
```diff
@@ -3,10 +3,10 @@ title: InClusterUpgrade
 authors: rmohr
 wiki_title: Features/InClusterUpgrade
 wiki_revision_count: 47
-wiki_last_updated: 2016-01-12
+wiki_last_updated: 2016-04-21
 feature_name: In Cluster Upgrade
 feature_modules: engine
-feature_status: Development
+feature_status: Completed
```
sandrobonazzola: Missing the version including the complete fix.

source/feature/inclusterupgrade.html.md
```diff
@@ -75,10 +75,6 @@ The weight policy unit gives an OS which is newer than the OS where the VM is cu
 *   No load balancing will happen
 *   Migrating a VM back to an older OS version is not possible during upgrade. To run a VM on an older host OS, the VM hast to be stopped first.
 
-#### What might stop you from putting a host to maintenance
-
-*   Missing networks
```
sandrobonazzola: Does it now works even if networks are missing?

sandrobonazzola:
Overall looks ok to me, just a couple of inline comments.

source/feature/inclusterupgrade.html.md
rmohr: Done

source/feature/inclusterupgrade.html.md
rmohr: There was just an idea to collect reasons.
There are many reasons why a host can't be put to maintenance and it has nothing to do with this feature. I decided that it is counterproductive to have that here because it will just be outdated soon. Maybe we have a useful link where I can point to.

rmohr:
Updated

PR #204, opened 2016-04-25
Add vi[m] swap files

mureinik:
Added vi[m]'s temporary swap files to .gitignore to make editing the
site wit vi[m] easier.

Change-Id: I1e1edc8f6b07570989b63e0f323403f8bab16754
Signed-off-by: Allon Mureinik <amureini@redhat.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@
 # Ignore emacs .. erm its backup files..
 *~
 
+# Ignore vi[m] swap files
+*.swp
+
 # local Docker image related files
 /.dockerid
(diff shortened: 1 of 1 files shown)


mykaul:
@mureinik https://github.com/github/gitignore/blob/master/Global/Vim.gitignore has more suggestions.

mureinik:
@mykaul None of the things they add there present in this project. I see no point in cluttering up the `.gitignore` file with stuff we don't need. If we ever encounter one of these, we can always add it later.

PR #218, opened 2016-04-28
added doc about migrating ovirt-engine from 3.6 to 4.0

rafaelmartins:
Simple documentation about how to migrate ovirt-engine from 3.6 to 4.0.

Don't know if this is the better location for this documentation, so feel free to move it around, if there's a better place for it.

Also, we need to make sure to update ovirt-release rpm after release :-)


sandrobonazzola:
I think you should explain why we're talking about migration.

3.6 support FC22 and EL6 engines which are not supported anymore on 4.0.
So we need to allow people using those OS to migrate to different OS in order to upgrade to 4.0.

rafaelmartins:
Done

sandrobonazzola:
:+1:

PR #220, opened 2016-04-28
added instructions on how to test locally

### LINKED ISSUE #186: Local test requested before pull, but no instructions and both server and docker scripts fail
Hi,

The CONTRIBUTING file asks for local testing of changes in the "Test your changes locally" section. But there are no instructions about how to actually do that.

I tried using the standard approach of bundle install && bundle exec ./run-server.sh, but it fails with:

```
/home/msivak/.gem/ruby/gems/middleman-core-3.3.10/lib/middleman-core/core_extensions/data.rb:167:in `method_missing': undefined method `events' for #<Middleman::CoreExtensions::Data::DataStore:0x00000004b30160> (NoMethodError)
	from /home/msivak/Work/ovirt-site/config.rb:138:in `block in initialize'
```

I also tried to use the ./docker-setup.sh && ./docker-run.sh script, but that also failed:

```
[msivak@dhcp131-31 ovirt-site]$ sudo ./docker-run.sh 
Running docker image springboard-20160420-10023
An error occurred while installing nokogiri (1.6.6.2), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.6.6.2'` succeeds before bundling.
```

So apart from having no instructions, the "obvious" ways do not work either.

gregsheremeta:
fixes #186

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,13 @@ the site locally and test your changes. We do have plans to add pre-merge stagin
 until this happens please follow this step and add a confirmation in the pull request description that the local
 test ran successfully.
 
+Run:
+```
+./setup.sh && ./run-server.sh
+```
+You should get:
+The Middleman is standing watch at http://0.0.0.0:4567
+
 ## Submit your changes
 
 when you finish creating or editing content, commit your changes to a **new branch** and submit a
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
:+1:

PR #236, opened 2016-05-10
Update gluster-self-heal-monitoring.html.md

dpmramesh:
Updating the self-heal monitoring feature page with more info about the design of host fencing and host maintenance flows


source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -41,21 +41,51 @@ Host Maintenance and Fencing will be enabled/disabled based on the following cri
...
+          if (quorum is maintained (bricks and nodes for ALL volumes) with this host down)
+            allow host fencing
+          else
+            skip host fencing
+
+###MaintenanceNumberOfVdssCommand (for Host Maintenance)
+  MaintenanceNumberOfVdssCommand is used for moving one or more number of hosts to maintenance. It will be enhanced to consider gluster self-heal and quorum before moving the host to maintenance. Following validations will be added to MaintenanceNumberOfVdssCommand.validate() method.
+  
+     if (host supports both gluster & virt services & not force-maintenance)
+      if(all bricks are down)
```
sabose: The logic is repeated - can it be extracted to a different class?

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -41,21 +41,51 @@ Host Maintenance and Fencing will be enabled/disabled based on the following cri
...
+ More info about standard host fencing is available at http://old.ovirt.org/Automatic_Fencing#Automatic_Fencing http://old.ovirt.org/Fence_kdump https://www.youtube.com/watch?v=V1JQtmdleaM
+
 ## Entity Changes
 Following entities will be changed as part of Gluster self heal monitoring.
 
 ### Gluster_volume_bricks
   Following columns will be added to gluster_volume_bricks table.
   unsynced_entries  - integer - No.of unsynced entries in the brick.
   unsynced_entires_history - text - History of unsynced entries in the brick. It will a list of comma separated values.
-
```
sabose: Can you also add the new option that will control the heal monitoring frequency

sabose:
@oourfali @mwperina - could you review changes?

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
dpmramesh: will add that

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
dpmramesh: We can have validator which will be used in all the places.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -41,21 +41,51 @@ Host Maintenance and Fencing will be enabled/disabled based on the following cri
...
+        allow host fencing
+      else(some || all bricks are up)
+        if (unsync'd entries  > 0 for any up brick) //Entris needs to be healed from this host
+          skip host fencing
+        else
+          if (quorum is maintained (bricks and nodes for ALL volumes) with this host down)
+            allow host fencing
+          else
+            skip host fencing
+
```
mwperina: Concept looks good, but I have several questions/notes:
1. All of the checks mentioned above are performed using only informations contained in engine db, right? There won't be any remote call to host to get this informations, right?
2. Those gluster related checks should be performed after "fence disable by cluster policy" and "connectivity broken threshold reached". Do you agree?
3. And the most important thing is timing issue: you stated that healing informations are refreshed only each 10 minutes, which is a quite long period and lots of things can happen during that time. How will we handle following scenario:
  a. Host-A is used as a source for healing Host-B and this information is contained in engine db
  b. After a while Host-A crashes due to some kernel or even hardware issue
  c. Engine will start NonRespondingTreatment, but fencing will be skipped due to no-longer-relevant healing informations in engine db -> Host-A will stay in NonResponsive status forever -> HA VM cannot be migrated until host is fenced (which will never happed due to above) or until administator will not mark host as rebooted manually
4. If we add those gluster logic into shouldFencingBeSkipped(), SSH Soft Fencing won't be executed, although during the call we agreed that SSH Soft Fencing cannot harm gluster heal process. So, wouldn't it be better to create different method (for example shouldPmRestartBeSkipped) which would be called prior to PM restart execution?

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -41,21 +41,51 @@ Host Maintenance and Fencing will be enabled/disabled based on the following cri
...
+###MaintenanceNumberOfVdssCommand (for Host Maintenance)
+  MaintenanceNumberOfVdssCommand is used for moving one or more number of hosts to maintenance. It will be enhanced to consider gluster self-heal and quorum before moving the host to maintenance. Following validations will be added to MaintenanceNumberOfVdssCommand.validate() method.
+  
+     if (host supports both gluster & virt services & not force-maintenance)
```
mwperina: Hmm, this is less harmful than NotRespondingTreatment, but here we have even worse timing issue, because migrating all VMs from the host can take a huge amount of time:

1. If we execute those checks in validate, host may not be use for healing at the moment, but as VMs migration from the hosts may take long time, healing may start during this migration and we have an issue.
2. On the other hand host may be used for healing, but healing may be finished before VMs migration, so we disable move to maintenance operation unnecessarily.

Due to above, wouldn't it be better following approach:
1. Before execution of MaintenanceNumberOfVdssCommand check for healing status and if healing is in progress display warning to user: "Healing process is performed so move to maintenance may not be successfull. Do you really want to continue? Yes/No"
2. If user chooses Yes, than execute MaintenanceNumberOfVdssCommand, but repeat gluster healing check after all VMs migrations are finished and prior to changing host status to Maintenance. And if healing is still runing, just fail the command with error "Host cannot be moved to Maintenance, because gluster healing process is currenlty performed".

What do you think?

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
oourfali: Adding also Moti

oourfali:
@masayag please review as well.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
oourfali: As for "1": Why not just to block in case healing is in progress? Even if it will end, the user can retry later on.
As for "2": Makes sense to do a double check I guess.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
oourfali: As far as I remember we agreed to add new fencing policies for each of those checks (quorum + healing).
I think it would be cleaner.
Thoughts?

dpmramesh: @oourfali. I can do that.  All current policies are available in UI and it can be customized at cluster leve. Do we need to expose these two new policies to UI?

dpmramesh: @mwperina 

For 1, Yes. All checks will be performed with the information in DB. There won't be any VDSM call to hosts.

For 2,  Yes. This will be added as a lost check after all other checks. 

For 3, This is the most critical point. Either we have to check the host(gluster host) to get the latest value, or wait for max 10 minutes. 'gluster volume heal info' adds some overhead on the gluster processes. So I don't think we can increase the frequency further. If we know for sure that the bricks are down then we can allow the Power Management. I see two solutions for this.

1. Run the 'gluster volume heal info' command to get the latest values before fencing.
2. Delay the power management for 10 mins. 

@sabose  What would you suggest?. Do you see any other option here?

For 4, Got it. SSH Soft fencing should be allowed. We will add a new fencing policy for self heal and quorum and it will be checked before PM Restart.

oourfali: two options:
1. expose, and have those disabled by default. and show those only in case of virt+gluster use-case.
2. Have them in the backend, but don't expose them.

I'd go with #1.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
dpmramesh: @mwperina 

For 1, Definitely we want to provide an option in the UI to force/override these checks. So we will check and warn the user but he can either decide to go ahead with maintenance or skip. 

For 2, Why do we need double check if the user takes the informed decision after the warning?.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
mwperina: > For 3, This is the most critical point. Either we have to check the host(gluster host) to get the latest value, or wait for max 
> 10 minutes. 'gluster volume heal info' adds some overhead on the gluster processes. So I don't think we can increase 
> the frequency further. If we know for sure that the bricks are down then we can allow the Power Management. I see two 
> solutions for this.
>
>    Run the 'gluster volume heal info' command to get the latest values before fencing.

Well, checking healing info on fence proxy host same as we currently check sanlock lease would be the most reliable way (assuming that getting volume healing info is not too time consuming). Also the option to check or not healing info can be passed from engine to VDSM as a fencing policy option same as in sanlock lease case.

>   Delay the power management for 10 mins.

I wouldn't go that way, because with 10 minutes delay we will break HA VMs contract (HA VMs cannot be restarted on different host until PM stop was successfully executed on host)

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
mwperina: > For 2, Why do we need double check if the user takes the informed decision after the warning?. 

Well, that depends on the warning message. If the message is "Host is part of volume healing. If the healing process will not finish before all VMs are migrated from the host, host cannot be switched to Maintenance. Do you really want to continue?"

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
sabose: While determining if a host has to be fenced, it is safe to say that host can be fenced if the brick processes on the host are not running - so I think we can add this to fencing policy option. At the gluster side, we're working on a quicker way to determine if a brick is a source for heal (so that we can query this from vdsm). Till this is available - let's add the "is brick process running check" to fencing policy option. @mwperina can you point to how sanlock lease case is handled.

mwperina: Sure, here is the whole flow:

1. Please take a look at org.ovirt.engine.core.common.businessentities.FencingPolicy, for sanlock lease there's skipFencingIfSDActive attribute (all FencingPolicy attributes are stored in clusters table)
2. Some attributes of FencingPolicy are checked in VdsNotRespondingTreatmentCommand.shouldFencingBeSkipped(), but skipFencingIfSDActive is passed down the stack
3. FencingPolicy instance is passed through NotRespondingTreatmentCommand -> RestartVdsCommand -> FenceActionExecutor -> FenceAgentExecutor where FencingPolicy is converted in convertFencingPolicy() into a Map which is passed to fenceNode VDSM verb
4. On VDSM please take a look at fenceNode in API.py, where in should_fence method we check sanlock lease status

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -29,43 +29,48 @@ Gluster Self Heal helps to heal data on the gluster bricks when there are some i
...
 ## Entity Changes
 Following entities will be changed as part of Gluster self heal monitoring.
 
 ### Gluster_volume_bricks
-  Following columns will be added to gluster_volume_bricks table.
+Following columns will be added to gluster_volume_bricks table.
+
   unsynced_entries  - integer - No.of unsynced entries in the brick.
+
   unsynced_entires_history - text - History of unsynced entries in the brick. It will a list of comma separated values.
```
sabose: Please mention the number of entries that will be stored in the history column

sabose:
I think changes look good.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -27,7 +27,7 @@ Gluster Self Heal helps to heal data on the gluster bricks when there are some i
 *   Status: In development.
 
 ## Monitoring Self Heal
-We will monitor self heal status at the volume level. ‘gluster volume heal VOLNAME info’ command will be used to get the heal info. If there are some unsynced entries in the volume then volume will be marked as ‘Needs Healing’ and bricks which are having unsynced entries will be marked accordingly. Ovrit will sync self heal info for all the gluster volumes with frequency of 10 minutes.
+We will monitor self heal status at the volume level. ‘gluster volume heal VOLNAME info’ command will be used to get the heal info. If there are some unsynced entries in the volume then volume will be marked as ‘Needs Healing’ and bricks which are having unsynced entries will be marked accordingly. Ovirt will sync self heal info for all the gluster volumes with frequency of 10 minutes. Sync frequency is configurable using VDC Option 'GlusterRefreshRateHealInfo'.
```
mwperina: I'd change following:

1. OVirt -> oVirt

2. Synchronization frequency can be changed using engine-config option 'GlusterRefreshRateHealInfo'.

dpmramesh: Done.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
```diff
@@ -37,7 +37,7 @@ Following columns will be added to gluster_volume_bricks table.
 
   unsynced_entries  - integer - No.of unsynced entries in the brick.
 
-  unsynced_entires_history - text - History of unsynced entries in the brick. It will a list of comma separated values.
+  unsynced_entries_history - text - History of unsynced entries in the brick. It will be a list of comma separated values. By default, last 40 entries will be stored in the field and it is configurable using VDC Option 'GlusterUnSyncedEntriesHistoryLimit'
```
mwperina: I'd change alter a bit:

By default, last 40 entries will be stored in the field. The limit can be changed using engine-config option 'GlusterUnSyncedEntriesHistoryLimit'.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
mwperina: Please don't forget to change also OVirt to oVirt.

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
dpmramesh: Done

source/develop/release-management/features/gluster-self-heal-monitoring.html.md
dpmramesh: Done

mwperina:
Looks good to me, thanks. :+1:

dpmramesh:
@mykaul Can you merge this pull request ?

PR #260, opened 2016-05-25
Add PyYAML and openvswitch as dependency for vdsm build

dpmramesh:
VDSM build requires two new packages PyYAML openvswitch. So including these packages in the vdsm developer guide.


dpmramesh:
@sandrobonazzola Can you review this pull request?

PR #268, opened 2016-06-01
Try to fix Markdown and other errors to make Travis happy.

mykaul:
Two issues, hopefully now Travis will not complain.


mykaul:
It fixes one issue (the one on the 3.5.2), but not the core issue.

thatdocslady:
@jasonbrooks @garrett plz help :(

jasonbrooks:
@thatdocslady @mykaul I poked around and found that it's missing frontmatter that's causing the problem. The bare minimum required is an authors field, though it seems like title would be nice to include, too. You could check out other feature pages to see their conventions:

```
---
title: Deploy Hosted Engine Hosts via Engine
authors: mykaul
---

```

PR #279, opened 2016-06-07
Install openssl for gem updates

mureinik:
Commit 4b49f9 changed the Gemfile to download files from
https://rubygems.org instead of http://rubygems.org. Using HTTPS
requires openssl to be installed, and while it's hard to imagine a
system that doesn't have it installed, we should not make any
assumptions. Especially in light of how cheap it is to verify this.

Change-Id: I90e4b25eade1c7a962014554bd76e6d43e5fd68c
Signed-off-by: Allon Mureinik <amureini@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV UID 1000
 ENV GID 1000
 
 # Install necessary deps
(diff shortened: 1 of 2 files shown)


mureinik:
@garrett can you review and keep me honest here please?

mykaul:
:+1:

PR #284, opened 2016-06-09
fail more gracefully when attempting to make a pretty_date

garrett:
Just in case the date object or string passed to `pretty_date` doesn't parse, fail gracefully and return the original object (or string).

### DIFF
--- a/lib/site_helpers.rb
+++ b/lib/site_helpers.rb
@@ -15,12 +15,16 @@ class SiteHelpers < Middleman::Extension
     def pretty_date(sometime, length = 'long')
       return unless sometime
 
-      sometime = Time.parse(sometime) if sometime.class == String
+      begin
+        sometime = Time.parse(sometime) if sometime.class == String
 
-      format = length == 'short' ? '%a %e %b' : '%A %e %B'
-      format << ' %Y' unless sometime.year == Time.now.year
+        format = length == 'short' ? '%a %e %b' : '%A %e %B'
+        format << ' %Y' unless sometime.year == Time.now.year
 
-      sometime.to_time.strftime(format) rescue ''
+        sometime.to_time.strftime(format)
+      rescue
+        sometime
+      end
     end
 
     def word_unwrap(content)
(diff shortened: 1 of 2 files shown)


duck-rh:
Is there any reason to accept a broken date?

I though there was a script to check the YAML data for the events. Also the admin interface should prevent these kind of errors (or be patched).

I could help on this, but I see no reason to accept broken data which would give a broken output without warning us to correct it.

garrett:
This doesn't have to do with the events; it's for everything that uses a date, including the blog posts.

Backstory: When running a final build test for PR #285 (wiki link redirects), after merging in master, I had a page causing a failure when compiling the search index, so I added this failsafe to ensure it compiles.

garrett:
In other words, people may type the wrong date — even if there's a timezone issue. Since this function is only used for string formatting a date, it's better to just show a slightly broken date (like CEST versus CET or whatever) instead of causing the build to fail completely.

garrett:
Ok, I ran another vanilla build from master without this patch or the wiki redirects ... and I guess someone wrote the wrong date somewhere and merged it in regardless of the PR CI passing. :disappointed:

Yet another bug to check down, with invalid content. At least this commit makes it so it won't break the build.

duck-rh:
well, if it is CEST instead of CET it might seem minor but if I'm willing to attend some event and I miss it because of this I would be furious.

So, in my opinion, having a nice error with a clear message instead of an horrible backtrace seems more appropriate than hiding it.

Why is a build failure felt as a… failure? we all make mistakes, and this is very easy and quick to solve when you understand why it broke.

garrett:
Again, this doesn't have to do with events. At all. The events are already validated by CI, and as of yesterday, no changes are allowed in rh-events except by a passing pull request.

This has to do with a pretty display of a page's date object in the search results.

garrett:
Okay, so where's the failure right now? I'm looking through the recent PRs in `ovirt-site` to find it... I haven't seen it yet. Did someone push to master without going through a PR? Perhaps. Switching to evaluating every commit manually now. Sigh.

garrett:
The underlying problem I hit was a local issue — but one that others may possibly hit. If a file hasn't been checked into git, but it exists in the source tree, it will cause an error when it is indexed. I had a file to test a bunch of URLs to index while rewriting the indexing for #285 that was triggering it for me.

To work around this problem, I made indexing fall back to checking the file's modified time if it doesn't have a git status.

I'm going to open up a new PR with a new title for this one commit as an alternate approach instead.

PR #285, opened 2016-06-09
improved (ported) wiki link redirects

garrett:
Handle the old wiki redirects directly in Middleman instead of relying on Apache.

(This may solve a few 404s, as the logic is a bit more sound than the Apache redirects. Plus, it's better to not have to rely on Apache redirects anyway.)

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -265,7 +265,7 @@ helpers do
         args[url_index].gsub!(/https?:\/\/(www.)?ovirt.org\//, '')
       end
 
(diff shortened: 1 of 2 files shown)


garrett:
Note: I've included a commit from pull request #284, which is necessary for this branch to compile. When PR #284 is merged in, it should drop out of this PR.

garrett:
(In other words, we should merge #284 first.)

duck-rh:
LGTM

PR #288, opened 2016-06-13
search index: fall back to file modified if there's no git time

garrett:
Forked from #284 — basically, fall back to using the modified file time on the FS if there's no git time.

When search indexing occurs, it looks for a datetime stamp in the metadata first, then in git (if it doesn't find it), and now also in the file's mtime if it is not found it git.

This covers the case when someone is building the site but hasn't checked in a file quite yet. (Basically: Development work.)


duck-rh:
LGTM

PR #290, opened 2016-06-14
Update page to include the guest OS freezing

dyaniv:


source/develop/release-management/features/storage/backup-restore-api-integration.html.md
```diff
@@ -167,7 +167,7 @@ N/A
 
 *   Live snapshot : Done. Available in oVirt.
 *   Live Merge: Done. Available in oVirt.
-*   Qemu-ga: To be integrated.
+*   Guest Quiesce for Application-Level Consistency in Windows\Linux: Done. Available in oVirt.
```
mureinik: "Windows\Linux" should be "Windows/Linux". Also, it's probably worth mentioning **somewhere** (probably not in this patch) that the guest agent is required.

dyaniv: Done.

PR #297, opened 2016-06-19
vdsm devel: Require python3-yaml (for fedora)

EdDev:
Adding a require py3 package: yaml.


EdDev:
@mmirecki, @bronhaim, could you please review?

Issue #181, opened 2016-04-17
automatic page-level "table of contents" menus


didib:
In the old site, many/most pages had "Contents" automatically generated based on their internal section/subsection hierarchy. IMO we should have that by default for most pages of the new site, perhaps with a simple option to turn off/on. I searched a bit and could not find how to do that and/or whether it's supposed to be supported.

As an example, compare these:

http://old.ovirt.org/Hosted_Engine_Howto

https://www.ovirt.org/documentation/how-to/hosted-engine/

In addition, sections/subsections should be links to themselves, e.g. the title "Migrate existing setup to a VM", both in the menu and in the body, should be a link to the already-working URL:

https://www.ovirt.org/documentation/how-to/hosted-engine/#migrate-existing-setup-to-a-vm

(And, BTW, it will be nice if pressing this link will open it with the section title below the top bar, instead of being covered by it, as it is now at least for me (firefox 45)).

garrett:
I thought I had the ToC code on oVirt.org, but only had it on the RDO website. I've pulled the commits over and tested it locally. As soon as PR #304 is merged, you will have the ability to use the "toc" layout to easily have a generated table of contents for nearly any page (aside from feature pages).

To use, set a `layout` field in the frontmatter to `toc`. In other words, at the top of a file, you'd want something like this:

```
---
title: Some page
layout: toc
---

# Some page

## Details about some page

Things!
```

As you see, in the frontmatter block (the metadata for the page), there is a `layout: toc`. This one command instructs Middleman to do all the necessary stuff to generate a nicely formatted table of contents.

garrett:
I have added a pull request to address your "BTW" about anchor offsetting @ PR #309 — and added a little highlight when the content is linked too (to make it a little more obvious).

gregsheremeta:
closing old issues

PR #304, opened 2016-06-20
add toc (table-of-contents) layout

garrett:
Ported directly from the RDO website (thought it existed on oVirt too — I guess it was an oversight)


garrett:
This addresses #181.

PR #334, opened 2016-06-26
hosted-engine-host-OS-upgrade: Minor fixes

didib:
typo chane->change
'`' for inline code


duck-rh:
There was a CI problem (https://www.traviscistatus.com/) so could you please rebase your branch, which should trigger another CI job.

PR #347, opened 2016-07-04
Blog post on how to write application scoped unit tests for oVirt

rmohr:
Hi a technical post for engineers working on the engine. I also published it on [my blog] (http://rmohr.github.io/continuous%20integration/2016/07/04/application-scoped-tests-for-ovirt-with-arquillian).


bproffitt:
Okay, two things:

When you add tags to the metadata, they do not have to be in quotes.
When you start code with ```java or whatever, make sure there is an empty line above that marker.

I don't know which of these problems caused the Travis problem, but it rendered fine after I fixed these two things.

Article manually merged, edited, and pushed. Should auto-post later today.

Peace,
Brian

PR #374, opened 2016-07-20
fix changelog lists in 3.6 release page

rstory:
changelogs on release pages are all one big long wrapped line. I think this will get it to bulleted lists as intended, though it's hard to tell for sure since github preview is a little different that the actual live site.


sandrobonazzola:
Thanks! :+1:

PR #377, opened 2016-07-21
properly strip http(s)://(www.)ovirt.org/ from URLs

garrett:


garrett:
When I changed the block of code previously, I didn't change this part. Oops. Oversight fixed.

Issue #443, opened 2016-08-31
Python-sdk page has wrong images


rollandf:
Look at:
http://www.ovirt.org/develop/release-management/features/infra/python-sdk/
There are unrelated images and messed up text. 

old wiki:
http://old.ovirt.org/Python-sdk

mureinik:
PR #445 fixes this page, not sure why the issue is still open.

PR #445, opened 2016-09-01
Fix Python-sdk page

rollandf:
Fixes issue #443 

- Rename the images files
- Update the page to images new name
- Update python code block

I confirm that this pull request was submitted according to the [contribution guidelines](/.github/CONTRIBUTING.md): @rollandf

This pull request needs review by: @mureinik


rollandf:
For some reason, if the image name was for example "1.png", a random image that has as 1.png as suffix was loaded instead (eg "DomainUpgrade1.png").

I did not investigate the root cause of the issue, I just renamed the images to fix it quickly.
There is probably a framework fix needed to tackle this.

mureinik:
Looks good to me.

@jhernand can you confirm this is how the page is supposed to look?

PR #454, opened 2016-09-07
Add appliance install command

irosenzw:


mykaul:
:+1: 

(perhaps worth mentioning it might take time, as the appliance download time might be long?)

fabiand:
:+1:

PR #456, opened 2016-09-08
Allow explicit dns configuration bug1160667

mmucha-redhat:

- feature page for "Allow explicit dns configuration bug1160667"

I confirm that this pull request was submitted according to the [contribution guidelines](/.github/CONTRIBUTING.md): (please @mention yourself to sign)


mmucha-redhat:
@jhernand 
@dankenigsberg 
@mmirecki 

please review.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+<xs:complexType name="DnsConfiguration">
+  <xs:complexContent>
+    <xs:extension base="BaseResource">
+      <xs:sequence>
+        <xs:element maxOccurs="1" minOccurs="0" name="ips" type="Ips"/>
+      </xs:sequence>
+    </xs:extension>
+  </xs:complexContent>
+</xs:complexType>
+```
```
jhernand: The XML schema is a by-product of the specification of the API. If you use it here people may think that it is the primary mechanism to specify API types. I'd suggest you use an XML (or JSON) example instead, like you did below.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+
+<host>
+    <name>$NAME</name>
+    <address>$ADDRESS</address>
+    <port>$PORT</port>>
+    <root_password>$PASSWORD</root_password>
+    <cluster>
+        <name>$CLUSTERNAME</name>
+    </cluster>
+    <dns_configuration>
```
jhernand: Will `dns_configuration` be an attribute of the `Host` type? Will it be returned when retrieving the host?

```
GET /hosts/123
<host>
  ...
  <dns_configuration>
    ...
  </dns_configuration>
</host>
```

This is acceptable only if the answer is _yes_, otherwise we need to find a different mechanism to send the DNS configuration.

Please don't use the `ip` type, it is already abused enough. It is better to new elements, and I also think that the name should be `dns_resolver_configuration`:

```
<dns_resolver_configuration>
  <domain>example.com</domain>
  <search_list>
    <search>example.com</search>
    <search>internal.net</search>
    ...
  </search_list>
  <name_servers>
    <name_server>1.1.1.1</name_server>
    <name_server>2.2.2.2</name_server>
  </name_servers>
  <options>
    <option>
      <name>timeout</name>
      <value>10</value>
    </option>
    <option>
      <name>debug</name>
    </option>
    ...
  </options>
</dns_resolver_configuration>
```

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+Content-Type: application/xml
+
+<network>
+  <dns_configuration>
+    <ips>
+      <ip>
+        <address>8.8.8.8</address>
+      </ip>
+    </ips>
+  </dns_configuration>
```
jhernand: Same, updating the network this way is only acceptable if `dns_configuration` will become an attribute of the network, and will be reported when retrieving the network (`GET /networks/123`).

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+Content-type: application/xml
+
+<network_attachment>
+  <dns_configuration>
+    <ips>
+      <ip>
+        <address>8.8.8.8</address>
+      </ip>
+    </ips>
+  </dns_configuration>
```
jhernand: Same, only acceptable if this becomes an attribute of the network attachment.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: I wrote this to explain, that there will be new element created. This element isn't defined by xml/json (however I can put example there), and most of our customers need not to understand our api-model. Therefore I put here xschema. 

So to avoid confusion you mentioned — how about to putting here api-model code followed with info, that this definition will be during build expanded to this xschema? That way we're specifying here true element definition in both ways: our api-model and it's equivalent xschema.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: I believe, that dns_configuration _can_ be part of management network. Same for network_attachment of such management network.

Does other agree with this?

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: As you prefer. It is just that XML schema isn't how we specify the API and isn't how it is used. Both the actual specification and the XML example are better. The XML example is better for users, as they can use it directly. So I prefer the XML (or JSON) example, but this isn't that important.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: following edward decision, that dns_configuration is _owned_ by network this configuration isn't property of host. However, we need to pass information about this configuration during new host creation. When new host is added to system, management network is created for it. So we need to say — during creation of new host, use this dns configuration for creating related management network. 

Can you recommend better way how to do that? Creating host and then updating management network would work, but I don't find it very nice approach.

about ip type: yes, it's ugly. I thought we want to reuse it. I'll update page to follow your recommendation, but if I'm not mistaken, we currently do not need anything else but nameservers. Should I in that case specify other fields as well (domain, search_list, options) or we could rather add them only when they are needed?

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: I believe, that dns_configuration can be part of management network network attachment. 

Does other agree with this?

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: When a host is created a management network is created for it? That is not my understanding. Isn't it that when a host is added one of its network interface cards is attached to the management network of the cluster? If the DNS configuration is associated to the network, then there is no need to pass it when adding the host, rather the engine should automatically take it from the selected network (or from wherever the engine considers appropriate). As that isn't what we do today (if I understand correctly) and we need to stay backwards compatible, this behavior should be controlled by an optional query parameter:

```
POST /hosts?preserve_dns_resolver_configuration=false
<host>
  ...
</host>
```

The default value of this `preserve_dns_resolver_configuration` should be such that things work as they used to work in the past: `true`.

Alternatively, and I think this is better, the `dns_resolver_configuration` should be an attribute of the API `Host` type. If it isn't (if it isn't reported during GET), then we won't allow providing it during POST or PUT operations.

During `setupnetworks` I think it doesn't make sense to specify the DNS resolver configuration as part of a specific network attachment, it should rather be a top level parameter of the operation:

```
POST /hosts/123/setupnetworks
<action>
  ...
  <dns_resolver_configuration>
    ...
  </dns_resolver_configuration>
</action>
```

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Allow Explicit Dns Configuration
+
+### Owner
+
+*   Name: [ Martin Mucha](User:mmucha)
+*   Email: mmucha@redhat.com
+
+## Summary
+When new host is added to the system, management network will be created for it. Dns configuration for such network will be obtained from resolv.conf file. However admin should be able to specify overriding configuration in WebAdmin. He can do that at several places:
```
dankenigsberg: please use short(er) text lines, which are easier to review.

"when new host" -> when *a* new host

keep DNS capitalization all over the doc.

Please add "As of ovirt-4.0.3, DNS configuration for such"...

With this feature implemented,  an admin should be able...

He can -> She can (to be polite)

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: What is the purpose of having the DNS resolver configuration as part of the network attachment type? Only to be able to include it when doing `setupnetworks`. In that case I think it is better to have it as an additional parameter of the `setupnetworks` operation.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: The additional `domain`, `search_list` and `options` attributes are just examples. They don't need to be added now if they aren't going to be used.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+* By updating management network of given DC (or later of given Cluster) 
+
+## Detailed Description
+
+So DNS configuration can be stored in multiple places. If DNS configuration is not specified in either of them, no DNS configuration data will be passed to VDSM during SetupNetworks command and default values from resolv.conf will be used. If either of them is used, it will be passed to VDSM overriding configuration from resolv.conf. For example we can set DNS configuration in Network related to certain DC (or later to certain Cluster) and this configuration will override configuration in resolv.conf. If we now setup configuration in NetworkAttachment, attaching ManagementNetwork of given DataCenter (or later of given Cluster) to specific Host, this configuration is more specific than one in Network of DC (or later Cluster) scope and will be used instead.  
+
+### DB
+Database needs to be updated so it can accommodate dns configuration. Two places will be altered:
+
+* network_attachments table
```
dankenigsberg: I love rich formatting. `network_attachment` looks cool.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+* network table
+
+both this tables must accommodate multiple DNS records. Naive solution would be to store them comma separated into single column. Better solution would be creating separate table for that. That's subject of further discussion.
```
dankenigsberg: this tables -> these tables.

Note that Vdsm can handle 0 to 3 DNS entries, but no more than that.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
dankenigsberg: I can add my vote for an xml snippet that an admin can use.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
dankenigsberg: I believe that @mmucha-redhat meant that a management network *attachment* is created for the host when it is added to Engine. Indeed, the logical management network must exist before adding a host.

I do not think that we need ?preserve_dns_resolver_configuration=false. If someone does not want to modify their host's dns, they can simply keep the dns config on the management network unset. this would be the case after upgrade. (@mmucha-redhat please refer to upgrade explicitly in the text).

Beyond taking the DNS from the management network, we wanted to have the flexibility to set the definition on a single host, via its management network attachment. For completeness, it should be possible to assert the host-specific definition right when adding a host. However, I don't see a very clear "business case" for this flexibility. I think that the few users who would like to have a host-specific DNS can start by having the network-defined one, and only later - replacing the configuration with a different command.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+          <ip>
+            <address>8.8.8.8</address>
+          </ip>
+        </ips>
+      </dns_configuration>
+</host>
+```
+
+#### Updating Management Network
+only management network can be updated with dns configuration.
```
dankenigsberg: Please note that "management network" is a role that multiple networks may have (but only one in a specific cluster). So I think that *any* network may have DNS configuration, but on a specific cluster, only the config of its management network applies.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
dankenigsberg: I think that the concept of "network attachment" is equivalent to "all the parameters to be sent in order to reproduce the connectivity of a host onto a network". It should also be stored somewhere, and retrieved in order to debug issues. an throwaway param to `setupnetworks` is not like that; an element of network attachment is.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+      <ip>
+        <address>8.8.8.8</address>
+      </ip>
+    </ips>
+  </dns_configuration>
+</network>
+```
+
+#### Updating 'network attachment' of Management Network on specific Host 
+only network attachment of management network can be updated with dns configuration.
```
dankenigsberg: English: *O*nly *the* network attachment of *the* (maybe *a*?)
dns->DNS

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Creating new Host
+![New Host Dialog with DNS Configuration](newHostDialogWithDnsConfiguration.png "New Host Dialog with DNS Configuration")
+
+#### Updating Management Network
+(note — when editing other network this option won't be available)
+![Edit Logical Network Dialog with DNS Configuration](editLogicalNetworkDialogWithDnsConfiguration.png "Edit Logical Network Dialog with DNS Configuration")
+
+#### Updating 'attachment' of Management Network on specific Host 
+![Editing Network Attachment Dialog with DNS configuration](editNetworkAttachmentDialogWithDnsConfiguration.png "Editing Network Attachment Dialog with DNS configuration")
```
dankenigsberg: Please add a "Testing" section with:
* When no DNS is defined, the host predefined one remains configured. This is the case upon upgrade to ovirt-4.1 or fresh installation.
* When adding a host, its DNS is taken from the logical management network, overridden by network attachment.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,132 @@
+---
+title: AllowExplicitDnsConfiguration
+category: feature
+authors: mmucha
+wiki_category: Feature
+wiki_title: Features/AllowExplicitDnsConfiguration
+wiki_revision_count: 12
+wiki_last_updated: 2015-10-14
+feature_name: Allow Explicit Dns Configuration
```
dankenigsberg: Dns⟶DNS

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: We need a mechanism to preserve backwards compatibility of the API. Currently when a host is added (with `POST /hosts`) its DNS resolver configuration isn't touched, if I understand correctly. Users may rely on that: an administrator may have manually and carefully configured the `resolv.conf`file, before adding the host, and these changes would be lost without warning if the management network has a custom DNS resolver configuration. That is the objective of the `preseve_dns_resolver_configuration`, keep backwards compatibility and still be able to use the new behavior.

We can't include the DNS resolver configuration in the request to add a host unless it becomes an attribute of the `Host` type. When you do a `POST` to the hosts collection the body of the request is the XML (or JSON) representation of the API `Host` type. So if we want to be able to send this:

```xml
<host>
  ...
  <dns_resolver_configuration>
    ...
  </dns_resolver_configuration>
</host>
```

Then we have no alternative but changing the definition `Host` type in the specification of the API like this:

```java
@Type
public interface Host {
  ...
  /**
   * Blah, blah, blah ...
   */
  DnsResolverConfiguration dnsResolverConfiguration();
  ...
}
```

If we do that then the generated documentation and the SDKs will list `dns_resolver_configuration` as just another attribute of the `Host` type. Same will happen for all the SDKs. Users will then expect to be able to use it when adding, retrieving and updating hosts. But that won't be true, it will only be available when adding hosts. That is confusing, at least. We could alleviate the problem with documentation:

```java
/**
 * This attribute only makes sense when adding a host. It won't be populated when
 * retrieving a host, and it will be ignored when updating.
 */
DnsResolverConfiguration dnsResolverConfiguration();
```

Will users ever read that? Most won't.

So the best solution is to either make this a real attribute, that works for adding, retrieving and updating, or else not have it at all.

It is OK to have it as an attribute of the `Network` type, assuming that it works for adding, retrieving and updating networks.

It is OK to have it as an attribute of the `NetworkAttachment` type, again assuming that it works for adding, retrieving and updating network attachments.

For `setupnetworks` it is OK to have it as attributes of the network attachments if the meaning is that the provided value will be part of the created network attachment. If the provided value will only be used to update the host and won't be stored as part of the created network attachment, then it is better to have it as parameter of the operation.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
jhernand: If the it stored as part of the new attachment, and available when retrieving, adding and updating it, then it is OK. If not, if it will just be used to configure the host but it won't be stored, then it should be a parameter of the `setupnetwors` operation.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: updated. 

- used 'dns_resolver_configuration' name instead
- removed 'ip' element usage in favor of new 'name_server' element
- listed both model and XSchema definition with stress on fact, that XSchema is generated one.
- added exemplary xml snippet

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: > Will users ever read that? Most won't. 
really I wouldn't care about illiterate users. Any STOP sign on road also isn't surrounded by bubble foil  to save those, who can't read or are just to lazy to do so ;)

> Users may rely on that: an administrator may have manually and carefully configured the resolv.conffile, before adding the host, and these changes would be lost without warning if the management network has a custom DNS resolver configuration.
well _user_ of this rest is actually an admin. So if he edited the network, to contain dns configuration, he shouldn't be surprised, that this change takes place when adding new host. I don't assume, that there are lots of engine admins, so the one, who is adding new host is completely unaware of having management network changed. Therefore I have similar opinion as dan — I don't see need for having preserve_dns_resolver_configuration. Touching of dns configuration during adding new host will happen only for admin, and management network could have been also edited only by admin. And who else than admins should be familiar with system upgrade and its consequences...

> I believe that @mmucha-redhat meant that a management network attachment
yes, indeed. Sorry about confusion. Reading your discussion I feel, that because of all consequences and little usability, we can strip out manually setting dns configuration from adding a new host. We can live with having it set on network, from which it will be copied to network_attachment during adding new host. If admin needs to override some single host, he can in this exceptional case update management network - network attachment _[management network on one specific host]_ using separate call. If you agree, I'd remove usecase of setting dns configuration as a part of adding new host.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: 1. I don't know how to use shorter lines — if I press enter, paragraph breaks into two. Please advise.
2. done
3. I hope it's done (maybe except for screen shot, which are harder to recreate).
4.
5.done
6. This is joke? Discriminating one gender is inappropriate, while discriminate another one is polite?!?? What a time to live in! Is simple 'he/she' sufficient, or shall I, not to offend anyone, enroll all currently "known" genders, which are(if I'm up-to-date): man, woman, not-identifiable, transgender-man, transgender-woman, pending-transgender-man, pending-transgender-woman and finally none.

dankenigsberg: 1. Pressing typing a single Enter should break only the line. Having two consecutive Enters should break a new paragraph. It works well in other feature pages that I've read. Maybe you have an issue with your editor?
3. ok!
6. It has been a bit jocular and out of the point, but there's a grain of truth in it - I think we should use both genders interchangeable.

mmucha-redhat: ad 3. One guy here did research on this subject. Word 'admin' is of male gender, so usage of 'he' is correct. This is obvious part. Another correct usage is plural form, so if we must be 'politically correct' this is the way to go, although it's customary based on human history to use 'he-admin'. Usage of 'she-admin' is allegedly grammatically incorrect. 

Therefore I changed it to plural form.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: I formatted 'network_attachments' to use monospace if that's what you've been after.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: what would be the benefit of having DNS records which does not apply? Allowing to update it on any network will a) clutter db with potentially useless data b) can confuse user. Therefore I proposed "only management network can be updated with dns configuration." ie. rejecting user input if he tries to pass dns configuration for non-mgmt network. Having it on any network is easier, of course, but only slightly. If you want it for some (hopefully good) reason, it's no problem to have it like that.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+# Allow Explicit DNS Configuration
+
+### Owner
+
+*   Name: [ Martin Mucha](User:mmucha)
+*   Email: mmucha@redhat.com
+
+## Summary
+When a new host is added to the system, management network will be
+created for it. As of ovirt-4.0.3, DNS configuration for such network
```
dankenigsberg: How about the more precise: "management network will be created for it" -> it would be attached to the management network
such network -> this network attachement

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+will be obtained from resolv.conf file. With this feature implemented,
+an admins should be able to specify overriding configuration in
+WebAdmin. They can do that at several places:
+
+* During creation new host
```
dankenigsberg: I believe we are dropping the first bullet.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+* By updating management network of given DC (or later of given Cluster) 
+
+## Detailed Description
+
+So DNS configuration can be stored in multiple places. If DNS
+configuration is not specified in either of them, no DNS configuration
+data will be passed to VDSM during SetupNetworks command and default
+values from resolv.conf will be used. If either of them is used, it
+will be passed to VDSM overriding configuration from resolv.conf. For
+example we can set DNS configuration in Network related to certain DC
```
dankenigsberg: configuration in Network -> configuration in *a* network

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+later of given Cluster) to specific Host, this configuration is more
+specific than one in Network of DC (or later Cluster) scope and will
+be used instead.  
+
+### DB
+Database needs to be updated so it can accommodate DNS configuration.
+Two places will be altered:
+
+* `network_attachments` table
+* network table
```
dankenigsberg: use the same font for `network` table (to avoid confusion)

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+
+both these tables must accommodate multiple DNS records. Naive
+solution would be to store them comma separated into single column.
+Better solution would be creating separate table for that. That's
+subject of further discussion. VDSM supports 0 to 3 DNS entries, so
+engine needs to comply to this limitation as well.
+
+### REST
+
+As mentioned, you can specify DNS Configuration at three places.
```
dankenigsberg: three -> two

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+      <name_servers>
+        <name_server>1.1.1.1</name_server>
+        <name_server>2.2.2.2</name_server>
+      </name_servers>
+    </dns_resolver_configuration>
+</host>
+```
+
+#### Updating Management Network
+only management network can be updated with DNS configuration.
```
dankenigsberg: It would be better to let any network have `dns_configuration`, because any network can become the management network of a future cluster. On each cluster, only the config of the management network applies.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+    <name_servers>
+      <name_server>1.1.1.1</name_server>
+      <name_server>2.2.2.2</name_server>
+    </name_servers>
+  </dns_resolver_configuration>
+</network>
+```
+
+#### Updating 'network attachment' of Management Network on specific Host 
+only the network attachment of the management network can be updated
```
dankenigsberg: only->Only

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+      <name_server>2.2.2.2</name_server>
+    </name_servers>
+  </dns_resolver_configuration>
+</network_attachment>
+
+```
+
+### GUI
+
+As mentioned, you can specify DNS Configuration at three places:
```
dankenigsberg: three->two

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Creating new Host
```
dankenigsberg: we're dropping this section, I believe!

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,196 @@
...
+![New Host Dialog with DNS Configuration](newHostDialogWithDnsConfiguration.png "New Host Dialog with DNS Configuration")
+
+#### Updating Management Network
+(note — when editing other network this option won't be available)
```
dankenigsberg: other->another

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
dankenigsberg: Mostly English nits, but a quite important change is needed all over the doc: we do not allow setting the DNS during add host flow. All references to this (including the image) should be dropped.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: a) actually if it is first host I believe the mgmt network will be created , but updated as requested.

b) if DNS configuration is not mentioned in engine backend, resolv.conf is used, but ONLY in VDSM. I have no instructions to obtain data from VDSM and store them into network attachment. (Have I?) Thus, with VDSM/host not having any network attachments, I do believe there should be 'network' and not 'network attachment'. Please confirm.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: this comment is obsoleted by request to have dns configuration on any network. Done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

mmucha-redhat:
mburman asked question about supporting 'out of sync' for this. Do we want to support this?
If so, it means, that:

- during creating new host DNS configuration must be copied from network to network_attachment
- each network DNS configuration must be reported in getCaps call.

if we don't want to support that, it means, that:

- after dns configuration is altered in engine, no further changes made in resolf.conf applies, and if they do (no idea how vdsm works here), it will be confusing for the engine user, since other value would be reported than one actually used.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,176 @@
...
+    </name_servers>
+  </dns_resolver_configuration>
+</network>
+```
+
+#### Updating 'network attachment' of Management Network on specific Host 
+You can update any network with DNS Configuration, however all such DNS
+Configurations will be simply ignored during creation of new host except 
+for the DNS configuration defined on the network, which happens to be 
+management network at time of creating new host.
```
dankenigsberg: This paragraph does not belong here, but to the one above it. It seems to be speaking about networks and not about network attachments. I think that the former text is correct:
 "Only the network attachment of the management network can be updated with DNS configuration."

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Updating (Management) Network
+You can update any network with DNS Configuration, however all such DNS
+Configurations will be simply ignored during creation of new host except 
+for the DNS configuration defined on the network, which happens to be 
+management network at time of creating new host.
+![Edit Logical Network Dialog with DNS Configuration](editLogicalNetworkDialogWithDnsConfiguration.png "Edit Logical Network Dialog with DNS Configuration")
+
+#### Updating 'attachment' of Management Network on specific Host 
+![Editing Network Attachment Dialog with DNS configuration](editNetworkAttachmentDialogWithDnsConfiguration.png "Editing Network Attachment Dialog with DNS configuration")
```
mburman5: Can we please add 'out-of-sync' scenario for manual changing of the resolv.conf under the engine's legs.
For example: Changing the name server in resolv.conf and then doing refresh caps in UI should end up in 'out-of-sync' management network. This should be applied also on multi-host command.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,177 @@
...
+</network_attachment>
+
+```
+
+### GUI
+
+As mentioned, you can specify DNS Configuration at two places:
+
+#### Updating (Management) Network
+You can update any network with DNS Configuration, however all such DNS
```
mburman5: We don't like the idea to allow to update any logical network with DNS configuration. 
It make no sense at all. Maybe we can find other solution on the host level, as DNS is a host property and not network.

mmucha-redhat: discuss it with edward, who thinks opposite and is against what you've suggested.

dankenigsberg: It's not just "edward"; it's also the simplified API discussed with juan, as well as simplified user flows.

In reality, it is extremely rare to have one host with a specific DNS configuration. Most commonly, the DNS configuration is shared among all hosts in the subnet. Think about your DHCP-supplied DNS address, for example. We want to allow overriding the network-wide default on a specific host (via edit network), but this is for exceptional cases.

myakove: We (QE) are OK with this implementation.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -40,11 +40,30 @@ values from resolv.conf will be used. If either of them is used, it
 will be passed to VDSM overriding configuration from resolv.conf. For
 example we can set DNS configuration in a Network related to certain DC
 (or later to certain Cluster) and this configuration will override
-configuration in resolv.conf. If we now setup configuration in
+configuration in resolv.conf. Note: when Network is being attached in
```
dankenigsberg: I'd start a new paragraph with "Note"

when Network-> when *a* network

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -40,11 +40,30 @@ values from resolv.conf will be used. If either of them is used, it
...
+certain host, NetworkAttachment record is created for this, and 
+dns configuration is copied from Network record to NetworkAttachment 
+record. There we can further modify it. If we now setup configuration in
 NetworkAttachment, attaching ManagementNetwork of given DataCenter (or
 later of given Cluster) to specific Host, this configuration is more
 specific than one in Network of DC (or later Cluster) scope and will
-be used instead.  
+be used instead.
+  
+Copying dns configuration from Network record to NetworkAttachment is
```
dankenigsberg: dns->DNS (here and elsewhere in your current commit)

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -40,11 +40,30 @@ values from resolv.conf will be used. If either of them is used, it
 will be passed to VDSM overriding configuration from resolv.conf. For
 example we can set DNS configuration in a Network related to certain DC
 (or later to certain Cluster) and this configuration will override
-configuration in resolv.conf. If we now setup configuration in
+configuration in resolv.conf. Note: when Network is being attached in 
+certain host, NetworkAttachment record is created for this, and 
+dns configuration is copied from Network record to NetworkAttachment 
+record. There we can further modify it. If we now setup configuration in
```
dankenigsberg: What happens when the cluster-wide configuration changes? It would be nice if all hosts are updates (similarly to a change of vlan or label), except for hosts that have a local override.

I think that it would be better if we can avoid this copy until someone wants to override the cluster-wide configuration. 

Please add a discussion about what happens when we change DNS config in the network while it is already the management network of multiple hosts.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: alright then:

- no copying. 
- updating Network, which was already attached to some host, will make all it's relevant network attachments (which does not override it) out-of-sync. User can then click sync all network to reapply dns configuration.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
dankenigsberg: In QoS, or VLAN we perform a multi-host action to update all attached hosts with the new value. Why not do it so with DNS as well? Even if the reason is the willingness to keep this feature small and simple, please state it in the feature page.

mmucha-redhat: done.

mykaul:
What happens with DHCP interfaces? Do we expect to override the DHCP settings or not?

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,202 @@
+---
+title: AllowExplicitDnsConfiguration
+category: feature
+authors: mmucha
+wiki_category: Feature
+wiki_title: Features/AllowExplicitDnsConfiguration
+wiki_revision_count: 12
+wiki_last_updated: 2015-10-14
```
mmirecki: Date

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
```diff
@@ -0,0 +1,202 @@
...
+
+# Allow Explicit DNS Configuration
+
+### Owner
+
+*   Name: [ Martin Mucha](User:mmucha)
+*   Email: mmucha@redhat.com
+
+## Summary
+When a new host is added to the system, it would be attached to the
```
mmirecki: would -> is

mmucha-redhat: done.

source/develop/release-management/features/network/allowExplicitDnsConfiguration.html.md
mmucha-redhat: done.

mmucha-redhat:
Btw. quite essential question: what addresses are supported/wanted? IPV4 only or ? If both, we probably have to update design to mitigate Ipconfiguration/IpAddressAssignment in NetworkAttachment, so both can be used. Can someone advise?

dankenigsberg:
We need to add a section about DHCP: if the network is attached to a host via DHCP, it might get a conflicting DNS definition from there. Currently in Vdsm, we have PEERDNS=yes, which means that DHCP overwrites ovirt-side definition.

mmucha-redhat:
added note about dhcp. Feature page already contains info, that out-of-sync calculation might not be delivered in first version, so I did not think it's needed to add any more stress on that. What is supported and what not in this are probably should be noted in release notes, once feature is merged.

dankenigsberg:
@mmirecki we can merge the feature page as it is, as long as @mmucha-redhat remembers that out-of-sync should not be set if the host has dhcp configured on it.

PR #461, opened 2016-09-14
removed wiki-conversion debugging bar

garrett:
As the old wiki archive went away, and the site is long out of beta, the debugging toolbar no longer makes sense, and its generated links are broken.

(cc: @duck-rh for review)


garrett:
Sorry; was duplicated; @duck-rh beat me to it in #460.

PR #467, opened 2016-09-16
Fix 3.6 to 4.0 upgrade documentation

mwperina:
Fixes 4.0 repository URL which should be used to install oVirt 4.0 when
doing upgrade from 3.6 using engine-backup.


mwperina:
@sandrobonazzola @didib Please take a look

Issue #475, opened 2016-09-19
None of the "Talk:" links in feature pages are working


rollandf:
In many feature pages there is a "Talk:<some topic>" link that leads to nowhere.

For example:
http://www.ovirt.org/develop/release-management/features/engine/get-route/

These links needs to be removed unless there is an equivalent possibility in the new wiki.

rollandf:
This issue can be closed.

PR #476, opened 2016-09-19
removed all lines with `Talk:` links

garrett:
This removes all (deprecated/broken) `Talk:` links.

See issue #475.

This is the result of running the following command (in zsh, for extended globbing support):

```zsh
ruby -i -ne 'print unless /[^ ]Talk:/i' **/*.md
```


rollandf:
:+1: 
It will leave a lot of empty "Comments and Discussion" sections, but is better than dead links.

sandrobonazzola:
:+1:  Merging.

PR #480, opened 2016-09-21
add feature page for gdeploy cockpit-ovirt plugin integration

dpmramesh:

- Add feature page for integrating gdeploy with cockpit-ovirt plugin. This will helps to facilitate the easy deployment of Hyper-Converged Ovirt-Gluster environment.

I confirm that this pull request was submitted according to the [contribution guidelines](/.github/CONTRIBUTING.md): @dpmramesh

This pull request needs review by: @sabose @fabiand  @evol262


source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
```diff
@@ -0,0 +1,140 @@
...
+
+We need following changes to implement this feature.
+
+### UI Changes:
+
+* Provide a option in 'Hosted-Engine' section to start the Hosted-Engine deployment with Gluster. This should invoke the gdeploy
+wizard.
+
+* Gdeploy wizard shown in previous section needs to implemented using react-js. At the final step, this wizard should call a script(python?) using cockpit.spawn() to generate the config files for 'gdeploy' and
+'hosted-engine'. Once user reviews and confirms the generated config files then execute the 'gdeploy' command using cockpit.spawn('gdeploy -c gluster.conf'). UI should track the execution and report the results as and when different steps are being executed by gdeploy.
```
evol262: Why python?

If the utility doesn't already exist, cockpit.file is very capable of writing a configuration file, and it may be significantly nicer than passing a bunch of state variables from react to a script.

Sending JSON over directly to the python script also works, but it would be preferable to have all the logic in one place (dashboard/src/helpers/Gluster.js, for example) if there's not an existing utility already.

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+    ![gdeploy-summary](gdeploy-summary.png)
+
+
+    ![gdeploy-log](gdeploy-log.png)
+
+
+* Hosted-Engine Setup:
+  After completing the gluster deployment, it will start the standard Hosted-Engine deployment by invoking the 'hosted-engine --deploy' with the generated answer file. Standard HE deployment work flow will start from this step.
```
evol262: Just to be clear, dashboard/src/helpers/HostedEngineSetup will need a change in order to accept some optional argument (for the file location).

After this, setup needs to be invoked directly from react, which will call hosted-engine --deploy, otherwise the setup UI won't appear.

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
dpmramesh: I am not aware of cockpit.file. But as of now, we don't have the utility to create gdeploy conf file. As you suggested, let me explore  cockpit.file and make use of it directly instead of python.

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
dpmramesh: Thanks @evol262 for the quick review.  You are right. We need enhancements in the current plugin to support extra arguments . Let me mention this explicitly in the feature page.

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
```diff
@@ -0,0 +1,140 @@
...
+
+* Easy deployment using Cockpit UI
+* Well integrated deployment of HCI solution
+* Consistent Gluster deployment using gdeploy
+
+### Dependencies
+gdeploy package should be installed and running. gdeploy requires password less ssh from the gdeploy host to other hosts where we want to deploy gluster. This needs to be done manually before starting deployment flow.
+
+**TBD:** explore a way to configure password less ssh using Cockpit itself.
+
```
mareklibra: Since the deployment is intended to work on newly added host without the need forf excessive pre-configuration, I think the easiest way to achieve the pwd-less ssh is just to scp public keys, so the password needs to be entered just once during this initialization process.

I don't think there's some explicit support for this task in cockpit so far.

dpmramesh: Thanks @mareklibra. We will take this approach for now.

sabose:
Ack from me

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
evol262: Well, cockpit can already connect to remote hosts. It's smart enough to try to reuse the same password.

I don't see this exposed in the API documentation (it may be a dbus call somewhere), or if reuse is supported, but it may be possible to use this in order to drop files into remote costs through the cockpit dbus API. Maybe a longshot, though. @stefwalter, any thoughts?

sabose:
Can this be merged?

dpmramesh:
@sandrobonazzola Can you merge this pull request?

satheesaran:
@dpmramesh - thanks for the cleanly written feature page and also the mock-ups. All looks good and clear

source/develop/release-management/features/gluster/gdeploy-cockpit-integration.html.md
stefwalter: So @petervo is doing work to directly connect via SSH through to remote hosts from a central cockpit-ws instance. He's doing this for integration with ManageIQ ... but it'll be ready for oVirt too.

We should plan this out. Perhaps talk about it on IRC first?

PR #481, opened 2016-09-22
Add a feature page for page for NGN RPM persistence

evol262:

- Add a feature page for Next Gen Node RPM persistence

This pull request needs review by: @fabiand @dougsland @irosenzw

### DIFF
--- a/source/develop/release-management/features/node/node-next-persistence.html.md
+++ b/source/develop/release-management/features/node/node-next-persistence.html.md
@@ -97,6 +97,7 @@ The yum plugin will add a hook into `pretrans`
 * Inside this transaction set, if a package is being installed or updated, `TransactionSet.installed` will contain all package objects
 * We can iterate over the package objects, and retrieve `TransactionSet.PackageObject.localpkg()` (this is actually `po.localpkg`, since yum uses short variable internally
 * The package will be copied over to `/var/imgbased/persisted-rpms`
+* Using the yum cache is not reliable, as `yum clean all` would remove cached RPMs, and `yum localinstall` never touches the cache
 
 Additionally:
 
--- a/source/develop/release-management/features/node/node-next-persistence.html.md
+++ b/source/develop/release-management/features/node/node-next-persistence.html.md
@@ -1,0 +1,90 @@
+---
+title: Next Gen Node RPM persistence
+category: feature
+authors: rbarry
+wiki_category: Feature|Next Gen Node RPM Persistence
+wiki_title: Feature/NextGenNodeRPMPersistence
+wiki_revision_count: 0
+wiki_last_updated: 2016-09-20
+feature_name: Reinstall/Persist RPMs after Node upgrades
+feature_modules: node persistence
+feature_status: WIP, 4.1 proposed feature
+---
+
+# Node RPM Persistence After Upgrades
+
+## Summary
+[oVirt Node NG](http://www.ovirt.org/develop/projects/node/4.0) is an advancement over vintage Node, especially that it allows users to install RPMs or customize the system through configuration management tools. Some of these changes are migrated over from /etc after upgrades, and some are kept persistenly in /var, but any packages which are installed on a running layer will be lost after upgrades and need re-installation. This presents some problems with vendor tooling, IPMI, utilities to collect statistics, and other add-ons.
+
+This feature presents a mechanism by which packages installed through yum/dnf can be saved and automatically re-applied when the OS is updated.
+
+### Owner
+* Name: Ryan Barry
+* Email: rbarry@redhat.com
+
+### Current Status
+* Status: Design
+* Last updated date: 20th Sep 2016
+
+### Benefit to oVirt
+
+This feature will greatly improve the user experience of Node by maintaining a consistent userland and system configuration after upgrades. This allows for Node customizations to "stick".
+
+### Dependencies
+None! We simply need to add an additional yum plugin to imgbased.
+
+## Detailed Description
+
+Node NG performs a number of actions when updates are applied.
+
+* A new layer is added
+* The new squashfs is extracted onto that layer
+* `/etc`/ and `/root` are rsynced to the new layer
+* UIDs and GIDs are synchronized across packages, in case a drift occurred inbetween images for UIDs/GIDs which are not fixed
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+* A new bootloader entry is added and set as the default
+
+The existing osupdater logic can be extended (as the new root filesystem is already mounted) in order to install a package (or packages) into the new layer before it is booted for the first time.
+
+A reliable mechanism must be written in order to capture packages as they are installed, however.
+
+[yum.PackageObject](http://yum.baseurl.org/api/yum-3.2.26/yum.packages.PackageObject-class.html) is able to immediately provide the location of a cached RPM (after downloading or from localinstall), and this is available as part of yum plugin hooks.
+
+It _may_ be possible to [extend RPM](http://www.rpm.org/wiki/DevelDocs/Plugins), but some research must be done in order to discover whether RPM's plugin interface is as informative as yum.PackageObject. It is possible that RPM's plugin mechanism only triggers during the execution of the RPM specfile, and cannot be used to pin down the location of the actual package. This also provides the additional burden of maintaining a small codebase in C, though that may be an acceptable tradeoff.
```
fabiand: Might be interesting.

Tho I'd suggest that we reach the right feature level with `yum` and then take off to other worlds.

source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+    posttrans
+    postverifytrans
+    predownload
+    prelistenabledrepos
+    prereposetup
+    preresolve
+    pretrans
+    preverifytrans
+    verify_package
+
```
fabiand: Can we limit this list to the one hook we use. We can still link to the complete list - which might be elsewhere.

source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+While not all of these hooks run while yum has the package cached, we can rely on `pretrans` to have the cached package available, both for `yum localinstall` and `yum install`.
```
fabiand: This is actually nice.

source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+
+Though the hook should be limited to only run during certain transaction types (install, upgrade), the transaction package object is available, and we can grab the local file from [YumAvailablePackage.localPkg](http://yum.baseurl.org/api/yum/yum/packages.html#yumavailablepackage) and save it off to /var, which is a persistent filesystem on Node NG, which is not versioned per layer.
```
fabiand: Reusing `/var` is actually a good idea.
I assumed that we'll have a separate volume - but `/var/…` sounds just fine.

Please add the following:

* A list of the yum verbs where we are hooking into - and how the hooking looks like
* A list which represents the flow
* A definition of how and where we store the rpms
* Discuss if there is a configuration file or not

source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+### imgbased extension
+
+With the packages saved in /var, imgbased.osupdater can set up a local yum repository which points at `/var/imgbased/persisted_packages` (pathname to be determined), and install them when osupdater triggers during a Node update.
```
fabiand: Did you consider to use a separate imgbased plugin for this?

evol262: See above.

source/develop/release-management/features/node/node-next-persistence.html.md
evol262: I mean, this is a feature page. Not a patch. The link to yum documentation on writing plugins will have to do for "how the hook" works. I could summarize it here, but it's better to link their docs.

Note that yum plugins do not hook into verbs. They hook into transactional steps, and the command arguments can be fetched (or extended with argparse) from there. This has some advantages. Primarily, the transaction set is able to be fetched from the conduit (see the "Writing Yum Plugins" documentation). For verbs which don't install anything ("remove", "uninstall", etc), the transaction set has no installed elements anyway. Directly reusing the transaction set to get package objects means that we don't need to worry about verbs at all.

Where the RPMs are stored is listed below.

source/develop/release-management/features/node/node-next-persistence.html.md
evol262: The complete list is not elsewhere, hence the list here. There is a link to the list on "Writing Yum Plugins", but it's out of date.

source/develop/release-management/features/node/node-next-persistence.html.md
evol262: That's the concept.

source/develop/release-management/features/node/node-next-persistence.html.md
```diff
@@ -0,0 +1,90 @@
...
+
+Node NG performs a number of actions when updates are applied.
+
+* A new layer is added
+* The new squashfs is extracted onto that layer
+* `/etc`/ and `/root` are rsynced to the new layer
+* UIDs and GIDs are synchronized across packages, in case a drift occurred inbetween images for UIDs/GIDs which are not fixed
+* A new bootloader entry is added and set as the default
+
+The existing osupdater logic can be extended (as the new root filesystem is already mounted) in order to install a package (or packages) into the new layer before it is booted for the first time.
```
evol262: Yes, it would be a different plugin. I mentioned osupdater, since that has similar behavior.

source/develop/release-management/features/node/node-next-persistence.html.md
fabiand: Yes, you are right - I might have requested to many details.
The level is good, and the important point that you want to hook into yum is clear.

evol262:
@fabiand +1? All your comments should be resolved in the follow-up patch sets from 26 days ago

fabiand:
My bad, yes :+1:

PR #489, opened 2016-09-26
summarize only the latest 3 blog posts on front page

tigert:

- load only 3 latest blog entries on front page, blog/ contains the rest.

### DIFF
--- a/source/index.html.haml
+++ b/source/index.html.haml
@@ -75,7 +75,9 @@ hide_metadata: true
           = data.site.name
           News
 
-        = partial :blog_posts, locals: locals
+        -### Load last 3 blog entry summaries ###
+        - page_articles[0..2].reject {|a| !a.published? }.each do |article|
+          = partial :blog_post, locals: {article: article, summarize: 1}
 
         = link_to 'See all blog posts', '/blog/'
 


rollandf:
Nice idea, it will make the homepage look better.
@garrett can you review ?

source/index.html.haml
```diff
@@ -75,7 +75,9 @@ hide_metadata: true
           = data.site.name
           News
 
-        = partial :blog_posts, locals: locals
+        -### Load last 3 blog entry summaries ###
+        - page_articles[0..2].reject {|a| !a.published? }.each do |article|
```
garrett: This rejects unpublished after selecting 3 articles, meaning if you have 1 unpublished, then you'll have 2 articles. If you have 2 unpublished, you'll have 1 article. It should be something like this:

```ruby
- page_articles.select(&:published?).take(3).each do |article|
```

source/index.html.haml
```diff
@@ -75,7 +75,9 @@ hide_metadata: true
...
+          = partial :blog_post, locals: {article: article, summarize: 1}
```
garrett: ...Or, of course, the `source/layouts/blog_posts.haml` could be modified to support a # of articles for the front page too, so you don't have to do any of this here.

(This would be the better approach.)

garrett:
See #496 for the alternate approach I suggest. (This has the benefit of being portable to other sites as well as having the abstraction in the correct place.)

garrett:
Merged #496.

PR #552, opened 2016-10-24
remove `http://www.ovirt.org` from the link

### LINKED ISSUE #518: Broken links due to wrong URL rewriting
### Description of the problem

http://www.ovirt.org/develop/infra/testing/test-cases/ is rewritten as http://www.ovirt.org/develop/release-management/features/os-support/fedora-24-support/develop/infra/testing/test-cases/

It shouldn't.

### Steps to reproduce

1.Go to http://www.ovirt.org/develop/release-management/features/os-support/fedora-24-support/

2.Try to follow "Test Case" link in Testing section

garrett:
Fixes #518:

Never link to the website as an external resource. It's a bad idea for so many different reasons, with the most important being that it breaks the possibility to test the site locally. (In addition, it seems to mess up server redirection in some cases.)

### DIFF
--- a/source/develop/release-management/features/os-support/fedora-24-support.html.md
+++ b/source/develop/release-management/features/os-support/fedora-24-support.html.md
@@ -40,7 +40,7 @@ Add support for Fedora 24
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
Thanks!

PR #555, opened 2016-10-24
Fix broken self links

garrett:
Applying the first to #518 across the whole site — stripped http(s)://(www.)ovirt.org from the Markdown files.


garrett:
@sandrobonazzola: Can you look this over please?

I've only changed the actual links themselves — not the appearance (that is, in the few cases where a link displays that it links to `http://ovirt.org` (or similar), I haven't changed what is there to read... just the link URL itself).

PR #565, opened 2016-10-28
Added 102816 blog

bproffitt:
Changes proposed in this pull request:
- New blog highlighting the new code of conduct
## 
## 

I confirm that this pull request was submitted according to the [contribution guidelines](/.github/CONTRIBUTING.md): @bproffitt

PR #570, opened 2016-10-31
'summarize' parameter must be numeric or Nil

duck-rh:
This fixes PR #565 's build failure. This problem was introduced in b7c24a0 with the article limiting feature.

I also wonder what this line wants to achieve:
`summarize ||= summarize`
I guess it is supposed to be numeric to hold the default value site-wide.

Having @garrett 's insight before merging would be much better.

### DIFF
--- a/source/index.html.haml
+++ b/source/index.html.haml
@@ -75,7 +75,7 @@ hide_metadata: true
           = data.site.name
           News
 
-        = partial :blog_posts, locals: {limit: 3, summarize: true}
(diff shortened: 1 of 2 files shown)


sandrobonazzola:
@garrett can you please review?

garrett:
As far as I can tell: it's fine. 

I don't see how this wasn't an issue ever before though. Did something else change (such as a gem version)?

PR #659, opened 2016-12-15
Fix url and pick a better word

nirs:
I did not complain but reported an issue.

@ifireball, @rollandf please review.


duck-rh:
Being out of it I can say this is misrepresentation, and the link is much better too.

ifireball:
Hmm looks like the message IDs in mailman changed. @duck-rh how could that be? We need trustworthy links to the ML archives!

PR #668, opened 2016-12-23
Replace DC 4.1 page with Storage Domain V4 page

mureinik:
Each oVirt release introduces a new DC level with, potentially, many
features. This page listed a very small subset of them, which was
incomplete, and would it be completed it should not have resided under
storage.

This patch removes the aforementioned patch and replaces it with a page
discussing the actual new feature here from the storage perspective -
the V4 domain, and the features it enables (which is not, of course, a
full list of storage features in oVirt 4.1).

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @mureinik

This pull request needs review by: @rollandf @maorlipchuk


rollandf:
👍

mureinik:
@maorlipchuk I'm not sure where to place this so it won't overwhelm the page. Like you suggested, I'll merge this one, and if you can send a patch for the REST details, we can discuss something concrete.

Issue #398, opened 2016-08-10
Provide a way how to include generated HTML into oVirt site


mwperina:
We should host oVirt API description on ovirt.org site, but those description are usually generated directly from source code into HTML. In order to support that we need to have some directory (for example /api) which will allow us to upload generated HTML content into (without doing markdown compilation). At the moment I see following docs which should be included like that:

  oVirt Engine Extension API
  oVirt Engine Extension AAA API
  oVirt Engine Extension Logger API
  oVirt RESTAPI & model
  oVirt Java SDK
  oVirt Python SDK
  oVirt Ruby SDK
  oVirt Ansible Module Document (until it's included into Ansible core/extras)

But they may be also others.

The idea is that all of those will have unique subdirectory under /api and so we can create links from markdown generated oVirt site into those directories.

sandrobonazzola:
@mwperina  IIRC all the html is generated from asciidoc sources right?

sandrobonazzola:
@garrett any suggestion?

mwperina:
@sandrobonazzola No, JavaDoc is generated directly from Java sources using javadoc tool, PythonSDK will most probably use pdoc tool which generates HTML from Python sources and comments using asciidoc formatting, Ansible Module is using Ansible tool to generate doc from Python source. So generally we should just include generate code as is and display it directly (no inclusion into oVirt site styles).

garrett:
Do you have URLs to any examples of the current input and ouput?

mwperina:
For example JavaDoc for ovirt-engine-extensions-api is generated from java sources at:

https://github.com/oVirt/ovirt-engine/tree/master/backend/manager/modules/extensions-api-root/extensions-api/src/main/java/org/ovirt/engine/api/extensions

The HTML output is contained in 

http://resources.ovirt.org/pub/ovirt-4.0/rpm/el7/noarch/ovirt-engine-extensions-api-impl-javadoc-4.0.1.1-1.el7.centos.noarch.rpm inside /usr/share/doc/ovirt-engine is directory ovirt-engine-extensions-api

which is the one we would like to put into ovirt.org site.

For Python SDK sources are at 

https://github.com/oVirt/ovirt-engine-sdk

prerelease version of HTML output which @machacekondra is currently working on is available at 

https://machacekondra.github.io/ovirt-engine-sdk/docs/ovirtsdk4/index.html

garrett:
So wait — are there two _different_ API sources, tools, and outputs?

[...]

PR #681, opened 2017-01-05
added rest-api documentation

rafaelmartins:

- added `ovirt-engine-api-model`'s gh-pages branch as a subrepo, to deploy rest-api docs to the website. This is not the final solution, it is just a PoC to show how it works, I'm not sure if `/documentation/rest-api/` is the best place for the docs, and if we should mention `master` docs on the website.

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @rafaelmartins

This pull request needs review by: @sandrobonazzola @jhernand @phoracek @MarSik @mykaul (anyone else is free to review)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -240,6 +240,3 @@ DEPENDENCIES
   therubyracer
   wdm (~> 0.1.0)
   wikicloth
-
-BUNDLED WITH
-   1.13.2
--- a/source/release/4.1.0/index.html.md
+++ b/source/release/4.1.0/index.html.md
@@ -7,17 +7,14 @@ authors: sandrobonazzola
 # oVirt 4.1.0 Release Notes
 
 The oVirt Project is pleased to announce the availability of 4.1.0
-Second Beta as of December 21st, 2016.
+First Release Candidate as
+of January 23, 2017.
 
 oVirt is an open source alternative to VMware™ vSphere™, and provides an awesome
 KVM management interface for multi-node virtualization.
 This release is available now for Red Hat Enterprise Linux 7.3, CentOS Linux 7.3
 (or similar).
 
-
-Please note that new qemu / kvm features introduced in Red Hat Enterprise Linux 7.3
-are not yet available on ppc64le since CentOS 7.3 has not been released at the time of
-this BETA compose.
 
 This is pre-release software.
 Please take a look at our [community page](/community/) to know how to
@@ -44,7 +41,7 @@ An updated documentation has been provided by our downstream
 
 ## RELEASE CANDIDATE
 
-In order to install this BETA RELEASE you will need to enable pre-release repository.
+In order to install this Release Candidate you will need to enable pre-release repository.
 
 
 In order to install it on a clean system, you need to install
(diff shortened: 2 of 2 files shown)


source/documentation/index.html.haml
```diff
@@ -47,6 +47,10 @@ wiki_last_updated: 2014-04-26
       - [Quick Start Guide](/documentation/quickstart/quickstart-guide/)
       - [oVirt Administration Guide](/documentation/admin-guide/administration-guide/)
       - [oVirt User Guide](/documentation/user-guide/user-guide/)
+      - REST API:
+          [oVirt 4.1](/documentation/rest-api/4.1/),
+          [oVirt 4.0](/documentation/rest-api/4.0/),
+          [oVirt master (development version, not recomended for end users)](/documentation/rest-api/master/)
```
rollandf: recomended->recommended

rafaelmartins: fixed.

bproffitt:
This is breaking in CI because files were checked out of the source/ directory, which fundamentally breaks things. There is a .git/ directory and a lot more vital files under this directory. Was this running on your local machine before you committed it @rafaelmartins ?

garrett:
1. Was this tested on someone's machine locally before being committed? (That is, was this actually built successfully? Just running the server isn't always enough, as it doesn't necessarily build all files.)
2. It's a bad idea in general to put non-document files in `source/`* (and in this case, that includes the `.git/` subdirectory) — Middleman does not like this

garrett:
Also:

1. Why is this a submodule?
2. Why is it a dump of (huge, that is > 2MB) HTML pages?

rafaelmartins:
@bproffitt yes, it runs using `./run-server.sh`.

@garrett I do not ran a full build, but looking at travis I know why it breaks. that's because we are lacking an index in the subrepo. I sent a pull request to the subrepo, adding it. the issue will be fixed as soon as it gets merged, and I update the subrepo here.

jhernand:
If it is that problematic to have this directly in the oVirt site, I'd suggest to just have the index page in the oVirt site, and then links to the external site.

rafaelmartins:
I think that having docs hosted in the main project domain should be a priority. I can't see a reasonable argument to do not integrate it this way.

garrett:
There was a previously discussion about API docs belonging on a separate subdomain (afer a discussion about the various pros and cons) @ https://github.com/oVirt/ovirt-site/issues/398. 

Suggested solution @ https://github.com/oVirt/ovirt-site/issues/398#issuecomment-244046989 also has many benefits, such as being in a uniform UI, would be searchable, and can be accessed offline as well.

rafaelmartins:
So, you are "suggesting" to "rewrite" the docs, to use a standalone tool, that needs to be forked, "adapted" and deployed, instead of integrating what we have already, and let users without "official" docs meanwhile?

I have no problem with subdomains, but we need something that is available right now. Users are without decent docs for very long already. We should not step back just because something is not ideal. We can improve it later, but the current situation is not acceptable, IMO.

bproffitt:
If you can get this fixed with the patch you do in the subdomain, I would do that now, and we can look at the #398 method later. I would rather have working documentation up ASAP, too.

rafaelmartins:
Ok, I'll comment here when it is done.

Thanks

rafaelmartins:
sorry, I broke this by mistake, and it was outdated anyway. will reopen a correct pr soon

PR #817, opened 2017-02-23
Plan for how to delegate parts of VM-lifecycle to Kubevirt

ahadas:
New wiki page that details the first phases in integrating ovirt with kubevirt in regards to VM-lifecycle

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @ahadas

This pull request needs review by: @fabiand @michalskrivanek @mykaul

### DIFF
--- a/source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
+++ b/source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
@@ -1,0 +1,130 @@
+---
+title: VM lifecycle in Kubevirt 
+category: feature
+authors: arik
+wiki_category: Feature
+wiki_title: Features/VM lifecycle in Kubevirt
+wiki_revision_count: 20
+wiki_last_updated: 2015-10-21
+feature_name: VM lifecycle in Kubevirt
+feature_modules: engine
+feature_status: 
+---
+
+This page proposes a gradual way to integrate the virtualization management platform oVirt and containers management platform Kubernetes.
+
+
+# Motivation
+
+The virtualization management platform oVirt and the containers management platform Kubernetes share common capabilities. For example, both platforms need to determine on which node to run their managed entity (VM/container), run and then monitor it.  
+
+It is thus tempting to reuse the common parts of these platforms instead of developing and maintaining a dedicated infratructure in each of them.  
+
+However, it is not as easy as it may sound. The differences between virtual machines and containers require different capabilities and decision making by the management platform. Moreover, these two platforms are written in different technology and designed differently.  
+
+Therefore, it only makes sense to consider a gradual integration between the two platforms by delegating stuff from oVirt into Kubevirt, an extended form of Kubernetes that includes components that are specific to virtual machines. A natural choice would be to start with delegating the scheduling and virtual machine creation from oVirt to Kubernetes.  
+
+# The Current Process in oVirt
+
+In this section I elaborates on the process of running a virtual machine in oVirt. The creation of the virtual machine is done as a separate operation before it is being run so we can assume that the VM already exists in the database at this stage.  
+
+The process looks as follow:  
+0. Optional: a request is sent from client (UI/REST-API).  
+1. Instance of RunVmCommand is created.  
+2. The input is being verified. Specifically, the scheduler validates that there is a host capable of running the VM.  
+3. The engine retrieves all the properties of the VM to run.  
+4. The scheduler chooses a host to run the VM on.  
+5. The engine locks host devices, plug lun disks and update cinder disks.  
+6. The engine locks the monitoring thread.  
+7. The engine sends a request to VDSM on the chosen host to run the VM.  
+8. If the response if positive, the engine set the VM to WaitForLaunch state (and few other fields).  
+9. The monitoring lock is released.  
+10. The monitoring receives the VM as running (positive flow) or down (negative flow). If the VM fails to run, the engine tries to run it few more time (by default 2 additional attempts).
+
+# Proposed Approach
+
+So this proposal split the integration between oVirt and Kubernetes in regards to VM lifecycle into several phases. Note that some of them can be done in parallel.   
+
+## Milestone 1: Basic integration
+
+### Goal
+1. Add Kubevirt into oVirt.  
+2. Kubevirt sends the run-request to the relevant node.  
+
+### Design
+We aim to have Kubevirt as a configuration of a cluster so eventually we would be able to have a Data-center with cluster with lower compatibility version that works as it used to work before and cluster with the up-to-date compatibility version that works with Kubevirt. That way, we enable the user to upgrade existing system without having to stop the VMs.  
+
+So a cluster would have a property called "type". Initially the type will have two values: legacy, kubevirt. If 'kubvirt' is chosen, the user will need to provide the URL and authentication details of Kubevirt.  
+
+Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
+
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) rather than VDSM. Note that the scheduling is still done in oVirt.  
+
+The represenation of the VM will be as follow (in JSON):  
+1. Domxml (Kubevirt passes it as-is to the node at this stage, later on Kubervirt will replace placeholders within the XML).  
+2. The scheduled host (Kubvirt uses this as the destination node).
+
+### Benefits
+1. First notion of Kubernetes in oVirt + start getting feedback on running real-world virtual machines with Kubevirt.    
+2. We bypass the 'create' verb in the engine<->vdsm API.  
+3. We do not depend on having a full blown scheduler in Kubervirt.
+
+### Limitations & Challenges
+1. At this stage the engine does not do the operations that it used to do after the detination host was chosen and before sending the request to VDSM.  
+2. The engine cannot lock the monitoring until VDSM acknowlege that the request arrives. So we need to set a timeout so if the VM is still down afterwards, we'll conclude that the run operation failed.
+3. oVirt still schedules the nodes.
+4. oVirt still monitors the nodes as it used before.  
+5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
+
+## Milestone 2: Scheduling
+
+### Goal
+Basic VM schedluing in Kubevirt.
+
+### Design
+1. Extend the VM representation with cluster-related data, e.g., host-pinning. This will be represented as properties in JSON. VM-related data needs to be retrieved from the VM configuration (see VM Represenation section).  
+2. Adjust VMs monitoring as described above. That is, to add a timeout for run-vm so the monitoring will end the run-vm operation if the VM is down after some time.  
+3. Some basic mointoring should be done in Kubevirt in order to rerun a virtual machine (oVirt cannot do this at this phase since the engine doesn't know which nodes were previously chosen). Kubevirt is reponsible to detecting failures to run the VM on the node and to attempt to reschedule the VM on another node.
+
+### Benefits
+1. We start to extract the scheduling part from oVirt.  
+2. We add basic monitoring capability to Kubevirt (to examine the state of the VM on the scheduled node and retrieve information that is needed for scheduling).  
+3. We extract the rerun mechanism for run-VM from oVirt.  
+
+### Limitations & Challenges
+1. Either use the 'stop-the-world' approach of the current scheduler in oVirt or invent an alternative approach.  
+2. The scheduling does not take into consideration host-devices.  
+3. We still do not do the post-scheduling and pre-interaction with VDSM operations that we used to do.  
+4. Most of the VMs monitoring remains the same.  
+5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
+6. If we fail to run a VM on a node several times in a short-period of time, the node should be non-operational (and thus not be scheduled until fixed).
+
+## Milestone 3: Dynamic Cluster State Managed by Kubevirt
+
+### Goal
+At this point we have basic monitoring of VMs. This milestone is about detecting more changes in running VMs in Kubevirt and report them to oVirt. Ideally, most of VmDynamic would be managed in Kubevirt and will be in oVirt only for display purpose. This allows us to switch VM management operations (shutdown, migrate, suspend, ..) to be done via Kubervirt.
+
+### Design
+1. Receive the VMs state from the node and updating the dynamic part in the VM representation.  
+2. Starting highly-available VMs in Kubevirt.
+3. Change VM management operations: hibernate, restore, resume, migrate, shutdown, reboot, cancel-migration, power-off.
+
+### Benefits
+1. Kubevirt becomes the owner of the runtime state of the cluster while oVirt manages its static state (host-deployments, VM templates, import VMs from external providers, and so on).  
+2. Many VM operations are delegated to Kubevirt.
+
+## Future Changes
+
+Future changes depend on having the knowledge of storage and network resources in Kubevirt and thus harder to predict.  
+
+## VM Representation
+
+I believe that the best way to represent the VM is in JSON as follow:  
+1. Scheduling data: the cluster-related data mentioned before, in the form of properties in JSON.  
+2. VM dynamic data: things like status of the VM, run_on_vds, console ip, and so on. Initially they should be set with default values. Once the monitoring in Kubevirt receives them from the node, it should update the VM representation and some component needs to listen to these changes and report them to the engine in oVirt. Note that we may want to report some of the dynamic data that exist today, like the applications list, in a different way.   
+3. VM configuration: the static configuration of the VM, including its devices.  
+
+In oVirt we are in the process of representing the VM configuration in XML (domxml) to eliminate the duplicate conversion of the data (once in the engine, converting the internal representation into a dictionary, and once in VDSM, converting the dictionary into domxml). We would prefer to keep this approach and minimize data conversions. In light of this approach, there are few options of how to represent the VM configuration:  
+1. To generate the VM representation in JSON. There is a proposal (on-going work?) to implement another binding for libvirt that will enable to pass the VM definition in JSON. With this, we would still have only one conversion of the VM configuration. That seems like the best approach since Kubernetes works with Yaml formats.  
+2. To have the VM configuration as XML blob inside JSON. This would mean that Kubevirt will have to parse VM-level data that it needs (for example, amount of memory for scheduling) by parsing the XML. The XML could be extended and place-holders can be replaced in Kubevirt but Kubevirt must not change other existing values in the XML.  
+3. To have the VM configuration as XML blob inside JSON but also to include some of the VM configuration, the ones that Kubevirt is interested in, as properties in JSON. The downside would be that the engine will need to provide higher amount of data. However, this way will simplify data retrieval in Kubevirt.  
--- a/source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
+++ b/source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
@@ -4,8 +4,8 @@ category: feature
 authors: arik
 wiki_category: Feature
 wiki_title: Features/VM lifecycle in Kubevirt
-wiki_revision_count: 20
-wiki_last_updated: 2015-10-21
+wiki_revision_count: 1
+wiki_last_updated: 2017-02-28
 feature_name: VM lifecycle in Kubevirt
 feature_modules: engine
 feature_status: 
@@ -48,7 +48,7 @@ So this proposal split the integration between oVirt and Kubernetes in regards t
 ## Milestone 1: Basic integration
 
 ### Goal
-1. Add Kubevirt into oVirt.  
+1. Initial notion of Kubevirt in oVirt.  
 2. Kubevirt sends the run-request to the relevant node.  
 
 ### Design
@@ -58,7 +58,7 @@ So a cluster would have a property called "type". Initially the type will have t
 
 Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
 
-Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) rather than VDSM. Note that the scheduling is still done in oVirt.  
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) that would send it to VDSM (VDSM is expected to run on each node in the cluster at the beginning) rather than sending it directly to VDSM. Note that the scheduling is still done in oVirt.  
 
 The represenation of the VM will be as follow (in JSON):  
 1. Domxml (Kubevirt passes it as-is to the node at this stage, later on Kubervirt will replace placeholders within the XML).  
@@ -72,8 +72,8 @@ The represenation of the VM will be as follow (in JSON):
 ### Limitations & Challenges
 1. At this stage the engine does not do the operations that it used to do after the detination host was chosen and before sending the request to VDSM.  
 2. The engine cannot lock the monitoring until VDSM acknowlege that the request arrives. So we need to set a timeout so if the VM is still down afterwards, we'll conclude that the run operation failed.
-3. oVirt still schedules the nodes.
-4. oVirt still monitors the nodes as it used before.  
+3. oVirt (engine) still schedules the nodes.
+4. oVirt (engine + VDSM) still monitors the nodes as it used before.  
 5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
 
 ## Milestone 2: Scheduling
@@ -94,7 +94,7 @@ Basic VM schedluing in Kubevirt.
 ### Limitations & Challenges
 1. Either use the 'stop-the-world' approach of the current scheduler in oVirt or invent an alternative approach.  
 2. The scheduling does not take into consideration host-devices.  
-3. We still do not do the post-scheduling and pre-interaction with VDSM operations that we used to do.  
+3. We still do not do the post-scheduling and pre-interaction with VDSM operations that we used to do (e.g., connect lun disks).  
 4. Most of the VMs monitoring remains the same.  
 5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
 6. If we fail to run a VM on a node several times in a short-period of time, the node should be non-operational (and thus not be scheduled until fixed).
@@ -107,7 +107,7 @@ At this point we have basic monitoring of VMs. This milestone is about detecting
 ### Design
 1. Receive the VMs state from the node and updating the dynamic part in the VM representation.  
 2. Starting highly-available VMs in Kubevirt.
-3. Change VM management operations: hibernate, restore, resume, migrate, shutdown, reboot, cancel-migration, power-off.
+3. Change VM management operations: hot-plug/unplug, hibernate, restore, resume, migrate, shutdown, reboot, cancel-migration, power-off.
 
 ### Benefits
 1. Kubevirt becomes the owner of the runtime state of the cluster while oVirt manages its static state (host-deployments, VM templates, import VMs from external providers, and so on).  


source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+3. Some basic mointoring should be done in Kubevirt in order to rerun a virtual machine (oVirt cannot do this at this phase since the engine doesn't know which nodes were previously chosen). Kubevirt is reponsible to detecting failures to run the VM on the node and to attempt to reschedule the VM on another node.
+
+### Benefits
+1. We start to extract the scheduling part from oVirt.  
+2. We add basic monitoring capability to Kubevirt (to examine the state of the VM on the scheduled node and retrieve information that is needed for scheduling).  
+3. We extract the rerun mechanism for run-VM from oVirt.  
+
+### Limitations & Challenges
+1. Either use the 'stop-the-world' approach of the current scheduler in oVirt or invent an alternative approach.  
+2. The scheduling does not take into consideration host-devices.
```
MarSik: The kubernetes scheduler is missing much much more than that. We are investigating a way to still call the oVirt scheduler from the kubernetes world for the first phases of development (either as a separate microservice, kubernetes scheduler extension or through a proxy acting as such).

ahadas: That's true that we may need to extend this point with more stuff that we declare we don't intend to provide at this phase, but  ideally this won't be much.
I'll describe what I had in mind while writing this: I imagine that the biggest challenge is not to rewrite the logic inside oVirt scheduler, but to receive all the data that the oVirt scheduler receives today. So I thought in this phase that someone is writing in kubervirt what we have in oVirt, but some of the input like host-devices won't be available and so won't be taken into consideration. But maybe I'm wrong here - do you see a major challenge in rewriting the scheduler?

MarSik: We are investigating that right now actually and that is why we want to use the current code from the engine at first. We are not even sure if we can extend the kubernetes scheduler with what we really need (for example - they compute cpu/memory load very differently from us), integrate with it (using https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduler_extender.md) or if we need to provide our own. And it all gets more complicated when you realize that our balancing needs to use the same rules to stay consistent. We do not even have a balancing design atm (neither live migration iirc).

ahadas: So we have to assume that we have an impact on where the pod runs because virt-handler is supposed to run the VM on the node that pod was scheduled to, right?
As long as either the scheduling or the monitoring is done by oVirt, I imagine that each node has a label with its identifier in oVirt so once the nodeName field of the VM objects is set to that identifier the VM specification is magically reaches the node and virt-handler starts the VM

ahadas: Names are typically something non-trivial to decide on :)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+10. The monitoring receives the VM as running (positive flow) or down (negative flow). If the VM fails to run, the engine tries to run it few more time (by default 2 additional attempts).
+
+# Proposed Approach
+
+So this proposal split the integration between oVirt and Kubernetes in regards to VM lifecycle into several phases. Note that some of them can be done in parallel.   
+
+## Milestone 1: Basic integration
+
+### Goal
+1. Add Kubevirt into oVirt.
```
fabiand: I understand this line as "add kubevirt to the environment in which oVirt runs" do you mean it this way or do you mean something else?

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+So a cluster would have a property called "type". Initially the type will have two values: legacy, kubevirt. If 'kubvirt' is chosen, the user will need to provide the URL and authentication details of Kubevirt.  
+
+Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
+
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) rather than VDSM. Note that the scheduling is still done in oVirt.  
+
+The represenation of the VM will be as follow (in JSON):
```
fabiand: Here we really need to speak about the details of the generated domxml and see how we represent it in kubevirt.

For the beginning however we can probably start with a simple template / hard coded spec.

/cc @admiyo @rmohr

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+So a cluster would have a property called "type". Initially the type will have two values: legacy, kubevirt. If 'kubvirt' is chosen, the user will need to provide the URL and authentication details of Kubevirt.  
+
+Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
+
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) rather than VDSM. Note that the scheduling is still done in oVirt.
```
fabiand: Should we note - eventually in a separate paragraph - that vdsm is still expected to run on every node in the cluster?

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+### Benefits
+1. First notion of Kubernetes in oVirt + start getting feedback on running real-world virtual machines with Kubevirt.    
+2. We bypass the 'create' verb in the engine<->vdsm API.  
+3. We do not depend on having a full blown scheduler in Kubervirt.
+
+### Limitations & Challenges
+1. At this stage the engine does not do the operations that it used to do after the detination host was chosen and before sending the request to VDSM.  
+2. The engine cannot lock the monitoring until VDSM acknowlege that the request arrives. So we need to set a timeout so if the VM is still down afterwards, we'll conclude that the run operation failed.
+3. oVirt still schedules the nodes.
+4. oVirt still monitors the nodes as it used before.
```
fabiand: oVirt (Engine + vdsm)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+4. oVirt still monitors the nodes as it used before.  
+5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
+
+## Milestone 2: Scheduling
+
+### Goal
+Basic VM schedluing in Kubevirt.
+
+### Design
+1. Extend the VM representation with cluster-related data, e.g., host-pinning. This will be represented as properties in JSON. VM-related data needs to be retrieved from the VM configuration (see VM Represenation section).
```
fabiand: This needs careful design of review.

@ahadas do you have an example domxml generated by Engine - to start undertanding scope and content?

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
+
+## Milestone 2: Scheduling
+
+### Goal
+Basic VM schedluing in Kubevirt.
+
+### Design
+1. Extend the VM representation with cluster-related data, e.g., host-pinning. This will be represented as properties in JSON. VM-related data needs to be retrieved from the VM configuration (see VM Represenation section).  
+2. Adjust VMs monitoring as described above. That is, to add a timeout for run-vm so the monitoring will end the run-vm operation if the VM is down after some time.
```
fabiand: @mojaves do you actually plan to have a collectd container? :)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: So, I am not sure how it fit's in - but for node feature exposure I'm considering a separate daemon which is just responsible for extarcting host features (based on plugins), and then there is a separate entitiy which is assigning labels to hosts based on features.

I.e. `label.has_virt: (feat.has_svm || feat.has_vmx)`

as a simple example.

My assumption would then be that the scheduler can take these labels into account. This would align well with kube itself.
The gap however is to understand how we can cover "limited" resources like SRIOV vfs.

fabiand: Kubernetes can already do host pinning. So if we know which host oVirt wants to use, then Kube will place it there, or the pod will not be scheduled (can be queried/monitored).

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+1. We start to extract the scheduling part from oVirt.  
+2. We add basic monitoring capability to Kubevirt (to examine the state of the VM on the scheduled node and retrieve information that is needed for scheduling).  
+3. We extract the rerun mechanism for run-VM from oVirt.  
+
+### Limitations & Challenges
+1. Either use the 'stop-the-world' approach of the current scheduler in oVirt or invent an alternative approach.  
+2. The scheduling does not take into consideration host-devices.  
+3. We still do not do the post-scheduling and pre-interaction with VDSM operations that we used to do.  
+4. Most of the VMs monitoring remains the same.  
+5. Most of the VM related operations are still done directly with VDSM (rather than through Kubervirt).
```
fabiand: A challenge will be to let vdsm interact with the VMs scheduled by kubevirt.
We need to see how this works in detail.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+6. If we fail to run a VM on a node several times in a short-period of time, the node should be non-operational (and thus not be scheduled until fixed).
```
fabiand: Two questions come to me:
Who and where is the data? Separate service, general metrics?
How do we avoid rescheduling? Probably using [node cordon](https://kubernetes.io/docs/admin/node/)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
+---
+title: VM lifecycle in Kubevirt 
+category: feature
+authors: arik
+wiki_category: Feature
+wiki_title: Features/VM lifecycle in Kubevirt
+wiki_revision_count: 20
+wiki_last_updated: 2015-10-21
```
fabiand: That old :)

This and other fields are out of date.

fabiand:
Nice document. LGTM I just added a few non-blocking comments.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: I agree with @ahadas. We can work with the assumption that kubevirt will launch the VM where Engine specified it to run.

There are some glitches when it comes to live migration, but we'll get to them and see how we can address them.

MarSik: Sure, the first phase can let the engine to compute like it does now and then just pass that to kubernetes as a hard constraint for the pod.

MarSik: Yeah, I think that is the right approach for capabilities filters. SR-IOV is indeed an issue (can we  limit how many times a label can be used?)

fabiand: @MarSik Labels can actually have values, so we can do something like:

```yaml
labels:
- io.kubevirt.sla.sriov.vf_count: 42
- io.kubevirt.sla.sriov.pf_count: 2
```

Whenever this changes, the labels can be updated.

These labels can then be used with (anti-)affinity selectors, which allow the lt (less than) and gt (greater than) operators, see https://kubernetes.io/docs/user-guide/node-selection/.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+3. VM configuration: the static configuration of the VM, including its devices.  
+
+In oVirt we are in the process of representing the VM configuration in XML (domxml) to eliminate the duplicate conversion of the data (once in the engine, converting the internal representation into a dictionary, and once in VDSM, converting the dictionary into domxml). We would prefer to keep this approach and minimize data conversions. In light of this approach, there are few options of how to represent the VM configuration:  
+1. To generate the VM representation in JSON. There is a proposal (on-going work?) to implement another binding for libvirt that will enable to pass the VM definition in JSON. With this, we would still have only one conversion of the VM configuration. That seems like the best approach since Kubernetes works with Yaml formats.
```
rmohr: Kubevirts VM representation represents a cluster-view of the VM. It does not map 1:1 to the domain xml, so libvirt accepting JSON does not really help ... 

E.g. Storage, Network, ... are not even necessarily part of the VM on the cluster wide view. They will just contain references to them and only combined and transformed on the node level. We need to follow that principle to reuse kubernetes principles.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+## VM Representation
+
+I believe that the best way to represent the VM is in JSON as follow:  
+1. Scheduling data: the cluster-related data mentioned before, in the form of properties in JSON.  
+2. VM dynamic data: things like status of the VM, run_on_vds, console ip, and so on. Initially they should be set with default values. Once the monitoring in Kubevirt receives them from the node, it should update the VM representation and some component needs to listen to these changes and report them to the engine in oVirt. Note that we may want to report some of the dynamic data that exist today, like the applications list, in a different way.   
+3. VM configuration: the static configuration of the VM, including its devices.  
+
+In oVirt we are in the process of representing the VM configuration in XML (domxml) to eliminate the duplicate conversion of the data (once in the engine, converting the internal representation into a dictionary, and once in VDSM, converting the dictionary into domxml). We would prefer to keep this approach and minimize data conversions. In light of this approach, there are few options of how to represent the VM configuration:
```
rmohr: Most likely it can't be done that way. What really always needs to be kept in mind is, that in contrast to oVirt, all the logic in kubernetes is on the node. Also see the comment below for more details.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+3. VM configuration: the static configuration of the VM, including its devices.  
+
+In oVirt we are in the process of representing the VM configuration in XML (domxml) to eliminate the duplicate conversion of the data (once in the engine, converting the internal representation into a dictionary, and once in VDSM, converting the dictionary into domxml). We would prefer to keep this approach and minimize data conversions. In light of this approach, there are few options of how to represent the VM configuration:  
+1. To generate the VM representation in JSON. There is a proposal (on-going work?) to implement another binding for libvirt that will enable to pass the VM definition in JSON. With this, we would still have only one conversion of the VM configuration. That seems like the best approach since Kubernetes works with Yaml formats.  
+2. To have the VM configuration as XML blob inside JSON. This would mean that Kubevirt will have to parse VM-level data that it needs (for example, amount of memory for scheduling) by parsing the XML. The XML could be extended and place-holders can be replaced in Kubevirt but Kubevirt must not change other existing values in the XML.
```
rmohr: The Kubevirt VM object has a metadata section where you can embed something, but kubevirt itself will not interact with that metadata. Also it will not parse xml.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+1. To generate the VM representation in JSON. There is a proposal (on-going work?) to implement another binding for libvirt that will enable to pass the VM definition in JSON. With this, we would still have only one conversion of the VM configuration. That seems like the best approach since Kubernetes works with Yaml formats.  
+2. To have the VM configuration as XML blob inside JSON. This would mean that Kubevirt will have to parse VM-level data that it needs (for example, amount of memory for scheduling) by parsing the XML. The XML could be extended and place-holders can be replaced in Kubevirt but Kubevirt must not change other existing values in the XML.  
+3. To have the VM configuration as XML blob inside JSON but also to include some of the VM configuration, the ones that Kubevirt is interested in, as properties in JSON. The downside would be that the engine will need to provide higher amount of data. However, this way will simplify data retrieval in Kubevirt.
```
rmohr: From my perspective, ovirt-engine should just generate the expected JSONs. You can of course start simple.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: To highlight: KubeVirt expects and validates the JSON - we can not pass arbitrary data.

But that is okay. Once we have the flow to submit data to KubeVirt, then we can see where and how we tweak and translate the domxml to create a matching json representation which KubeVirt accepts.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: @fabiand , tes, that's something we should dive into indeed.

@rmohr , good point. I guess we can not support hotplug/hotunplug operations at the beginning then. I added this to Milestone 3.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: Changed

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: Added

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: Yes, rephrased

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: Updated :)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: sure: https://paste.fedoraproject.org/paste/V5HC94tf9nWMner61tpANF5M1UNdIGYhyRLivL9gydE=

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+So a cluster would have a property called "type". Initially the type will have two values: legacy, kubevirt. If 'kubvirt' is chosen, the user will need to provide the URL and authentication details of Kubevirt.  
+
+Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
+
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) rather than VDSM. Note that the scheduling is still done in oVirt.  
+
+The represenation of the VM will be as follow (in JSON):  
+1. Domxml (Kubevirt passes it as-is to the node at this stage, later on Kubervirt will replace placeholders within the XML).
```
ahadas: No, I meant libvirt's domxml (https://paste.fedoraproject.org/paste/V5HC94tf9nWMner61tpANF5M1UNdIGYhyRLivL9gydE=)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Benefits
+1. We start to extract the scheduling part from oVirt.  
+2. We add basic monitoring capability to Kubevirt (to examine the state of the VM on the scheduled node and retrieve information that is needed for scheduling).  
+3. We extract the rerun mechanism for run-VM from oVirt.  
+
+### Limitations & Challenges
+1. Either use the 'stop-the-world' approach of the current scheduler in oVirt or invent an alternative approach.  
+2. The scheduling does not take into consideration host-devices.  
+3. We still do not do the post-scheduling and pre-interaction with VDSM operations that we used to do.
```
ahadas: Added an example (there aren't many) - connection of lun disks

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: Is it that different than what we have in oVirt when thinking about the information in the engine as cluster wide view and the information in VDSM as node-specific view? 

Sure, we have to know the network configuration on the cluster wide view because we may use host-device passthrough so thus it may affect the scheduling - so I guess we will have to break these assumptions here and there.

But for the most part, the engine doesn't know things like the path that disks are mounted on the host (ok, it knows, but we try not to rely on that). So in the domxml I mentioned before you can see things like 'DISPLAY-NETWORK:ovirtmgmt' - that is a placeholder that engine used to signal VDSM "listen, I know that I need to display-network, but this is a node-specific information that I'm not aware of so please replace it with the actual display network for 'ovirtmgmt'".

If my comment is totally wrong, could you please explain in what aspect does the cluster wide configuration different than what the engine knows about the VM in oVirt? :)

rmohr: Sure, I think a good example is starting a VM **V** with a disk **D** from storage **S**

In oVirt it works basically the following way ( I guess it can differ a little bit depending on the storage used):

oVirt:

 1) Add storage **S** to the engine
 2) engine tells vdsm(s) to attach the storage **S**
 3) engine tells vdsm or storage to create an empty disk
 4) engine generates a domain xml (at least soon it will), and passes it to vdsm
 5) vdsm fills in some missing details into the domain xml, only it can know
 6) vdsm tells libvirt to start the VM

Kubevirt:

 1) Define a storage **S** on Kubevirt (Kubevirt does nothing with them so far)
 2) Define a volume claim **C** for **S** to Kubevirt (again Kubevirt does nothing with it so far)
 3) Add a VM Spec **V** which references the volume claim **C**
 4) Kubevirt schedules the VM **V** (find out the target node and start a container there where the VM should be at the end)
 5) virt-handler on the node now sees that VM **V** needs to be started on the node it is running on (it is not told to start it, it is watching the VMs)
 6) virt-handler sees that the VM **V** needs a volume from claim **C** from storage **S** (which are all not part of the VM spec, it will fetch and examine these resources)
 7) virt-handler mounts (or provides it for passthrough) that storage to the host if not present (so it is lazy, on a need basis)
 8) virt-handler makes sure that the volume to fulfill the volume claim **C** is created if not already present
 9) virt-handler generates the resulting domain xml with all the informations it now has
 10) virt-handler starts the VM

Note that the steps 6) to 10) might not necessarily be done by virt-handler. Some of the work might already be done between 4) and 5) by native kubernetes constructs, but in principle this is how it works.

Also note that the order of defining the dependent components 

For networks it works in a similar fashion.

So really the node is responsible for figuring out everything. For instance if you define a VM which references a non existent claim, the node will tell you that after it tried using it.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
rmohr: Could you elaborate who is doing that and what you mean by duplicating?

michalskrivanek: the most feasible way today for ovirt is to generate vdsm. now when you have vm spec with some properties then most/all of them are going to be in that xml as well

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
```diff
@@ -0,0 +1,130 @@
...
+
+So a cluster would have a property called "type". Initially the type will have two values: legacy, kubevirt. If 'kubvirt' is chosen, the user will need to provide the URL and authentication details of Kubevirt.  
+
+Note that we may use the same mechanism used in external provider to hold the authentication info. Another alternative is to use a full blown external provider for this, as we did for the integration with virt-v2v (where external provider is used only to contain pre-defined configuration of the remote system). Anyway, we do not have to decide this at this stage.  
+
+Then, the mentioned run-VM flow will be modified. The only change would be that in Kubvirt-cluster, CreateBrokerVDSCommand will generate a bit different configuration of the VM and will send it to Kubevirt (using the Kubernetes API) that would send it to VDSM (VDSM is expected to run on each node in the cluster at the beginning) rather than sending it directly to VDSM. Note that the scheduling is still done in oVirt.
```
fabiand: > that would send it to VDSM

KubeVirt would rather speak to libvirt directly to define the VM.
vdsm would not be involved here.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: Small note, for step 8: The claim is either bound to an existing volume, or a dynamic provisioner would have created a matching volume. But we should not replicate this functionality in KubeVirt, as it's present In K8s.
Thus in step 8 it would fail if the claim is unbound.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: To me it looks as if we are not to far apart. I understand the flow as follows:
In this initial phase, let's start with:
1. Generate the domxml as if it was for vdsm
2. In a second phase, transform this 1:1 into json
   Referenced host-paths or networks are kept as is - because we can assume that they will be there on the host, as vdsm is still available on the host.

So to me the first step really is to just transform the domxml into json - into the scheme which is recognized by kubevirt - and put it into a VM spec.

In some following steps we can then see how we slowly change host resource related parts to abstract counter parts.
I.e. if we take a disk from https://paste.fedoraproject.org/paste/V5HC94tf9nWMner61tpANF5M1UNdIGYhyRLivL9gydE=

In future this disk would be exposed in KubeVirt as a `PersistentVolumeClaim` of the type `vdsm`.
The claim would contain all the details which vdsm needs to mount/attach the correct SD.
In the domxml (then json) we would reference this claim i.e.: `<disk type='PersistentVolumeClaim' claimName='fooclaim'> … glory details about the bus connectivity … </disk>`
The handler would be responsible for writing a domxml (a real one he passes to libvirt) which is correct for volume claims.
Then the kubelet would call a (tbd) flevolume driver for `vdsm` type `PersistentVolumes` to attach the SD to the host and expose the image to the VM.

That's roughly how it could work.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: We will need to put some effort in order to make it possible.
For example, we may need to run a VM with payload. Currently VDSM creates a volume that stores its content and puts the path to that volume inside the domxml. If we skip VDSM, we need to prepare the payload volume and puts the path into the domxml in advance. The more common case is the call to prepareImage for the disks.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
ahadas: ack, so I guess that explains the extra steps we would need for migrations that @fabiand mentioned before, right? In oVirt if the host is up then, when putting aside host-devices and lun-disks, it is supposed to be capable to run the VM or to serve as a destination for migration, while in kubevirt some preparation is required and that may fail (like mounting the disk)

rmohr: @ahadas That too. I guess the main difference is that, while VDSM is the long arm of ovirt-engine in doing its tasks, Kubevirt manages VMs on it's own, it's own way. That for instance also means that a VM start *might* mean that the engine has to push multiple JSONs for different purposes ...

Again not clear enough what I say here. Let my try it again from a different angle: Kubevirt does nothing for ovirt-engine or vdsm and there will not be any oVirt related code in Kubevirt (like this templating thing with the xml where kubevirt would then just fill in some missing data). Kubevirt should have enough logic itself to run VMs, oVirt can just take the role of higher manangement. Finally in case of doubt which way to go, either the oVirt way or the Kubernetes way, Kubevirt will always prefer the Kubernetes way.

Of course that does not mean that Kubevirt can't support oVirt requirements (like providing storage or networks via vdsm, while starting  is done by kubevirt), but we will need generic extensions like [CNI](https://github.com/containernetworking/cni) or [flexVolumes](https://github.com/kubernetes/kubernetes/tree/master/examples/volumes/flexvolume) which can be plugged into KubeVirt. These plugins can the fill in the missing gaps.

To conclude this: I think the milestones @ahadas created here are reasonable, I just find it important to clarify things a little bit that we can see if the proposed ways fit into kubevirt and what needs to be changed there. We don't have to do it in this proposal, but we definitely need then follow up proposals for every milestone, where we can discuss details. But definitely XML plays no role in Kubevirt.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
fabiand: We'll need to see, but I think we can make this work, using a hook mechanism and eventually updating the paths.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: sounds ok. However the amount of labels will be significant and keeping order not so easy. There is a thin line between labels and host stats (e.g. in host devices with specific properties affecting relevance of various features), we also may need to update them often.
What is the tested/feasible amount of labels per host?

michalskrivanek: +1

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: @ahadas: the link doesn't work for me.
@fabiand - so what do you suggest as next step. We can think of a couple of those params already, we can start with those, push PR to kubevirt? At least for those which are already handled by kubernetes.
What about versioning? I assume the list is going to significantly change

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: I would not be eager to do that right away. Since the critical part is libvirt stats I'd rather stay as close to it as possible.
Current ovirt's collectd(added for metrics) is not containerized ATM so I'd just follow that (ansible deployment). 
We can revise that later on, it's not important

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: if we manage to move forward with collectd in ovirt there's not going to be much of VMs interaction in VDSM

@rmohr, hm how about supporting "learning" mode in kubevirt to get the updates from underlying libvirt xml? It would be handy eventually for import(externalVM-like functionality, but finally useful)

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: which data? about failures? I suppose keeping it as a part of host metrics is not a bad idea

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: rmohr, right. Though the fundamental part here is the GUI->config "mapping" which has a lot of logic itself.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: it should not be viewed as "engine->kubevirt" architecture. Indeed we strive for kubevirt to be *the* thing doing all the handling, eventually.
Your example with volumes is good, and preferring kubernetes way also makes sense, however I'm afraid that after handful of these examples we're in the area of virtualization or even ovirt specific features which simply have no kubernetes equivalent and we won't really have anything to follow, it just won't really fit.

The XML is just a start, I expect it to go away, and I expect the placeholders evolving into specific pieces of spec, in a lock-step with GUI changes generating those.

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: +1

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: yes, we need to start simple, hence the push on xml despite it being inconvenient. But it is the most simple way to move forward...

source/develop/release-management/features/virt/vm-lifecycle-with-kubevirt.md
michalskrivanek: @mojaves - I'd like to start splitting the "create" VM flow. It should be possible to get there, though I do not see that as important right now - we can e.g. call parts of vdsm separately, then send it the xml with placeholders, get back expanded xml and pass it to libvirt directly. But that tiny step of passing it to libvirt directly vs proxying through vdsm when you anyway have to call it couple times is insignificant...

ahadas:
@fabiand , @michalskrivanek, seems that this page now contains high-level details we agree on for how to proceed. can you please approve it so we can merge it?

michalskrivanek:
LGTM

fabiand:
:+1:

PR #873, opened 2017-03-27
added setup commands for Debian

lgh2:

- added a test of the linux distro to the setup.sh file and added commands to setup for Debian 

I tested the script on Fedora 25 (workstation edition) and on Mint Serena 18.1 (based on Ubuntu Xenial), and it worked on both. 
After running the setup.sh file on both systems, the run-server.sh script worked in both cases, so I didn't change it. 

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @lgh2

This pull request needs review by: @jasonbrooks

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,7 +1,21 @@
 #!/bin/sh
 
-git submodule init && git submodule update
+#setup for redhat
+if [ -f /etc/redhat-release ]; then
+  git submodule init && git submodule update
 
-sudo yum install -y ruby-devel rubygems-devel gcc-c++ curl-devel rubygem-bundler patch zlib-devel redhat-rpm-config openssl
+  sudo yum install -y ruby-devel rubygems-devel gcc-c++ curl-devel rubygem-bundler patch zlib-devel redhat-rpm-config openssl
 
-bundle install
+  bundle install
+# setup for debian 
+elif [ -f /etc/debian_version ]; then
+  git submodule init && git submodule update
+
+  sudo apt install -y build-essential ruby-bundler libcurl4-openssl-dev zlib1g-dev ruby-dev
+
+  bundle install 
+
+else
+    echo "Could not verify system is redhat or debian."
+    exit 1
+fi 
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -1,7 +1,21 @@
...
+# setup for debian 
+elif [ -f /etc/debian_version ]; then
+  git submodule init && git submodule update
+
+  sudo apt install -y build-essential ruby-bundler libcurl4-openssl-dev zlib1g-dev ruby-dev
+
+  bundle install 
+
+else
+    echo "Could not verify system is redhat or debian."
```
mykaul: We usually use the term 'EL' - Enterprise Linux, for Red Hat (not redhat) derivatives. 
In any case, 'Red Hat' and 'Debian' sounds better.

setup.sh
```diff
@@ -1,7 +1,21 @@
...
+  git submodule init && git submodule update
 
-sudo yum install -y ruby-devel rubygems-devel gcc-c++ curl-devel rubygem-bundler patch zlib-devel redhat-rpm-config openssl
+  sudo yum install -y ruby-devel rubygems-devel gcc-c++ curl-devel rubygem-bundler patch zlib-devel redhat-rpm-config openssl
 
-bundle install
+  bundle install
+# setup for debian 
+elif [ -f /etc/debian_version ]; then
+  git submodule init && git submodule update
```
mykaul: Both this line and the 'bundle install' are the same for both EL and Debian, no? Can be outside the if statement?

setup.sh
lgh2: Thank you - I will change the error message move the bundle install line to outside the if statement.

setup.sh
```diff
@@ -1,7 +1,19 @@
 #!/bin/sh
 
-git submodule init && git submodule update
+#setup for RedHat
+if [ -f /etc/redhat-release ]; then
+  git submodule init && git submodule update
```
duck-rh: As mykaul said, this is common to all distro

duck-rh:
Thanks for your contribution

PR #894, opened 2017-04-05
Blogpost411

JohnMarksRH:
Fixes issue # (delete if not relevant)

Changes proposed in this pull request:

- This is a new blog post about the 4.1.1 release

-

-

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @jmarks

This pull request needs review by: @mykaul @ykaul

PR #905, opened 2017-04-12
new commit for blog post 411

JohnMarksRH:

Changes proposed in this pull request: I have resubmitted the 4.1.1 blog post, this time with the correct filename format. I have changed the subheadings from "##" - which look the same size as the title - to "###". This time I previewed the post on middleman (. ./run-server.sh). This PR is associated with PR # 894. 

-

-

-

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @jmarks

This pull request needs review by: @bproffitt @mykaul

### DIFF
--- a/source/blog/2017-04-05-say-hello-to-ovirt-4-1-1.md
+++ b/source/blog/2017-04-05-say-hello-to-ovirt-4-1-1.md
@@ -15,30 +15,30 @@ So what's new in oVirt 4.1.1?
 
 READMORE
 
-## Storage Team
+### Storage Team
 
 * LUNs can be removed from a block data domain, provided that there is enough free space on the other domain devices to contain the data stored on the LUNs being removed.
 * Support for NFS version 4.2 connections (when supported by storage).
 
-## Integration Team
+### Integration Team
 
 * oVirt-hosted-engine-setup now works with NetworkManager enabled.
 
-## Network Team
+### Network Team
 
 * NetworkManager keeps running when a host is added to oVirt. This allows users to review networking configurations in cockpit whenever they want.
 
-## Infra Team
+### Infra Team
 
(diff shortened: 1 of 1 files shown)


duck-rh:
The link on #894 did not show up for some reason (seem there is a space after #)

JohnMarksRH:
This PR is associated with https://github.com/oVirt/ovirt-site/pull/894

duck-rh:
So, the problem with doing two things in the same PR is that things are mixed and you have to either agree on all changes to merge or block until everything is fine.

So the renaming is fine.

But the proposed change in the headers are not right AFAIK. If there is a costemic problem with level 2 headings, then the CSS should be improved. Heading levels are meant to organize titles in a tree and this breaks the semantic (and removes one possible depth level which could be useful in some pages).

So I'm in favour of staying with the current headings usage and rejecting this change, sorry.

mykaul:
We can fix the CSS later, let's get the blog entry up. 
I agree in general we should not perform several unrelated changes in a single PR of course.

duck-rh:
@mykaul, I suggested to isolate the renaming in another PR I would have merged right away, but @JohnMarksRH refused.

I think the headings change is wrong and it is unneeded to get the blog entry in anyway, so I'm not going to accept a bunch of unrelated changes with one I disagree upon.

mykaul:
You are mostly correct, but due to the holidays, I wanted the blog post to come out ASAP.

duck-rh:
Creating a new PR and approving it would have taken 2 minutes at most, so this is totally irrelevant.

The fact you have a deadline does not mean you can violate the contribution policy. This place is for open contribution on the oVirt's website, which means respect and cooperation with other contributors.
Disputed changes were slipped in for no good reason, please fix this ASAP.

Issue #952, opened 2017-05-12
middleman is not working on Fedora 26


sandrobonazzola:
### Description of the problem

middleman is not working on Fedora 26 so it will be difficult for contributors to verify the patch before submitting

### Steps to reproduce

1. git clone ovirt-site on Fedora 26

2. run setup.sh

3. Get error:

Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/sbonazzo/.gem/ruby/gems/json-1.8.3/ext/json/ext/generator
/usr/bin/ruby -r ./siteconf20170512-14061-unoh6t.rb extconf.rb
Ignoring curb-0.8.8 because its extensions are not built.  Try: gem pristine curb --version 0.8.8
Ignoring eventmachine-1.2.0.1 because its extensions are not built.  Try: gem pristine eventmachine --version 1.2.0.1
Ignoring eventmachine-1.0.7 because its extensions are not built.  Try: gem pristine eventmachine --version 1.0.7
Ignoring ffi-1.9.14 because its extensions are not built.  Try: gem pristine ffi --version 1.9.14
Ignoring ffi-1.9.8 because its extensions are not built.  Try: gem pristine ffi --version 1.9.8
Ignoring hitimes-1.2.4 because its extensions are not built.  Try: gem pristine hitimes --version 1.2.4
Ignoring hitimes-1.2.2 because its extensions are not built.  Try: gem pristine hitimes --version 1.2.2
Ignoring http_parser.rb-0.6.0 because its extensions are not built.  Try: gem pristine http_parser.rb --version 0.6.0
Ignoring json-2.0.3 because its extensions are not built.  Try: gem pristine json --version 2.0.3
Ignoring json-1.8.2 because its extensions are not built.  Try: gem pristine json --version 1.8.2
Ignoring nokogiri-1.6.8.1 because its extensions are not built.  Try: gem pristine nokogiri --version 1.6.8.1
Ignoring nokogiri-1.6.6.2 because its extensions are not built.  Try: gem pristine nokogiri --version 1.6.6.2
Ignoring oj-2.17.4 because its extensions are not built.  Try: gem pristine oj --version 2.17.4
Ignoring oj-2.12.2 because its extensions are not built.  Try: gem pristine oj --version 2.12.2
Ignoring rainbow-2.2.1 because its extensions are not built.  Try: gem pristine rainbow --version 2.2.1
[...]

PR #995, opened 2017-05-30
Update Gemfile.lock

sandrobonazzola:
Fixes issue #952 

- Update Gemfile.lock allowing to run middleman on Fedora 26

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @sandrobonazzola


duck-rh:
One possible solution would be to use SCL but the builder script would need to be adapted. This is not impossible but some more work. When @mscherer is back from vacation and all the recent changes I made are reviewed I will discuss plans with him about this feature.

mscherer:
Another solution is to use the travis docker support to run stuff on Fedora 26 and run the builder on Fedora 26 as well. I kinda fear that something we may have to do sooner or later :/

sandrobonazzola:
Not needed anymore

Issue #1000, opened 2017-06-01
Latest case study not displaying on home page


JohnMarksRH:
I added a new case study to the site.
I also added a summary of the case study to the summary page.
The case study is not displaying on the home page:
https://www.ovirt.org/#case-study

Help needed from @bproffitt @duck-rh

duck-rh:
The builder is running fine.

I tested on my laptop and it does not show up either. So something is wrong in the content. It renders an empty div…

JohnMarksRH:
@duck-rh @bproffitt Which page affects / controls the display of the case study on the home page:
The user stories page?
https://www.ovirt.org/community/user-stories/user-stories/
Or the main case study page?
https://www.ovirt.org/community/user-stories/RCC-case-study/

Thanks

duck-rh:
So first, the comment in _source/index.html.haml_ to look at _case-studies.html.md_ is wrong, it does not exist.

So it is looking for the first page having _page_classes: case-studies_ header defined in it, which leads to _source/community/user-stories/user-stories.html.md_ (the only one). It then extracts the content of the last study.

Unfortunately in the previous doc there is an empty block (just an empty line):
```
<div class="case-study">

</div>
```
So I guess the problem lies here.

This is really ugly and cryptic… At least we should fix the explanations in _source/index.html.haml_.

sandrobonazzola:
@rollandf @JohnMarksRH can we close this? I see #1008 has been merged.

PR #1008, opened 2017-06-06
Fix user story issue in home page

rollandf:
Fixes issue #1000 

- Removed empty section in users-stories.html.md

- Moved Florida State University to the end of the file in order to be seen in the home page

- Fix comment in index,html about where the user stories are located

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @rollandf

This pull request needs review by: @JohnMarksRH @duck-rh

### DIFF
--- a/source/community/user-stories/user-stories.html.md
+++ b/source/community/user-stories/user-stories.html.md
@@ -133,6 +133,3 @@ That solution would be oVirt. Today, more than 3,000 students use this virtual d
 Read the full [Universidad de Sevilla case study](/community/user-stories/universidad-de-sevilla-case-study/)
 
 </div>
-<div class="case-study">
-
-</div>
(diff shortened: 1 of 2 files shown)


duck-rh:
The changes are fine.

I think we should add a comment in _users-stories.html.md_ to explain which case is gonna be on the front page.

Also I wonder why the **last** one is selected. On the case study page this is the last, so the least visible, so it seems illogical. I would imagine the most important and recent ones to be one the top of the page, and the old ones to go down with time and at some point disappear. If people agree I think it is trivial to change the logic of the selection and I could help.

JohnMarksRH:
@rollandf @duck-rh Looks good and Duck's suggestion makes sense.

rollandf:
@duck-rh I fix according to your comment.
It makes much more sense now.

duck-rh:
works nicely, thanks :-)

JohnMarksRH:
Great!

PR #1040, opened 2017-06-27
ova feature: updating the feature page flows as discussed

danielerez:
@nirs 
@ahadas 
@mureinik


source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
-   OVA.
-6. The complete OVF specification is sent from the back end to
-   imageio-proxy. imageio-proxy sends the specification to imageio, and
-   imageio streams it into the OVA file.
-
+1. The frontend sends the backend a VM and an export destination.
+2. Backend creates new disk (for the ova) if the destination requires.
+   The backend should also prepare and teardown the disks if needed.
+3. Backend asks VDSM to create an OVA file with the complete OVF 
+   specification and export destination.
```
ahadas: does "complete" means "including size, populated size, actual size" of the disks?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
+4. VDSM runs ovirt-ova tool - creates the OVA in the destination
+   (using qemu-img convert which may include file format conversion).
```
ahadas: what about checking the possibility of streaming?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
+   
 Notes:
 
-1. Step 5 is equivalent to `qemu-img convert` with the collapse option
+1. Step 4 is equivalent to `qemu-img convert` with the collapse option
    set and the target format set to QCOW.
-2. The OVF configuration needs to be streamed to the beginning of the
-   OVA file (according to the OVA specification, the OVF should be the
-   first tar entry inside the OVA file).
```
ahadas: why is the note about the OVF being in the first entry of the tar file removed?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
-   to be set inside the OVF configuration.
-4. The OVA file will be created with a special extension (step 3). At
-   the end of the process (step 6), the file will be renamed, and the
-   extension dropped. That is done in order to enable tools that
-   synchronize shares between various deployments to differentiate
-   partial and complete files.
+2. Only after building the OVA file we could update the ovf with the
+   export size.
+3. If possible, we'll consider using streaming the disk data directly
+   from the qemu chain into OVA.
```
ahadas: This has major implication about the usability of the feature, I'm afraid.
It should be noted not as a note in this section, so it will be probably exposed. Remember that this wiki page was reviewed by PMs and other stackholders and this changes a major assumption of what it said before. I suggest to start exploring the streaming option, and if we see that it is not possible, then letting everyone know that we take this shortcut.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
+   process to the backend.
+5. The backend persists the VM into the database.
+6. The backend triggers the creation of the target VM disks on the
    storage domain.
-7. The back end sends the import parameters, including the destination
-   disks, to imageio-proxy, that sends them to imageio.
-8. Imageio reads the disk from the OVA and according to the import
-   parameters streams the data into the corresponding target disks.
+7. The backend sends the import parameters, including the destination
+   disks to vdsm.
```
ahadas: so I wonder, can we start doing that with ansible? one of the defined goals were to minimize the changes in vdsm..

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
-   imageio-proxy delivers it to imageio.
-6. Imageio starts reading the disk - converting it to QCOW - and sends
-   it to the URL. The front end then streams the data into the target
-   file.
-7. The back end sends the OVF configuation to imageio-proxy.
-   Imageio-proxy sends it to imageio, and imageio writes the OVF
-   configuration to the URL. The front end then streams the data into
-   the target file.
+1. The frontend passes the backend a VM (or template) to download.
+2. Exporting to an OVA on the host or shared storage (as describe in
```
ahadas: something is wrong in this sentence..

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
+   'Exporting a VM or Template as an OVA')
+3. Frontend displays a link for downloading the OVA.
+4. Press on the download link will start an ovirt-imageio download
+   to the browser (not implemented yet in the frontend).
```
ahadas: is that how download image going to be?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -118,41 +119,30 @@ generated by oVirt:
...
-   storage domain.
-7. The back end sends the import parameters - including the destination
-   disks - to imageio-proxy, which sends them to imageio.
-8. Imageio-proxy sends the engine a URL to write to. This URL is sent to
-   the front end.
-9. The front end streams the **whole OVA** to imageio-proxy, which
-   streams it to imageio.
-10. Imageio reads the data and according to the import parameters
-    streams the data into the corresponding target disks.
+1. The frontend retrieves the OVF configuration from the OVA.
```
nirs: Lets remove these changes (frontend/backend), they are not related to the flow change.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -118,41 +119,30 @@ generated by oVirt:
...
+2. The frontend sends the OVF configuration to the backend.
+3. The backend parses the OVF configuration and returns a VM object
+   to the frontend.
+4. The frontend sends back the VM (optionally customized) to the backend.
+5. The backend persists the VM into the database and lock it.
+6. The backend sends the list of disks metadata to the frontend,
+   which starts the process of uploading the disks.
```
nirs: No, just the list of disks to upload, the the metadata need to extract the disks from the ova.

The actual disks are created (and attached to the vm) when uploading a disk. This way if the operation fail we need to clean up only the disk that were uploaded, the disk we did not upload yet do not exit yet.

If we want to create all the disks when creating the vm, we need to the change the image upload api to upload to existing disk, which is not supported now. If there is a good reason to do this, why not.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -118,41 +119,30 @@ generated by oVirt:
...
+7. From this point the disks upload is identical the regular image
+   upload process.
```
nirs: We don't upload the whole ova, only the disks that needs to be uploaded from the ova.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -118,41 +119,30 @@ generated by oVirt:
...
-   This is the preferable format because these volumes will serve as
-   base volumes, and the RAW format delivers better performance.
-   However, the destination volumes may initially be in the COW format
-   (copied as-is from the OVA file) and only later on converted to the
-   RAW format.
+   within the OVA. In order to support this, the backend would need to
+   pass an indication to the frontend as to whether or not each disk is
+   compressed (step 6), so the ovirt-imageio-daemon can decompress the
+   data on-the-fly while streaming it to the destination volume (step 7).
+2. The upload process supports handle both raw and qcow images in the OVA.
```
nirs: We don't care about the format, we can upload anything. 

Sparseness is not supported for raw files. If the disk is mostly empty, with only some data at the start of the disk, it will work, uploading the data in the ova, and leaving empty space at the end of the file/block device. If there are holes in the file, they will be uploaded as zeros. If the file is compressed, sparseness should not matter.

I think we should use qcow2 format when creating ova, since they support sparseness, but for upload we can upload whatever the user like to upload.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
 ## Exporting a VM or Template as an OVA
 The user will be able to select a VM (or a template) to be exported as
 an OVA file. In such cases, the user needs to specify a path on a host
-where the OVA file will be created.
+or a disk on shared storage where the OVA file will be created.
```
nirs: It is important that we do not base the export feature on having enough space on the host, or having to mount manually on the host. We should support these options but not require them.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
-4. The disk is streamed from the back end to imageio-proxy, and then to
-   imageio.
-5. Imageio reads the disk and streams it into a QCOW volume within the
-   OVA.
-6. The complete OVF specification is sent from the back end to
-   imageio-proxy. imageio-proxy sends the specification to imageio, and
-   imageio streams it into the OVA file.
-
+1. The frontend sends the backend a VM and an export destination.
+2. Backend creates new disk (for the ova) if the destination requires.
```
nirs: Right, creating disks is done by vdsm,  to be more specific, the spm.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
+   The backend should also prepare and teardown the disks if needed.
```
nirs: The one that we write to?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: Whatever info needed to convert the ovirt disk to files in the ova and write it to the destination.

Since we convert existing disks, all the information about the disks can be extracted from the disks, including virtual size, and format. The actual size in the ova is know only after a disk was converted to the destination format, maybe compressed.

I think we need to add that the ova tool will update the off if needed (e.g. disk size in the tar file)

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: I checked qemu-img commands, there is no option to stream content from the existing tools.

There may be a way to stream to nbd server - in this case we need to implement nbd server that qemu-img convert will stream the data. This server can write the received data to the ova. Not sure how feasible is this solution, but if we have it, we can use it for both export and download.

See https://github.com/libguestfs/nbdkit

I don't think we can promise anything about streaming at this point.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: Isn't this required by the ova format? I don't see a reason to repeat the details here, unless it make the flow more clear.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: We cannot promise anything about streaming now.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: I don't see any reason to use ansible for this flow. Uploading/downloading images require two components: vdsm for control, imageio for doing the io. The suggested vdsm change is very minimal, I don't see any problem with it.

I think our goal is to create a reusable component, that we can run later on openshift as a job without engine or vdsm. Sending a request to vdsm and starting the tool from vdsm is the minimal change to make this work with the current system, the tool itself should be able to work without change given right options.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: This means that engine will trigger the export flow to create an ova file on some destination.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: but AFAIR, we support raw+sparse on file storage..

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: If we don't have a way to stream data from the disks, yes, we need to create the ova somewhere, and then we can download it. If we find a way to stream the data, we will have to rethink the flow.

Daniel, I think we should add a note that this solution allows resuming a download at any point, while a streaming solution will most likely require starting the entire export/convert/download from scratch.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: but you know, maybe by not saying that we convert any format to any format we then can promise streaming, at least on upload, because then we can avoid using qemu-img commands. The compression should not be an issue since the OVA specification was designed for being consumed by streaming, so that part (compression) should be covered.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: The url is created by the backend, and the format is always /images/ticket-uuid.

The daemon has is configured with a ticket with the same uuid, specifying the path that this upload should write to, the size that can be written, etc.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: We support raw sparse, but we cannot preserve sparseness when uploading/downloading files in raw format. The holes in the files are converted to zeros, and libvirt is doing the same when you download a volume. This is why qcow2 format is better for transferring disks.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: We can stream data on upload - Daniel, maybe add a note abut this is the upload flow?

We can stream data directly from vm images if we don't change the format, but this will work only for raw disk or for qcow2 disks without snapshots. Do you think these use case are important enough to write special code for them?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: ack

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: right, and that's probably the reason vmware are using vmdk and not raw images as well. we can support raw, but because of this argument that you just wrote, we should promote using qcow2 when exporting the vm

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: yeah, of course. take into consideration that there are no snapshots within OVAs, the specification doesn't support it, and that it is extremely important to minimize the time of the upload. Uploading the OVA efficiently (from the duration perspective) will be the difference between a feature that 'is working' and a killer-feature.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -75,10 +75,11 @@ improved portability. By being aligned with the common OVA specification
 (rather than our unique structure of a storage domain), we achieve
 better compatibility with other platforms.
 
-The OVA file would be streamed into imageio proxy, enabling its creation
-on a variety of locations (path on host in the data-center, NFS share,
-etc). In addition, it will include neither snapshots nor the template
-that the VM may be based on.
+The OVA file would be stored on a variety of locations: path on host 
+in the data-center, NFS share, etc. We might consider to build the ova
```
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -118,41 +119,30 @@ generated by oVirt:
...
-   disks - to imageio-proxy, which sends them to imageio.
-8. Imageio-proxy sends the engine a URL to write to. This URL is sent to
-   the front end.
-9. The front end streams the **whole OVA** to imageio-proxy, which
-   streams it to imageio.
-10. Imageio reads the data and according to the import parameters
-    streams the data into the corresponding target disks.
+1. The frontend retrieves the OVF configuration from the OVA.
+2. The frontend sends the OVF configuration to the backend.
+3. The backend parses the OVF configuration and returns a VM object
```
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: the note still exists in 'Uploading an OVA Generated by an External Source' section: "2. The complete OVA file is streamed into the disk that has just been created."

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,81 +170,71 @@ Note:
...
-3. Imageio creates the OVA file.
-4. The disk is streamed from the back end to imageio-proxy, and then to
-   imageio.
-5. Imageio reads the disk and streams it into a QCOW volume within the
-   OVA.
-6. The complete OVF specification is sent from the back end to
-   imageio-proxy. imageio-proxy sends the specification to imageio, and
-   imageio streams it into the OVA file.
-
+1. The frontend sends the backend a VM and an export destination.
```
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: done

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: fixed

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: Nir's still checking streaming. Will update.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: according to the OVA format, the OVF should be the first member so the note seemed obvious.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
danielerez: fixed the sentence to be clearer

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -120,39 +121,30 @@ generated by oVirt:
...
-   the front end.
-9. The front end streams the **whole OVA** to imageio-proxy, which
-   streams it to imageio.
-10. Imageio reads the data and according to the import parameters
-    streams the data into the corresponding target disks.
+3. The back end parses the OVF configuration and returns a VM object
+   to the front end.
+4. The front end sends back the VM (optionally customized) to the back end.
+5. The back end persists the VM into the database and lock it.
+6. The back end sends the list of disks metadata to the front end,
```
ahadas: so shouldn't we create the volumes on the storage domain in advance (what was step 6 before)? imageio will do that? (if it does then we need to specify the actual size of the disk from the front end, right?)

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -120,39 +121,30 @@ generated by oVirt:
...
-   This is the preferable format because these volumes will serve as
-   base volumes, and the RAW format delivers better performance.
-   However, the destination volumes may initially be in the COW format
-   (copied as-is from the OVA file) and only later on converted to the
-   RAW format.
+   pass an indication to the front end as to whether or not each disk is
+   compressed (step 6), so the ovirt-imageio-daemon can decompress the
+   data on-the-fly while streaming it to the destination volume (step 7).
+2. The upload process supports handling both raw and qcow images in the OVA.
+3. Currently, we support resuming a single image upload. We should consider
```
ahadas: We should consider -> TBD:

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: right right, I wasn't suggesting to focus only on the local filesystem on the host. my point was whether we treat shared storage as yet another path on the host - and by that assume that it was already mounted to that path. because otherwise we would need to make a relatively big change in the import dialog (for getting the details, including the authentication details, of the NFS share) and that is something we didn't plan to do.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,40 +172,32 @@ Note:
...
-3. Imageio creates the OVA file.
-4. The disk is streamed from the back end to imageio-proxy, and then to
-   imageio.
-5. Imageio reads the disk and streams it into a QCOW volume within the
-   OVA.
-6. The complete OVF specification is sent from the back end to
-   imageio-proxy. imageio-proxy sends the specification to imageio, and
-   imageio streams it into the OVA file.
-
+1. The front end sends the back end a VM and an export destination.
```
ahadas: same as above, would "export destination" mean that we copy much of the new/import data/export/iso domain dialog?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,40 +172,32 @@ Note:
...
+2. The back end creates new disk (for the ova) if the destination requires,
+   and prepares the VM source disks.
+3. The back end asks VDSM to create an OVA file with the complete OVF
```
ahadas: you explains it afterwards - back end cannot provide the complete OVF at that stage since it doesn't know the size of the converted disks. so maybe "complete OVF" -> "complete OVF" with placeholders to be replaced later with the actual size of the disks"?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: ok, indeed - it was important to mention before because it had an impact on the design of the process

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -180,40 +172,32 @@ Note:
...
-   first tar entry inside the OVA file).
-3. The back end should be provided with the overall amount of data
-   written to the disk entry in step 4, because this information needs
-   to be set inside the OVF configuration.
-4. The OVA file will be created with a special extension (step 3). At
-   the end of the process (step 6), the file will be renamed, and the
-   extension dropped. That is done in order to enable tools that
-   synchronize shares between various deployments to differentiate
-   partial and complete files.
+2. Only after building the OVA file we could update the ovf with the
```
ahadas: just for consistency - OVF

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
ahadas: I'm not asking to promise something you cannot be sure in being able to deliver. But let's say that explicitly. Currently is seems like "well, if we can do it efficiently (= if possible), we will think about doing that (= we'll consider)" - it doesn't sound good (because if it is supported, we surely need to leverage that). So instead, we can say "The qemu tools cannot be utilized for reading the disks as a stream. TBD: enable reading the disks as a stream".

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -243,18 +228,16 @@ theory, it should look like this:
...
+2. The back end starts the exporting to OVA flow (as described in 
+   'Exporting a VM or Template as an OVA')
+3. The front end displays a link for downloading the OVA.
+4. Press on the download link will start an ovirt-imageio download
+   to the browser (not implemented yet in the front end).
+
+Notes:
+
+1. If possible, we'll consider using streaming the disk data directly
+   from the qemu chain to the browser.
```
ahadas: again, I'm sure that you agree that if it is possible then we should leverage that. let's rephrase - it's not possible so we don't do that at the moment, TBD: to allow that.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -243,18 +228,16 @@ theory, it should look like this:
...
-5. The back end passes the disk to imageio-proxy to copy, and
-   imageio-proxy delivers it to imageio.
-6. Imageio starts reading the disk - converting it to QCOW - and sends
-   it to the URL. The front end then streams the data into the target
-   file.
-7. The back end sends the OVF configuation to imageio-proxy.
-   Imageio-proxy sends it to imageio, and imageio writes the OVF
-   configuration to the URL. The front end then streams the data into
-   the target file.
+2. The back end starts the exporting to OVA flow (as described in
```
ahadas: "starts the export" or "starts exporting"

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -243,18 +228,16 @@ theory, it should look like this:
...
-6. Imageio starts reading the disk - converting it to QCOW - and sends
-   it to the URL. The front end then streams the data into the target
-   file.
-7. The back end sends the OVF configuation to imageio-proxy.
-   Imageio-proxy sends it to imageio, and imageio writes the OVF
-   configuration to the URL. The front end then streams the data into
-   the target file.
+2. The back end starts the exporting to OVA flow (as described in 
+   'Exporting a VM or Template as an OVA')
+3. The front end displays a link for downloading the OVA.
```
ahadas: unlike image download, the download link will not be available immediately since it may (and probably will) take long time to generate the OVA, right? so how do you see this - I select "download OVA", there is a long async operation and at the end there will be a popup with the link? or would we display the link immediately as disabled and enable it only at the end of the async operation?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
```diff
@@ -243,18 +228,16 @@ theory, it should look like this:
...
+4. Press on the download link will start an ovirt-imageio download
+   to the browser (not implemented yet in the front end).
+
+Notes:
+
+1. If possible, we'll consider using streaming the disk data directly
+   from the qemu chain to the browser.
+2. The flow should be supported in the sdk and future automation using
+   Ansible.
```
ahadas: some would say that we should even start with the sdk since it is more important that integrating it with the current UI, but ok..

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: :+1:

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: I agree, a section about streaming would make this clear.

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: :+1:

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: It think it can work like this:
- user starts download ova operation
- system show progress of the operation with a description like "exporting vm..."
- when the export is done, system show "download  ova" button
- user clicks the button, browser show "save as" dialog

But let's discuss this with Eldan?

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: :+1:

source/develop/release-management/features/virt/enhance-import-export-with-ova.md
nirs: Starting with the sdk is a good idea, and work on the ui later. This will allow QE to write tests for this feature early.

ahadas:
@mykaul , @bproffitt , please merge

PR #1071, opened 2017-07-12
purge all wiki related headers

duck-rh:
wiki redirects and conversion/progress pages are gone too.

Fixes issue discussed in #966 

I confirm that this pull request was submitted according to the [contribution guidelines](/CONTRIBUTING.md): @duck-rh

This pull request needs review by: (please @rbowen @apevec)

Please note I also removes quite some magic, like wiki redirects (as they need the wiki_title).


duck-rh:
so, this is a mistake, I was tired and worked on the wrong website :-)
Nevertheless I do think the same arguments apply and we should do this cleanup (especially if we want to migrate to another system one day).

@mykaul @JohnMarksRH and others, what do you think?

apevec:
This was meant for https://github.com/redhat-openstack/website/pull/966#discussion_r126956495

duck-rh:
thanks @apevec , my bad. I'm preparing the one for RDO don't worry.

as things are similar I think we should do the same and get rid of a long past. On oVirt there are special redirects which seem not to be present for RDO, but that's old and much of the content has changed, so I think it's not really relevant. The search engine also is using the old wiki names.

I'd like oVirt people's opinion for this one.

rbowen:
On 07/12/2017 10:18 PM, Marc Dequènes (Duck) wrote:
> so, this is a mistake, I was tired and worked on the wrong website :-)
> Nevertheless I do think the same arguments apply and we should do this
> cleanup (especially if we want to migrate to another system one day).
> 
> @mykaul <https://github.com/mykaul> @JohnMarksRH
> <https://github.com/johnmarksrh> and others, what do you think?


Well, it needs to be done for oVirt, too, so no wasted time! :-)


-- 
Rich Bowen - rbowen@rcbowen.com - @rbowen
http://apachecon.com/ - @apachecon

duck-rh:
True, but it was not discussed beforehand. Also in oVirt there are those redirects and search by wiki names they need to grieve. I think it's useless by now but I'm not a writer or a user so I can't be sure.

PR #1169, opened 2017-08-23
Add design for listing VMs pinned to a host.

akrejcir:
A design page for showing VMs pinned to a host.

We should discuss:
- what is a good way to expose the new query through REST API
- what is a good way to show the pinned VMs in the webadmin

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @akrejcir

This pull request needs review by: @jhernand  @pcbailey


akrejcir:
@vojtechszocs @sjd78

vojtechszocs:
For WebAdmin, either showing a separate list of pinned VMs (right below the list of running VMs) within "Host / Virtual Machines", or having a separate "Host / Pinned VMs" detail tab, sounds reasonable. I'd personally prefer the first of these two, since the available space is much bigger now (no more "main tab" stuff showing above).

pcbailey:
I agree 100% with @vojtechszocs. I think it best to keep the VM-related information together in the same sub tab, but as you pointed out, mixing the pinned VMs and those running on the host could get confusing. A separate list would avoid that problem while keeping the information in the Virtual Machines sub tab.

akrejcir:
I have added an image of how the two lists would look on one page.

doron-fediuck:
@vojtechszocs @pcbailey any objections to merge?

pcbailey:
No objections from me.

PR #1211, opened 2017-09-17
Add Multipath events feature

rollandf:

- Feature page for Multipath events

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @rollandf

This pull request needs review by: @nirs @mureinik

### DIFF
--- a/source/develop/release-management/features/storage/multipath-events.html.md
+++ b/source/develop/release-management/features/storage/multipath-events.html.md
@@ -1,21 +1,35 @@
 ---
-title: Multipath Events
+title: Multipath Alerts
 category: feature
 authors: Fred Rolland
-feature_name: Multipath Events
+feature_name: Multipath Alerts
 feature_modules: engine,vdsm
 feature_status: Design
 ---
 
-# Multipath Events
+# Multipath Alerts
 
 
 ## Overview
 
-This feature adds the ability to display events to the user in
-the engine UI when multipath device path changes its status.
-The supported events are when a path moves to down or up state.
-With these events, the user will be able to detect storage issues faster.
+This feature adds the ability to display alerts to the user in
+the engine UI when some multipath device have faulty paths, when
+some multipath devices are non-operational because all paths became faulty,
+and when all multipath devices on a host recover.
+
+The purpose of this feature is to help users troubleshoot storage issues
+before they become critical when a multipath device is lost, and to
+make it easier to troubleshoot cases when some LUNs are not available
+because all paths became faulty.
+
+Currently, the user has no indication in oVirt of faulty paths.
+As a result only when all paths are faulty and the access to the storage
+domain is not working, the user will get an indication:
+- If the all the hosts in the Data Center are reporting a problem in the 
+storage domain. the storage domain status that will become 'Inactive'.
+- If only one host reports a problem in the storage domain, the status
+of the host will become 'Non Operational'.
+
 
 
 ## Owner
@@ -29,65 +43,72 @@ With these events, the user will be able to detect storage issues faster.
 
 ### How it works
 
-When a host is connected to a DC (Storage Pool), Vdsm will register
-to udev multipath events. For each event, a message will be sent to Engine
-with the details of the event. Engine will create an audit log event. 
-Once the host is disconnected from the DC, Vdsm will unregister from
-the 'udev' events.
+Vdsm will register to udev multipath events during storage initialization.
+Vdsm will also build a map of all the faulty paths. 
+For each event, Vdsm will update the list, according to the event.
+The details of the faulty paths will be part of the statistics provided by
+'Host.getStats' verb.
+
+Engine polls the hosts for the statistics every 15 seconds.
+Engine creates an audit log event according to the information provided
+in the statistics. Only if there are changes in the faulty paths report,
+an event will be created.
 
 
 ### New vdsm API
 
-New multipath event type
+New property in 'HostStats': 'MultipathAlertsMap' which is a mapping of faulty paths
+indexed by the device mapper GUID.
 
 
-#### Multipath Event
+#### Multipath Alerts Map
 
-An event with information about the multipath device, path and status.
+The key of the map is the device mapper GUID.
+The value is a MultipathAlert that contains:
+- List of faulty paths.
+- Number of valid paths
 
-A mapping with these keys:
-- type: Path Failed/Path Reinstated (DM_ACTION)
-- device (GUID): unique GUID of this device (DM_NAME)
-- valid_paths - number of valid paths (DM_NR_VALID_PATHS)
-- seqnum - event sequence number, may be used to detect missed messages later (DM_SEQNUM)
+### Vdsm multipath event listener
 
+Introduce a new Vdsm internal service for listening to multipath events:
 
-### Vdsm multipath event monitoring
-
-Introduce a new Vdsm internal service for handling events:
-
-- Code interested in receiving mutlipath events will subscribe a callback to receive the event
+- Code interested in receiving multipath events will subscribe a callback to receive the event
 - When an event is received, all subscribers will receive the events (in another thread).
 - If nobody is registered for events, the events are dropped
-- The multipath event monitor will be started when connecting to the pool
-- The multipath event monitor will be stopped when disconnecting from the pool
+- The multipath event monitor will be started during storage initialization
 
 
-A Udev event monitor is introduced, listening to device-mapper udev events.
-The relevant events for this feature are PATH_FAILED and PATH_REINSTATED:
+The service will be implemented using 'pyudev', listening to these events:
+- action=change,DM_ACTION=PATH_FAILED
+- action=change,DM_ACTION=PATH_REINSTATED
+- action=remove (for multipath devices)
 
-Variable Name: DM_ACTION
-Uevent Action(s): KOBJ_CHANGE
-Type: string
-Description:
-Value: Device-mapper specific action that caused the uevent action.
-	PATH_FAILED - A path has failed.
-	PATH_REINSTATED - A path has been reinstated.
+We don't know yet how to detect multipath devices events efficiently for detecting removed devices.
 
-- The Vdsm will suscribe the monitor to events during startup.
-- The Vdsm will start the Udev event monitor when connecting to the storage pool.
-- The listener will send an event to the engine when its callback is called.
-- The Vdsm will unregister the listener to the Udev event monitor when 
-disconnecting to the storage pool.
+
 - The user can disable this feature in a specific host in Vdsm configuration file
 
 In the future, the monitor can be used for other code, for example maintaining LVM filter.
 
+### Vdsm multipath health monitor
+
+To report multipath health status to engine, we will add a multipath health monitor.
+
+- The monitor subscribes to multipath events during startup
+- The monitor builds a map of all the faulty paths during startup.
+- The monitor updates the data set of the faulty paths when its callback is called:
+    - PATH_FAILED - Add the faulty path to the map according to the device GUID
+    - PATH_REINSTATED - Remove the faulty path to the map according to the device GUID
+    - Multipath device removed  - remove the device from the map
+- Multipath health status is provided to engine via 'Host.getStats' verb.
+
+
 ### New engine API
 
 New audit logs are added:
-- Multipath device path up
-- Multipath device path down
+- Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ...
+- Devices without valid paths on host "HOST_NAME" : "GUID", "GUID" ...
+- No faulty multipath paths on host "HOST_NAME"
 
 ## Benefit to oVirt
 
@@ -97,6 +118,18 @@ New audit logs are added:
 ## User Experience
 
 New audit logs are added with information on the device path status.
+The logs are per host.
+Since the length of the information available in the audit log is limited,
+the full information of the failing paths will be available in the logs.
+The events are not generated on "real-time", but are based on polling of
+the statistics. Only if a change in the data received is detected, a new
+audit log will be created
+
+If a single LUN goes down on all the hosts, the user will get an alarm
+for each host in the DC.
+
+If a single path goes down on one of the hosts, the user will get an alarm
+only for this host.
 
 
 ## Installation/Upgrade
@@ -106,14 +139,15 @@ This feature does not affect engine or Vdsm installation or upgrades
 
 ## User work-flows
 
-- Once a user connects to an ISCSI server, 'Multipath device path up' event
-will be created.
+- Once a user connects to a new ISCSI server, no events will be created.
 
-- If there is an issue with the connection to the storage server, and the 
-multipath reports an issue with one or more device path,
-'Multipath device path down' will be created.
+- If there is an issue with the storage server, and the 
+multipath reports an issue with one or more device path, a "Faulty multipath paths on host" will be created
 
-- Once the issue is fixed, 'Multipath device path up' event will be created.
+- Once the issue is fixed, and all paths are up a "No faulty multipath paths on host "HOST_NAME"
+event will be created
+
+- If only part of the paths are fixed, a "Faulty multipath paths on host" will be created with the details of the devices.
 
 
 ## Dependencies / Related Features
@@ -135,32 +169,27 @@ NA
 
 ## Event Reporting
 
-- 'Multipath device path up'
-- 'Multipath device path down'
-
+- Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ... 
+- No faulty multipath paths on host "HOST_NAME"
+- Devices without valid paths on host "HOST_NAME" : "GUID", "GUID" ...
 
 ## Documentation / External references
 
 This feature is required for resolving
 [Bug 723931](https://bugzilla.redhat.com/723931)
 
-Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
+Pyudev - [pyudev documentation](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
 
-Uevent [example](https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt)
+Uevent [device mapper events kernel documentation](https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt)
 
 ## Testing
 
-- Connect to a new ISCSI storage server
-  - For each multipath device path a 'Multipath device path up' is triggered
-
-
 ### Negative flows
 
-- Block network to ISCSI storage server
-  - For each multipath device path a 'Multipath device path down' is triggered
-- Unblock access to iscsi server
-  - Wait until all paths go up. There should be 'Multipath device path up' for 
-    every path on this iscsi server
+- Block network to ISCSI storage server on a specific host
+  - 'Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ...' is triggered
+- Unblock access to ISCSI server on a specific host
+  - Wait until all paths go up. a 'No faulty multipath paths on host "HOST_NAME"' is triggered
 
 ## Release Notes
 
@@ -168,7 +197,5 @@ TBD
 
 ## Open Issues
 
-- Connecting to a new ISCSI storage server will trigger a lot of events,
-that are not useful to the user.
 - In a scale environment, issue on the storage server or network can trigger
-an event storm.
+a lot of events. (Note that the events are generated per hosts and not per path).
--- a/source/develop/release-management/features/storage/multipath-events.html.md
+++ b/source/develop/release-management/features/storage/multipath-events.html.md
@@ -31,7 +31,6 @@ storage domain. the storage domain status that will become 'Inactive'.
 of the host will become 'Non Operational'.
 
 
-
 ## Owner
 
 - Name: [Fred Rolland](https://github.com/rollandf)
@@ -45,7 +44,7 @@ of the host will become 'Non Operational'.
 
 Vdsm will register to udev multipath events during storage initialization.
 Vdsm will also build a map of all the faulty paths. 
-For each event, Vdsm will update the list, according to the event.
+For each event, Vdsm will update the map, according to the event.
 The details of the faulty paths will be part of the statistics provided by
 'Host.getStats' verb.
 
@@ -57,8 +56,7 @@ an event will be created.
 
 ### New vdsm API
 
-New property in 'HostStats': 'MultipathAlertsMap' which is a mapping of faulty paths
-indexed by the device mapper GUID.
+Host.getStats will report now new 'multipath' key. The contents is a map of multipath alerts.
 
 
 #### Multipath Alerts Map
@@ -137,17 +135,17 @@ only for this host.
 This feature does not affect engine or Vdsm installation or upgrades
 
 
-## User work-flows
+## User workflows
 
 - Once a user connects to a new ISCSI server, no events will be created.
 
-- If there is an issue with the storage server, and the 
-multipath reports an issue with one or more device path, a "Faulty multipath paths on host" will be created
+- If there is an issue with the storage server, and multipath reports
+an issue with one or more device path, a "Faulty multipath paths on host" event will be created
 
 - Once the issue is fixed, and all paths are up a "No faulty multipath paths on host "HOST_NAME"
 event will be created
 
-- If only part of the paths are fixed, a "Faulty multipath paths on host" will be created with the details of the devices.
+- If only part of the paths are fixed, a "Faulty multipath paths on host" event will be created with the details of the devices.
 
 
 ## Dependencies / Related Features
--- a/source/develop/release-management/features/storage/multipath-events.html.md
+++ b/source/develop/release-management/features/storage/multipath-events.html.md
@@ -1,0 +1,174 @@
+---
+title: Multipath Events
+category: feature
+authors: Fred Rolland
+feature_name: Multipath Events
+feature_modules: engine,vdsm
+feature_status: Design
+---
+
+# Multipath Events
+
+
+## Overview
+
+This feature adds the ability to display events to the user in
+the engine UI when multipath device path changes its status.
+The supported events are when a path moves to down or up state.
+With these events, the user will be able to detect storage issues faster.
+
+
+## Owner
+
+- Name: [Fred Rolland](https://github.com/rollandf)
+- Email: <frolland@redhat.com>
+
+
+## Detailed Description
+
+
+### How it works
+
+When a host is connected to a DC (Storage Pool), Vdsm will register
+to udev multipath events. For each event, a message will be sent to Engine
+with the details of the event. Engine will create an audit log event. 
+Once the host is disconnected from the DC, Vdsm will unregister from
+the 'udev' events.
+
+
+### New vdsm API
+
+New multipath event type
+
+
+#### Multipath Event
+
+An event with information about the multipath device, path and status.
+
+A mapping with these keys:
+- type: Path Failed/Path Reinstated (DM_ACTION)
+- device (GUID): unique GUID of this device (DM_NAME)
+- valid_paths - number of valid paths (DM_NR_VALID_PATHS)
+- seqnum - event sequence number, may be used to detect missed messages later (DM_SEQNUM)
+
+
+### Vdsm multipath event monitoring
+
+Introduce a new Vdsm internal service for handling events:
+
+- Code interested in receiving mutlipath events will subscribe a callback to receive the event
+- When an event is received, all subscribers will receive the events (in another thread).
+- If nobody is registered for events, the events are dropped
+- The multipath event monitor will be started when connecting to the pool
+- The multipath event monitor will be stopped when disconnecting from the pool
+
+
+A Udev event monitor is introduced, listening to device-mapper udev events.
+The relevant events for this feature are PATH_FAILED and PATH_REINSTATED:
+
+Variable Name: DM_ACTION
+Uevent Action(s): KOBJ_CHANGE
+Type: string
+Description:
+Value: Device-mapper specific action that caused the uevent action.
+	PATH_FAILED - A path has failed.
+	PATH_REINSTATED - A path has been reinstated.
+
+- The Vdsm will suscribe the monitor to events during startup.
+- The Vdsm will start the Udev event monitor when connecting to the storage pool.
+- The listener will send an event to the engine when its callback is called.
+- The Vdsm will unregister the listener to the Udev event monitor when 
+disconnecting to the storage pool.
+- The user can disable this feature in a specific host in Vdsm configuration file
+
+In the future, the monitor can be used for other code, for example maintaining LVM filter.
+
+### New engine API
+
+New audit logs are added:
+- Multipath device path up
+- Multipath device path down
+
+## Benefit to oVirt
+
+- The user will be able to detect storage issues faster.
+
+
+## User Experience
+
+New audit logs are added with information on the device path status.
+
+
+## Installation/Upgrade
+
+This feature does not affect engine or Vdsm installation or upgrades
+
+
+## User work-flows
+
+- Once a user connects to an ISCSI server, 'Multipath device path up' event
+will be created.
+
+- If there is an issue with the connection to the storage server, and the 
+multipath reports an issue with one or more device path,
+'Multipath device path down' will be created.
+
+- Once the issue is fixed, 'Multipath device path up' event will be created.
+
+
+## Dependencies / Related Features
+
+Additional package dependency in Vdsm:
+- pyudev
+
+No related features.
+
+
+### Entity Description
+
+NA
+
+## CRUD
+
+NA
+
+
+## Event Reporting
+
+- 'Multipath device path up'
+- 'Multipath device path down'
+
+
+## Documentation / External references
+
+This feature is required for resolving
+[Bug 723931](https://bugzilla.redhat.com/723931)
+
+Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
+
+Uevent [example](https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt)
+
+## Testing
+
+- Connect to a new ISCSI storage server
+  - For each multipath device path a 'Multipath device path up' is triggered
+
+
+### Negative flows
+
+- Block network to ISCSI storage server
+  - For each multipath device path a 'Multipath device path down' is triggered
+- Unblock access to iscsi server
+  - Wait until all paths go up. There should be 'Multipath device path up' for 
+    every path on this iscsi server
+
+## Release Notes
+
+TBD
+
+## Open Issues
+
+- Connecting to a new ISCSI storage server will trigger a lot of events,
+that are not useful to the user.
+- In a scale environment, issue on the storage server or network can trigger
+an event storm.


source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+
+An event with information about the multipath device, path and status.
+
+A mapping with these keys:
+- Type: Path Failed/Path Reinstated (DM_ACTION)
+- Device (GUID): unique GUID of this device (DM_NAME)
+- TBD
+
+
+### Vdsm event monitoring
```
nirs: Vdsm multipath event monitoring

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+
+A new framework in Vdsm is introduced to register to events:
```
nirs: Introduce a new Vdsm internal service for handling multipath events:

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+- A listener can register to a topic, providing a callback method/object.
```
nirs: Do we need a topic? I think a listener would simply register itself for receiving multipath events.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+- A topic monitor will be created with the first listener.
+- Once the monitor receives an event, it will call the listener callback
+(in a different thread than the one the listener registered)
+- A listener can unregister from a topic.
+- The topic monitor will stop monitoring once the last listener unregister.
```
nirs: I think we can simplify this to:

- Code interested in receiving mutliapth events will subscribe a callback to receive the event
- when an event is received, all subscribers will receive the events (in another thread).
- If nobody is registered for events, the events are dropped
- The multipath event montior will be started when connecting to the pool
- The multipath event monitor will be stopped when disconnecting from the pool

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+Value: Device-mapper specific action that caused the uevent action.
+	PATH_FAILED - A path has failed.
+	PATH_REINSTATED - A path has been reinstated.
+
+- The Vdsm will register a listener to the Udev event monitor when connecting
+to the storage pool.
+- The listener will send an event to the engine when its callback is called.
+- The Vdsm will unregister the listener to the Udev event monitor when 
+disconnecting to the storage pool.
+- The user can disable this feature in a specific host in Vdsm configuration file
```
nirs: The previous paragraph describe this monitor, we should merge the paragraphs.

I think we should do:

- Start the montior when connecting to storage, stop when disconnecting
- Subscribe to events during startup, this is one time operation

We need a new section about future use of the monitor by registering for multipath events in other code, for example maintaining lvm filter.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+## Installation/Upgrade
+
+This feature does not affect engine or Vdsm installation or upgrades
+
+
+## User work-flows
+
+Once a user connects to an ISCSI server, 'Multipath device path up' event
+will be created.
+If there is an issue with the connection to the storage server, and the
```
nirs: We will not get down events in this case, since there are no devices yet.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+- 'Multipath device path up'
+- 'Multipath device path down'
+
+
+## Documentation / External references
+
+This feature is required for resolving
+[Bug 723931](https://bugzilla.redhat.com/723931)
+
+Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
```
nirs: Please add also https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+
+- Connect to a new ISCSI storage server
+  - For each multipath device path a 'Multipath device path up' is triggered
+
+
+### Negative flows
+
+- Block network to ISCSI storage server
+  - For each multipath device path a 'Multipath device path down' is triggered
+
```
nirs: Lets add the recovery flow:

1. Block access to iscsi server
2. Wait util all paths go down
3. Unblock access to iscsi server
4. Wait until all paths go up
5. There should be "multipath device path up" for every path on this iscsi server

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Release Notes
+
+TBD
+
+## Open Issues
+
+- Connecting to a new ISCSI storage server will trigger a lot of events,
+that are not useful to the user.
```
nirs: I think we can avoid the noise by ignoring device up events if the device was not in down state before.

This may be needed on the vdsm side to avoid sending tons uninteresting messages to engine when connecting to new storage, and on engine side to avoid the spam of connect messages from multiple servers.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+- In a scale environment, issue on the storage server or network can trigger
+an event storm.
```
nirs: Maybe we should ignore messages for a device if they were already reported on other hosts.

For example:

1. Device X on host A path went down, 3 paths active of 4 - log
2. Device X on host B path went down, 3 paths active of 4 - ignore (already reported)
3. Device X on host C path went down, 0 paths active of 4 - log - device cannot be used now 
4. Device X on host A path went down, 0 paths active of 4 - log - ignore (already reported)

We can keep a memory mapping of all devices, report changes when the event changes the mapping.

I think we have several priorities of events:
- info: number of valid paths increased on some host
- warnings: number of valid path decrease, but the device is still usable
- error: number valid paths is 0 on some host - the device is not usable or will be unusable soon (depending on multipath configuration)

We assume that all host must see all paths of all devices, so we can focus on the device state instead of the host state. We are already handling host state via storage monitoring.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+New multipath event type
+
+
+#### Multipath Event
+
+An event with information about the multipath device, path and status.
+
+A mapping with these keys:
+- Type: Path Failed/Path Reinstated (DM_ACTION)
+- Device (GUID): unique GUID of this device (DM_NAME)
```
nirs: Lets add:
- valid_paths - number of valid paths (DM_NR_VALID_PATHS)
- seqnum - event sequence number, may be used to detect missed messages later (DM_SEQNUM)

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+
+New multipath event type
+
+
+#### Multipath Event
+
+An event with information about the multipath device, path and status.
+
+A mapping with these keys:
+- Type: Path Failed/Path Reinstated (DM_ACTION)
```
nirs: Lets use lower case names, we are not going to provide key named "Type". We can use \```type``` to make it more clear.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+the 'udev' events.
+
+
+### New vdsm API
+
+New multipath event type
+
+
+#### Multipath Event
+
```
nirs: Event also have a topic, something like ```multipath|status```.

Maybe we should include the multiapth device guid in the event - will allow code to register events to specific multipath device on engine side.

nirs:
@pkliczewski, can you review this feature? We are concerned about sending many events to engine when connecting to storage.

pkliczewski:
@nirs When I tested jsonrpc code I saw that throughput tests (between two hosts in local network) showed ability to deliver 40k messages per second without any content. How many messages do you expect?

source/develop/release-management/features/storage/multipath-events.html.md
mykaul: guid = WWN/WWPN, I assume?

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,167 @@
...
+---
+
+# Multipath Events
+
+
+## Overview
+
+This feature adds the ability to display events to the user in
+the engine UI when multipath device path changes its status.
+The supported events are when a path moves to down or up state.
```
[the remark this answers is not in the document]
rollandf: I added: "With these events, the user will be able to detect storage issues faster."

source/develop/release-management/features/storage/multipath-events.html.md
mykaul: Do we listen to events of multipath devices that we have not connected/disconnected? (I think we should)

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Sure we can. What is the right field ? 
https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
mykaul: I'm not sure the assumption that all hosts must see all paths is a valid assumption. They may be connected differently, via different number of initiators or to different number of targets. (In fact, this is a good 'soft' warning to the user - which he may be OK with or act upon if it wasn't intended) - but this is a bit too complex for this implementation ATM.
I would also not drop any alert - imagine you've put one switch down for maintenance, which affected (as you anticipated, host A. You really do wish to know if anything bad happened to host B, which was on a different switch that was not supposed to go down!

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: it is here (below)

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Fixed identation,

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: How can you know if the path was not down ? I don't think we are tracking it anywhere.

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: This kind of logic can be misleading. You can think that host B is OK by ignoring it.
Let's start with keeping it simple and see how it is a real issue.

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: @mykaul  yes, we will listen to events on all multipath devices

source/develop/release-management/features/storage/multipath-events.html.md
nirs: guid is the last path component in the multipath device path /dev/mapper/xxxyyy. We don't call it WWN in vdsm or engine, so better not use this term.

source/develop/release-management/features/storage/multipath-events.html.md
nirs: If we want to avoid sending tons of messages we will need to keep some multipath device state in vdsm, for example, we can build a mapping based on the last event received for every device.

Lets wait with this until we have real event stream for all flows.

source/develop/release-management/features/storage/multipath-events.html.md
nirs: We can start simple for development, but I don't think this feature will be usable if we don't have a way to aggregate and filter events.

nirs:
@pkliczewski on a system with lot of LUNs (e.g. 300), we may have 1200 events when host loose access to storage.

Typically we have lower number of LUNs, 30 is more reasonable number.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+With these events, the user will be able to detect storage issues faster.
```
nirs: I'm not convinced that logging raw events is useful, and it may also be harmful, hiding other audit logs with 1000's of unhelpful detailed logs about single paths.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+#### Multipath Event
+
+An event with information about the multipath device, path and status.
+
+A mapping with these keys:
+- type: Path Failed/Path Reinstated (DM_ACTION)
+- device (GUID): unique GUID of this device (DM_NAME)
+- valid_paths - number of valid paths (DM_NR_VALID_PATHS)
+- seqnum - event sequence number, may be used to detect missed messages later (DM_SEQNUM)
+
```
nirs: Let add here a section about the design:

### Vdsm multipath event monitoring

Lets separate the part about reporting events. The design includes 2 components:
1. event listener
2. event notifier(s)

We plan now one event notifier for reporting events to engine. In the future we may add other notifiers, for example for updating lvm filte.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Vdsm multipath event monitoring
```
nirs: Lets rename this to "Vdsm multipath event listener"

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+
+Introduce a new Vdsm internal service for handling events:
```
nirs: for listening to multipath events.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+	PATH_REINSTATED - A path has been reinstated.
+
+- The Vdsm will suscribe the monitor to events during startup.
+- The Vdsm will start the Udev event monitor when connecting to the storage pool.
+- The listener will send an event to the engine when its callback is called.
+- The Vdsm will unregister the listener to the Udev event monitor when 
+disconnecting to the storage pool.
+- The user can disable this feature in a specific host in Vdsm configuration file
+
+In the future, the monitor can be used for other code, for example maintaining LVM filter.
```
nirs: - Please change "The vdsm" -> "Vdsm"
- We already described when the monitor is started and stoopped

Lets add a section about reporting events:

### Vdsm multipath event notifier

To report multipath events to engine, we will add a multipath notifier.

- The notifier will subscribe to multipath events during startup
- Received events will be reported to engine using the jsonrpc events infrastructure

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+- Multipath device path down
+
+## Benefit to oVirt
+
+- The user will be able to detect storage issues faster.
+
+
+## User Experience
+
+New audit logs are added with information on the device path status.
```
nirs: Device path status per host? per DC? We need to describe here the user experience in various flows:
- single LUN goes down on all hosts
- switch goes down
- other scenarios?

mykaul:
Another reason why we do not wish to use events (to the engine) but send on
a regular basis is to aggregate (naively) the events:
If a device had 4 paths, then lost one, then another, ...  - we'll send the
latest status, especially if the paths failed in between vdsm<->engine
communication cycles. The user doesn't care that one path went down, then
another, then ... they want the latest status.
This needs to be done in VDSM, I reckon.
So there is 'last event' per device that is stored and sent to engine.

However, then we need to handle flapping links - if within that time a path
went down and up, the event that will be sent is the up (reinstated) while
the 'down' will be lost, which will be a bit confusing (and not really
relevant anymore, except for diagnostics?).

Perhaps a less naive aggregation would be per device to aggregate identical
actions, and different actions do send separately?

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+
+NA
+
+
+## Event Reporting
+
+- 'Multipath device path up'
+- 'Multipath device path down'
+
+
```
nirs: Don't we have an issue increasing the database size a lot by keeping millions of new events?
Do we have policy for dropping old audit logs from the database?

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+[Bug 723931](https://bugzilla.redhat.com/723931)
+
+Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
+
+Uevent [example](https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt)
+
+## Testing
+
+- Connect to a new ISCSI storage server
+  - For each multipath device path a 'Multipath device path up' is triggered
```
nirs: I'm not sure we will get these events, needs testing.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+## Documentation / External references
+
+This feature is required for resolving
+[Bug 723931](https://bugzilla.redhat.com/723931)
+
+Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
+
+Uevent [example](https://www.kernel.org/doc/Documentation/device-mapper/dm-uevent.txt)
```
nirs: This is not an example but "device mapper events kernel documentation"

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,174 @@
...
+- 'Multipath device path up'
+- 'Multipath device path down'
+
+
+## Documentation / External references
+
+This feature is required for resolving
+[Bug 723931](https://bugzilla.redhat.com/723931)
+
+Pyudev - [Monitoring devices](http://pyudev.readthedocs.io/en/latest/guide.html#monitoring-devices)
```
nirs: This is not "Monitoring devices" but "pyudev documention"

nirs:
@rollandf, I think we should change this to use periodic reporting as @mykaul  suggested above.

The event notifier can store the last state for each device, and send one event for all changes collected in the last reporting cycle.

- If all paths went down during the current reporting cycle, send one event instead of event per path.
- If a device went down and then up in the current cycle, send one event for the change, maybe "flip" event?

We can also report the changes using repoStat instead of events, in the same time we report storage domain state. The code reporting repostat can get the current state from the multipath notifier.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Vdsm multipath event listener
+
+Introduce a new Vdsm internal service for listening to multipath events:
+
+- Code interested in receiving multipath events will subscribe a callback to receive the event
+- When an event is received, all subscribers will receive the events (in another thread).
+- If nobody is registered for events, the events are dropped
+- The multipath event monitor will be started when connecting to the pool
+- The multipath event monitor will be stopped when disconnecting from the pool
```
nirs: We should start this when vdsm starts during storage initialization, never stop.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+A Udev event monitor is introduced, listening to device-mapper udev events.
+The relevant events for this feature are PATH_FAILED and PATH_REINSTATED:
```
nirs: We must listen also to ```remove``` event for multipath devices, to remove mapping of fault devices.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+Description:
+Value: Device-mapper specific action that caused the uevent action.
+	PATH_FAILED - A path has failed.
+	PATH_REINSTATED - A path has been reinstated.
+
+- The user can disable this feature in a specific host in Vdsm configuration file
+
+In the future, the monitor can be used for other code, for example maintaining LVM filter.
+
+### Vdsm multipath statistics handler
```
nirs: Vdsm multipath health monitor? Vdsm multipath monitor?

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+To report multipath statistics to engine, we will add a multipath statistics handler.
```
nirs: This is not about statistics but about health.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+- The handler builds a map of all the faulty paths during startup.
+- The handler subscribes to multipath events during startup
```
nirs: We must first register for events - only then we can check the current status.

We can process events only when we finished to fetch the current status, otherwise we may loose events.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+- The handler updates the data set of the faulty paths when its callback is called:
+    - PATH_FAILED - Add the faulty path to the map according to the device GUID
+    - PATH_REINSTATED - Remove the faulty path to the map according to the device GUID
```
nirs: - When multipath device is removed from the system, remove the device from the map.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+- Statistics are provided to engine via 'Host.getStats' verb.
+
+
+### New engine API
+
+New audit logs are added:
+- Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ... 
+- No faulty multipath paths on host "HOST_NAME"
```
nirs: We need 2 type of events for faulty paths:
- devices have some faulty paths
- devices have no valid path

We can have one event reporting both, or 2 events to make it easier to detect the second event which is more critical.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+the full information of the failing paths will be available in the logs.
+The events are not generated on "real-time", but are based on polling of
+the statistics. Only if a change in the data received is detected, a new
+audit log will be created
+
+If a single LUN goes down on all the hosts, the user will get an alarm
+for each host in the DC.
+
+If a single path goes down on one of the hosts, the user will get an alarm
+only for this host.
```
nirs: We should consider aggregating the info on the engine side per DC. In this case we can generate an event about multipath hosts:

- Host A, B, ... have multipath devices with faulty paths
- Multipath device A, B, ... has faulty path on all hosts

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+This feature does not affect engine or Vdsm installation or upgrades
+
+
+## User work-flows
+
+- Once a user connects to a new ISCSI server, no events will be created.
+
+- If there is an issue with the connection to the storage server, and the 
+multipath reports an issue with one or more device path, a "Faulty multipath paths on host" will be created
+
```
nirs: I don't think we will get events in connection flow,  no need to mention connection.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+## CRUD
+
+NA
+
+
+## Event Reporting
+
+- Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ... 
+- No faulty multipath paths on host "HOST_NAME"
```
nirs: Add the extra event for devices without any valid path

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Testing
+
+### Negative flows
+
+- Block network to ISCSI storage server on a specific host
+  - 'Faulty multipath paths on host "HOST_NAME" on devices: "GUID", "GUID" ...' is triggered
+- Unblock access to iscsi server on a specific host
+  - Wait until all paths go up. a 'No faulty multipath paths on host "HOST_NAME"' is triggered
+
```
nirs: We need more detailed description that can help QE person to test this.

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Do you have a link with an example ?

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Vdsm multipath health monitor

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Added

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Let's start with alert per host

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: fixed on earlier version

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: OK, added an additional one

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: I will add more details later

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+- Name: [Fred Rolland](https://github.com/rollandf)
+- Email: <frolland@redhat.com>
+
+
+## Detailed Description
+
+
+### How it works
+
+When a host is connected to a DC (Storage Pool), Vdsm will register
```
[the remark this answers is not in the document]
rollandf: Done

rollandf:
@nirs , can you please review the latest version?

source/develop/release-management/features/storage/multipath-events.html.md
rgolangh: About the push/pull approach - we can do maybe a mixed pull/push approach - vdsm will emit an event if there was an error detected - engine will react to this and will send a message back to get events. When the error rate goes down, or goes down below a threashold, the engine will send a message to stop sending events. Essentially a backpressure. Could be a race here that the engine will ask to don't send events and vdsm will detect an error rate again - just to take this into account

- Requierment wise - do we want users to get a general picture or with actual details, how what is failing specifically, in what rate etc.. ?

nirs: @rgolangh On the engine side the user does not have anything to do with 1000 events when a switch goes down, so I don't see why we should work hard to get the events when there is nothing useful to do with them.

What we need is a better way to aggregate events on the engine side so instead of sending event per host (e.g 100), we will send event per DC (1), like 100 host have problematic multipath device.

In a future version we may add a way to see multipath state on a specific host or less likely state of a multipath device in the cluster (some paths are problematic on some hosts).

Anyway I think starting simple with one event per host when host multipath status changes is good enough.

rgolangh: Maybe I wasn't clear, I want us to get the min event we need to fill the requirement.
But what is the requirement? I see "... to detect storage issues faster" but what should we be presenting on error? how/when does it turn back to green? how reliable should that be to users?

nirs: We have 3 new events per host:

1. Some multipath devices are problematic on a host - the device still function, but no all paths are available - in this case the user want to troubleshoot this issue (e.g. trouble with a switch).
2. Some multipath devices are non-functional (have no valid path). This is the result of ignoring the previous alert, or a bigger issue. The host may move to non-operational due to storage monitoring issues, but it depends
3. All devices are ok on a host - back to normal operation.

The events represent multipath state reported by udev events on the host in the last monitoring cycle.

We don't plan any action on these events at this point, only log the status change. In the future we may modify the representation of the host so user can see the alarm in the host tab or maybe in the dashboard, similar to the way you see host that may need to be upgraded.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,195 @@
...
+A Udev event monitor is introduced, listening to device-mapper udev events.
+The relevant events for this feature are PATH_FAILED and PATH_REINSTATED:
+
+Variable Name: DM_ACTION
+Uevent Action(s): KOBJ_CHANGE
+Type: string
+Description:
+Value: Device-mapper specific action that caused the uevent action.
+	PATH_FAILED - A path has failed.
+	PATH_REINSTATED - A path has been reinstated.
```
nirs: I think we can replace the above paragraph (line 72-81) with:

The service will be implemented using pyudev, listening to these events:

- action=change,DM_ACTION=PATH_FAILED
- action=change,DM_ACTION=PATH_REINSTATED
- action=remove (for multiapth devices)

We don't know yet how to detect multiapth devices events efficiently for detecting removed devices.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,195 @@
...
+
+- The user can disable this feature in a specific host in Vdsm configuration file
```
nirs: When this feature is disabled, engine has to detect that it is disabled, so it does not assume that host does not have any multipath alerts (all is good).

Maybe if this feature is disabled, host will not return the "multiapth" key in repoStat response?

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Owner
+
+- Name: [Fred Rolland](https://github.com/rollandf)
+- Email: <frolland@redhat.com>
+
+
+## Detailed Description
+
+
```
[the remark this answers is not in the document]
nirs: Copying comment from older version:

We need a new section describing how the system behave before, explaning the need for this feature. Maybe start with another paragraph in the overview?

A user/customer looking at this should be able to understand the problem and the benefit in the user level.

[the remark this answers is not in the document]
rollandf: Done

source/develop/release-management/features/storage/multipath-events.html.md
nirs: I guess engine will do nothing in this case.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+because all paths became faulty.
+
+Currently, the user has no indication in oVirt of faulty paths.
+As a result only when all paths are faulty and the access to the storage
+domain is not working, the user will get an indication:
+- If the all the hosts in the Data Center are reporting a problem in the 
+storage domain. the storage domain status that will become 'Inactive'.
+- If only one host reports a problem in the storage domain, the status
+of the host will become 'Non Operational'.
+
```
nirs: Too many blank lines, 2 is enough.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Detailed Description
+
+
+### How it works
+
+Vdsm will register to udev multipath events during storage initialization.
+Vdsm will also build a map of all the faulty paths. 
+For each event, Vdsm will update the list, according to the event.
```
nirs: update the map

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+Engine polls the hosts for the statistics every 15 seconds.
+Engine creates an audit log event according to the information provided
+in the statistics. Only if there are changes in the faulty paths report,
+an event will be created.
+
+
+### New vdsm API
+
+New property in 'HostStats': 'MultipathAlertsMap' which is a mapping of faulty paths
+indexed by the device mapper GUID.
```
nirs: This repeat the next section content. Maybe replace with:

   Host.getStats will report now new "multipath" key. The contents is a map of multipath alerts.

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+If a single path goes down on one of the hosts, the user will get an alarm
+only for this host.
+
+
+## Installation/Upgrade
+
+This feature does not affect engine or Vdsm installation or upgrades
+
+
+## User work-flows
```
nirs: workflow is single word.
http://www.dictionary.com/browse/workflow?s=t

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Once a user connects to a new ISCSI server, no events will be created.
+
+- If there is an issue with the storage server, and the 
+multipath reports an issue with one or more device path, a "Faulty multipath paths on host" will be created
```
nirs: the multipath -> multipath
a "Faulty multipath paths on host" will be created -> a "Faulty multipath paths on host" event will be created

source/develop/release-management/features/storage/multipath-events.html.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Once the issue is fixed, and all paths are up a "No faulty multipath paths on host "HOST_NAME"
+event will be created
+
+- If only part of the paths are fixed, a "Faulty multipath paths on host" will be created with the details of the devices.
```
nirs: "event" missing again.

nirs:
@rollandf lets post a new commit with the fixes, github seems to have trouble handling amending of the same commit.

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: Done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

source/develop/release-management/features/storage/multipath-events.html.md
rollandf: done

PR #1267, opened 2017-10-09
Add feature page: New policy language for MOM

akrejcir:
- Add a feature page describing the design of a new policy language for MOM.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @akrejcir

This pull request needs review by: @MarSik @doron-fediuck @aglitke

### DIFF
--- a/source/develop/release-management/features/sla/mom-declarative-language.html.md
+++ b/source/develop/release-management/features/sla/mom-declarative-language.html.md
@@ -1,0 +1,311 @@
+---
+title: New policy language for MOM
+category: feature
+authors: akrejcir
+feature_name: New policy language for MOM
+feature_modules: mom
+feature_status: design
+---
+
+# New policy language for MOM
+Currently, [MOM](/documentation/sla/host-mom-policy/)
+policies are written in a lisp-like scripting language
+that can be confusing to use.
+We want to change the model and the language,
+to make it easy for users to write their own policies.
+
+## Owner
+* Owner: Andrej Krejcir
+* Email: akrejcir@redhat.com
+
+## Declarative policy representation
+The rule representation uses YAML format.
+The main idea can be explained on a simple example:
+
+```yaml
+scope: VM
+
+conditions:
+  io_or_net_overutilized:
+    any:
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/sla/mom-declarative-language.html.md
```diff
@@ -0,0 +1,276 @@
...
+- `max_relative_change` - Limits the relative change.
+
+```yaml
+rules:
+  balloon_shrink:
+    output: memory.balloon
+    min: policy.balloon_min
+    min_relative_change:
+        value: min_balloon_change_percent
+        interval: 10 sec
```
MarSik: The minimal change does not need the `interval` property. It just won't be saved until the current difference is big enough.

The maximal allowed change needs the interval as it is part of the limit.

source/develop/release-management/features/sla/mom-declarative-language.html.md
```diff
@@ -0,0 +1,276 @@
...
+
+The next section defines rules.
+
+```yaml
+rules:
+    cpu_limit:
+        output: cpu.max_load
+        min: 10
+        target: 0
+        function:
```
MarSik: We should probably mention that in this case the same can be achieved using max_absolute_change: 1 / sec.

The function is supposed to decide on what the next value of the output variable is going to be. And we can even say that the default will be immediate jump to `target` (subject to change rate limits).

source/develop/release-management/features/sla/mom-declarative-language.html.md
akrejcir: Fixed.

source/develop/release-management/features/sla/mom-declarative-language.html.md
akrejcir: Done. I added a section listing possible functions and mentioning that 'constant' function is default.

sandrobonazzola:
@MarSik can you please re-review?

PR #1386, opened 2017-11-28
Update engine-development-environment.html.md

bronhaim:
Updating the requirement to install postgresql-server to install rh-postgresql95-postgresql - the one you get by installing postgresql-server is 9.2 which doesn't work with current engine


bronhaim:
It's not enough, we should also the change rest to work with the scl version ...

PR #1454, opened 2017-12-22
Update Browser Requirements in chap-System_Requirements.html.md

sandrobonazzola:

- Drop reference to user portal / admin portal in browser support now that the user portal has been dropped

### DIFF
--- a/source/documentation/install-guide/chap-System_Requirements.html.md
+++ b/source/documentation/install-guide/chap-System_Requirements.html.md
@@ -62,13 +62,13 @@ Browser support is divided into tiers:
 
 **Browser Requirements**
 
-| Support Tier | Operating System Family | Browser | Portal Access |
+| Support Tier | Operating System Family | Browser |
 |-
-| Tier 1 | Enterprise Linux | Mozilla Firefox Extended Support Release (ESR) version | Administration Portal and User Portal |
-| Tier 2 | Windows | Internet Explorer 10 or later | Administration Portal and User Portal |
-|        | Any | Most recent version of Google Chrome or Mozilla Firefox | Administration Portal and User Portal |
-| Tier 3 | Any | Earlier versions of Google Chrome or Mozilla Firefox | Administration Portal and User Portal |
-|        | Any | Other browsers | Administration Portal and User Portal |
+| Tier 1 | Enterprise Linux | Mozilla Firefox Extended Support Release (ESR) version |
+| Tier 2 | Windows | Internet Explorer 10 or later |
+|        | Any | Most recent version of Google Chrome or Mozilla Firefox |
+| Tier 3 | Any | Earlier versions of Google Chrome or Mozilla Firefox |
+|        | Any | Other browsers |
 
 
 ### Client Requirements


sandrobonazzola:
@gregsheremeta can you please review?

gregsheremeta:
+1 on dropping the user portal references.

I'm not actually sure we follow those tiers anymore, though. I don't know how much oVirt is tested on FF ESR. We don't use the terminology "tiers" internally.

And this conflicts with what I put in the browsers section at https://www.ovirt.org/download/ where I basically said "use evergreen browsers."

So, in summary, browser matrix is up to @oourfali or @jelkosz -- in my view we don't currently have one.

jelkosz:
+1 for dropping portal references too.

About tiers: we still do have them and ovirt still should be tested accordingly. So I would not hesitate to communicate it accordingly here.

About  https://www.ovirt.org/download/ - even though in theory I agree with that statement, I really doubt it is being systematically tested on Safari as much as on FF.

source/documentation/install-guide/chap-System_Requirements.html.md
```diff
@@ -62,13 +62,13 @@ Browser support is divided into tiers:
...
-| Support Tier | Operating System Family | Browser | Portal Access |
+| Support Tier | Operating System Family | Browser |
 |-
-| Tier 1 | Enterprise Linux | Mozilla Firefox Extended Support Release (ESR) version | Administration Portal and User Portal |
-| Tier 2 | Windows | Internet Explorer 10 or later | Administration Portal and User Portal |
-|        | Any | Most recent version of Google Chrome or Mozilla Firefox | Administration Portal and User Portal |
-| Tier 3 | Any | Earlier versions of Google Chrome or Mozilla Firefox | Administration Portal and User Portal |
-|        | Any | Other browsers | Administration Portal and User Portal |
+| Tier 1 | Enterprise Linux | Mozilla Firefox Extended Support Release (ESR) version |
+| Tier 2 | Windows | Internet Explorer 10 or later |
```
mykaul: I highly doubt anything below IE Edge is working. IE 11 and lower are slow and we know of issues with them.

source/documentation/install-guide/chap-System_Requirements.html.md
```diff
@@ -62,13 +62,13 @@ Browser support is divided into tiers:
 
 **Browser Requirements**
 
-| Support Tier | Operating System Family | Browser | Portal Access |
+| Support Tier | Operating System Family | Browser |
```
mykaul: I'm not sure what 'support tier' mean in oVirt.

jelkosz: Something like "how much do we recommend to use it" :) But I agree it does not make sense in oVirt to call it like this. When I think of it a bit more I tend to agree with @gregsheremeta to have a statement similar to https://www.ovirt.org/download/  - even though Im not really sure the "We validate against and fully support " statement is correct. Maybe something like:
"We recommend to use one of the recent versions of the following "evergreen" browsers:..."?

sandrobonazzola:
:+1: for changes included up to now. @mykaul we should probably raise the IE Edge / IE 11 issue to Derek for updating RHV docs too. @jelkosz so, should we close this PR and replace the whole support matrix with the general statement about latest browser? can you push that PR?

jelkosz:
@sandrobonazzola done: https://github.com/oVirt/ovirt-site/pull/1465

sandrobonazzola:
Ok, abandoning this in favor of #1465

source/documentation/install-guide/chap-System_Requirements.html.md
gregsheremeta: > Im not really sure the "We validate against and fully support " statement is correct

Yeah, fair enough :) I'd love to get there with the automated testing I'm working on, though :D

PR #1530, opened 2018-02-20
Add HA VMs howto

mz-pdm:
This howto should clarify basic facts about HA VMs, with a special focus on I/O errors.

This pull request needs review by:  @michalskrivanek


mkumku:
Milan, if I understand correctly, you recommend using KILL with leases. However the screenshot at the very end shows NO Leases + KILL. I would use a different screenshot just to keep things visually clear and consistent.

mkumku:
Another suggestion - maybe we want to clearly state at the VM Leases section that if you choose to use LEAVE_PAUSED, you should not use leases.
And lastly - should we explain here that once we kill the VM, the lease is released as well? (or I am wrong in this assumption? and I should read the leases document?)

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+- A user asks the VM to start paused.
+- A VM can be paused for a short while in the final stages of migrations or when making a snapshot with memory.  This typically doesn't harm VM availability.
+- When a thin provisioned disk image is getting full and must be enlarged.  This is again only temporary situation.
+- An I/O error due to inaccessible or broken storage occurs.  This is a real problem.
+
+A paused VM can be later switched to another state (typically Running or Down) either automatically (e.g. on migrations) or manually by the user.
+
+## Highly available VMs and I/O errors
+
+When a VM attempts to access an unreachable or otherwise broken storage and I/O error occurs, the VM gets paused.  Then it can't get running again until the storage problem disappears.  The paused VMs can then be handled in different ways by oVirt.  Once storage is working again, Vdsm tries to resume all the VMs, highly available or not, on the given host that have been paused due to storage errors (it doesn't try to resume any other paused VMs).  Whether any such VM is actually resumed automatically by Vdsm or handled differently depends on VM *resume behavior*.
```
jelkosz: You first write that "Once storage is working again, Vdsm tries to resume all the VMs, highly available or not" and than "Whether any such VM is actually resumed automatically by Vdsm or handled differently depends on VM *resume behavior*". I would propose to write it slightly differently, something like:
"Once storage is working again, VDSM goes over all paused VMs and handles them according to the configured *resume behavior*"

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+## VM leases
+
+VM lease is a feature that guarantees a VM can't be run in multiple instances at the same time.  That guarantee makes restarting highly available VMs safer and thus easier.  Split brain is not possible when a VM has been started with a lease and the lease is kept enabled for the VM.
+
+VM leases can be enabled only for highly available VMs, in High Availability section of the VM editing dialog again.  Just select *Target Storage Domain for VM Lease* in the dialog.
+
+VM leases are thoroughly described in a separate [document about leases](/develop/release-management/features/storage/vm-leases/).
+
+To prevent split brain it's best to use VM leases on highly available VMs and set `KILL` resume behavior for them.  The same also helps restarting highly available VMs on other hosts in case of prolonged storage problems on some of the hosts.  The primary purpose of `KILL` resume behavior is to destroy a highly available VM before Engine moves the VM to Unknown status and attempts to start it elsewhere.  Of course, `KILL` behavior means killing the VM under the given conditions, so think about your setup before using it.
```
jelkosz: "it's best to use VM leases on highly available VMs and set `KILL` resume behavior for them." -> you don't have an option here. If the VM is a HA VM with lease, it has to have the KILL behavior.

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Resume behaviors
+
+VM resume behavior defines how a VM previously paused due to an I/O error is handled once the previously broken storage starts working again.  Resume behavior can be selected in High Availability section of VM editing dialog, together with other VM high availability settings.
+
+The following resume behaviors are defined:
+
+- `AUTO_RESUME`: The VM is resumed automatically, as soon as the storage gets working again.  This is the simplest scenario – the VM simply pauses for the time of storage problems and then continues running again, without user intervention.  It is supposed to be used for non-highly available VMs without special requirements and for highly available VMs when using `KILL` behavior is not desirable for some reason.
```
jelkosz: I would mention somewhere (maybe here) that the auto_resume might get in your way quite badly if you are using HA on application level. In that case, if one of your nodes disappear, your application level HA handles it somehow and than, if the same node re-appears later with old state, it might be confusing. So in this case it is better to use one of the leave_paused or kill options.

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+VMs manually shut down or powered off from oVirt or from the guest operating system are not restarted automatically.  Switching VMs off manually is a user action and oVirt honors it.  The user is responsible for starting such a VM again manually.
+
+There may be other reasons preventing start of a highly available VM, such as impossibility to run the VM due to resource constraints at a given moment.
```
jelkosz: might be a good place to add a link to the recent devconf presentation from Arik here about this topic: https://www.youtube.com/watch?v=Gh2eB06IE2A

mz-pdm:
@jelkosz, thank you for you comments, I fixed everything in the updated version.

mz-pdm:
@mkumku, thank you for your suggestions. I updated the screenshot. As Tomáš pointed out, only KILL resume behavior can be selected for VMs with leases, I added the information to the document. You are right with your assumption about VM release, I mentioned that in the updated document.

jelkosz:
@mz-pdm thank you, LGTM

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+The very first condition is that oVirt Engine must be fully certain that the VM is not running anywhere at the moment.  Failing to satisfy this condition could lead to multiple instances of the VM running concurrently on different hosts, so called *split brain* situation.  oVirt uses safety mechanisms such as sanlock data, VM leases and resume behaviors to prevent split brain.  But it's still important to be aware that if oVirt Engine doesn't know (under certain setups) for sure that a VM is not already running, it can't start it automatically elsewhere.  For instance, if a host running a VM is unreachable it doesn't mean that the VM is not happily running there.  If corresponding protection mechanisms are not enabled then oVirt Engine may not be able to start the VM elsewhere, despite it is highly available.
+
+VMs manually shut down or powered off from oVirt or from the guest operating system are not restarted automatically.  Switching VMs off manually is a user action and oVirt honors it.  The user is responsible for starting such a VM again manually.
```
michalskrivanek: I would also mention the relatively recent RFE 1401054 where on admin-issued host shutdown VMs are shut down too and it is considered a "user shut down from within the guest" case - i.e. without HA restart

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+- A user asks the VM to start paused.
+- A VM can be paused for a short while in the final stages of migrations or when making a snapshot with memory.  This typically doesn't harm VM availability.
+- When a thin provisioned disk image is getting full and must be enlarged.  This is again only temporary situation.
+- An I/O error due to inaccessible or broken storage occurs.  This is a real problem.
+
+A paused VM can be later switched to another state (typically Running or Down) either automatically (e.g. on migrations) or manually by the user.
+
+## Highly available VMs and I/O errors
+
+When a VM attempts to access an unreachable or otherwise broken storage and I/O error occurs, the VM gets paused.  Then it can't get running again until the storage problem disappears.  The paused VMs can then be handled in different ways by oVirt.  Once storage is working again, Vdsm goes over all the VMs, highly available or not, on the given host that have been paused due to storage errors (it doesn't try to resume any other paused VMs).  What happens with each of the paused VMs depends on VM *resume behavior*.
```
michalskrivanek: "the VM gets paused" - the VM may get paused. I would mention it qemu can also hang in host i/o access
"Once storage is working again" - once Storage Domain is working again?

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,86 @@
...
+- `LEAVE_PAUSED`: The VM is never resumed automatically.  A user must handle the VM (run it or power it off) manually.  This behavior is suitable for VMs that require additional manual interventions after being paused, e.g. some remedy in their guest OS or software running there.
+- `KILL`: This is similar to `AUTO_RESUME` unless the VM is paused for too long.  If the storage problem is remedied within a given time interval (80 seconds by default), the VM is automatically resumed.  But if the storage problem persists for longer time, the VM is going to be killed sooner or later.  This behavior is suitable for highly available VMs, in order to permit them to start on another host, which doesn't suffer from the storage problem.
+
+Resume behavior can be selected on any VM, not just highly available VMs.  But there are special considerations regarding highly available VMs:
+
+- As long as a VM is paused, it may be prevented from starting on another, working, host.
```
michalskrivanek: "it may prevent engine to start"?

michalskrivanek:
few minor comments, looks good

mz-pdm:
@michalskrivanek , thank you for review, I made the changes you suggested.

@mykaul , the document is now reviewed by different persons and completed. Can it be merged?

mz-pdm:
@mykaul , could this be merged, please?

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+Basically, a VM with a lease can start on a given host only after it acquires its lease there.  Once the VM stops, gets paused, is killed or its QEMU process crashes, it releases the lease, making it available for starting the VM on another host.  VM leases are thoroughly described in a separate [document about leases](/develop/release-management/features/storage/vm-leases/).
+
+To prevent split brain it's best to use VM leases on highly available VMs.  The same also helps restarting highly available VMs on other hosts in case of prolonged storage problems on some of the hosts.  Since paused VMs release their leases, the VMs become available for running on other hosts, but can cause trouble when attempting to resume them later on the original host.  For that reason `KILL` is the only possible resume behavior for VMs with leases, which helps destroy the paused VM instance when Engine moves the VM to Unknown status and attempts to start it elsewhere.  Of course, `KILL` behavior means killing the VM under the given conditions, so think about your setup before using it.
```
ahadas: 1. I think this part is incorrect, specifically the "For that reason `KILL` is the only possible resume behavior for VMs with leases". theoretically, VMs should attempt to reacquire their lease when they are resumed from a paused state so it should be safe to use AUTO_RESUME.
2. I think we should emphasize that the `KILL` option may not fit some stateful applications. For instance, let's say that my application holds a state and it takes the application 30 minutes to create this state. In this case, I may prefer to do my best to resume a paused VM that runs this application rather than killing it and starting it elsewhere. There is a trade-off between maximizing the time that application is running (e.g., I may prefer to restart the VM as soon as I find that it can run properly on another host), and reusing its existing state (e.g., I may prefer to wait for a couple of minutes rather than waiting for the VM to boot and 30min for the application to create the state that makes it functional again). A possible solution for this would be not to kill the VM but rather migrate the paused VM to another host where it can be functional with its current state.

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Additional notes
+
+VM resume behaviors are available since oVirt 4.2.  Older oVirt releases applied `AUTO_RESUME` behavior on all VMs.
+
+Up to oVirt 4.2.2 VMs with `KILL` behavior used to be killed only once the storage became available, i.e. Vdsm tried to resume the paused VMs.  Starting from oVirt 4.2.3 VMs are killed shortly after the timeout expires even when the storage problem persists.
```
ahadas: I think this should be optional setting, and should be a decision made by the engine - the host is not aware of the cluster's state that may allow us to do smarter (yes, and more complicated but not that much) things

michalskrivanek: this article describes the current state. Such enhancements are to be documented/designed in RFEs

source/documentation/how-to/ha-vms.html.md
```diff
@@ -0,0 +1,84 @@
...
+authors: mzamazal
+category: howto
+layout: toc
+---
+
+# Highly Available VMs
+
+## What is a highly available VM?
+
+Highly available VMs (*HA* VMs) are VMs restarted by oVirt automatically whenever they disappear accidentally, such as due to failure of the host running the VM, storage problems or QEMU crash.
```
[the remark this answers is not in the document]
michalskrivanek: @ahadas can you please add a section with more details about how the HA VMs are actually restarted by engine? (# of retries, other stuff?)

source/documentation/how-to/ha-vms.html.md
michalskrivanek: 1. - again, this describes current state, not what it could potentially do. We do not do reacquire because of the auto restart of unknown VMs
2. makes sense to mention the timeouts tradeoff. Though we do not really have a very good estimate of how long it's going to take in all cases

ahadas: 1. but I don't think reacquiring the lease is something we should do - libvirt is supposed to do that, do we know that they don't?
2. it depends on the target audience actually - if we want this page to be a technical report of how we consider high availability in oVirt and the mechanism we implemented for that than maybe it is not needed, but if it is targeted for users, that most probably less interested in how we implemented this, but more interested in how they can leverage HA VMs in oVirt in order to make **their application(s)** highly available, then it worth mentioning this trade-off even without concrete numbers IMHO.

source/documentation/how-to/ha-vms.html.md
ahadas: I see, so 4.2.3 should be dropped then

mz-pdm: The change has already been merged to 4.2 so I updated the mentioned versions (and it indeed documents the current state).

source/documentation/how-to/ha-vms.html.md
mz-pdm: 1. I don't know.

2. Good remark, I added note about that.

michalskrivanek: > but I don't think reacquiring the lease is something we should do - libvirt is supposed to do that, do we know that they don't?

it is reacquired by qemu on resume. But it breaks with our feature to start Unknown VMs blindly (from this host POV). But unless we have more logic on engine side to know that the VM was paused there and we didn't really restart it elsewhere in the meantime we cannot use it.
So again as a reflection of current state this is accurate I believe

michalskrivanek:
thanks. @ahadas, you can now amend the final page with more details about engine side HA restart

PR #1544, opened 2018-03-05
Fix a blog post header problem

rgolangh:


rgolangh:
@mykaul @sandrobonazzola

rgolangh:
Locally with ./run-server I see it - and did see it even prior to this fix, so I'm not sure what's going on.

image

PR #1546, opened 2018-03-05
Cleanup CLI code blocks and indentation

duckbrain:
It was difficult to read the auto-completion outputs because they were not in a code block. There were many inconsistencies with indentation. Some parameters were not in code blocks that needed it to prevent being turned into an emdash.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @duckbrain

This pull request needs review by:

### DIFF
--- a/source/develop/release-management/features/infra/cli.html.md
+++ b/source/develop/release-management/features/infra/cli.html.md
@@ -37,32 +37,40 @@ __TOC__
 
 #### available commands
 
-`   [oVirt shell (connected)]# `<TAB><TAB>
-         EOF         clear       console     remove      echo        help        ping        
-         show        update    action        connect     add          disconnect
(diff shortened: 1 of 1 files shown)


mykaul:
@duckbrain - thanks, but the most important part is missing - ovirt-shell is deprecated, as it uses the v3 REST API. It lacks many features (essentially all 4.x features) and should be replaced with Ansible for automation purposes.

Can you add a comment in that spirit?

duck-rh:
I wondered why sometimes it felt difficult to think straight. I should not let my brain roam on the Internet ;-P.

duckbrain:
@mykaul Good to know. I guess I'll have to look into Ansible myself.

PR #1624, opened 2018-05-21
remove old stats

duck-rh:
The old Bitergia system was taken down.

It was replaced by a redirect, see OVIRT-2038


duck-rh:
@sandrobonazzola could you have a look please?

PR #1626, opened 2018-05-21
SSL for install instructions

quaid:
SSL everywhere is a good idea overall, but is a MUST condition for installation instructions to use sudo to install packages.

While we're at it, might as well update all the links to good ol' resources.ovirt.org .

I recommend updating the other links for the mirrors, but did not find a viable HTTPS endpoint (I didn't look further than to attempt to load the page as https://). The mirrors that did have a valid HTTPS endpoint were updated in this PR. Perhaps you can request your mirror providers to use an SSL certificate now that e.g. Let's Encrypt is ubiquitous, useful, automatable, etc.

- Modify project URLs to use HTTPS

- Modify mirror URLs to use HTTPS where available

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @quaid

### DIFF
--- a/source/download/index.html.haml
+++ b/source/download/index.html.haml
@@ -88,7 +88,7 @@ authors: bproffitt, dneary, knesenko, mburns, sandrobonazzola, theron
   you have the correct repositories enabled by running the following commands before upgrading with the usual procedure.
   <br/>
   ```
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
Can you please rebase?

gregsheremeta:
I took over, see #1689

PR #1704, opened 2018-08-01
Added 4k blocks support feature page.

akashihi:

- Added VDSM 4k block support feature page

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): (please @akashihi)

This pull request needs review by: (@tnisan @fromanirh @nirs @alhino @aglitke)

### DIFF
--- a/source/develop/release-management/features/storage/4k-blocks.md
+++ b/source/develop/release-management/features/storage/4k-blocks.md
@@ -1,0 +1,165 @@
+---
+title: 4k drive support
+category: feature
+authors: dchaplyg
+feature_name: 4k drive support
+feature_modules: vdsm
+feature_status: Planning
+---
+
+# 4k drive support
+
+# Summary
+
+VDSM assumes, that 512 bytes is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, cause while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM. 
+
+Unfortunately, some 4k block storage devices either refuse to read/write smaller blocks atomically (some SAN appliances) or have a performance impact while emulating 512 bytes block device (like VDO). 
+
+Therefore without 4k block device support oVirt either can't use those devices as a storage domain or will have performance degradation on those domains.
+
+## Owner
+
+*   Name: [Denis Chaplygin](https://github.com/akashihi)
+*   Email: <dchaplyg@redhat.com>
+
+## Current status
+
+*   Target Release: 4.3
+*   Status: Prototyping
+*   Last updated: August 22, 2018
+
+# Implementation details
+
+VDSM heavily relies on 512 block support and it is hardcoded to a lots of places.
+
+The general approach is to refactor all block size dependent code to use block size from the storage domain, while keeping it hardcoded to 512 bytes in it and in a second step, implement block size detection at the SD level and replace hardcoded value with actual value.
+
+Subsystem specific notes are below:
+
+## Sanlock
+
+Sanlock is used by oVirt for locking and fencing shared resources. Latest released sanlock have native support for 4k devices, but it comes with price of great disk space overhead. 
+
+Sanlock is planned to be extended with new on-disk formats, incluging:
+
+ - ALIGN1M | SECTOR512: max_hosts 2000
+ - ALIGN1M | SECTOR4K:  max_hosts 250
+ - ALIGN2M | SECTOR4K:  max_hosts 500
+ - ALIGN4M | SECTOR4K:  max_hosts 1000
+ - ALIGN8M | SECTOR4K:  max_hosts 2000
+
+Thus VDSM may use ALIGN1M | SECTOR512 format for 512 bytes devices, consuming 1MB for each sanlock file. On the 4k block devices VDSM could use ALIGN1M | SECTOR4K or ALIGN2M | SECTOR4K, thus consuming either 1MB or 2MB for each sanlock file. In case of ALIGN2M | SECTOR4K usage, all lockspace files will be bigger, comparing to 512 bytes format. 
+
+With patch [1] it will be possible, to specify desired alignment, but it is not yet available in the python bindings (needs to be implemented).
+
+## Blocksize detection
+
+There is no reliable way to detect size of the logical block of the underlying device in case of a network filesystem. Both NFS (v3, v4, v4.1, v4.2) and Ceph have no aligment restrictions enforced, therefore they can't be probed for a mimally allowed block IO size. 
+
+Therefore desired block size for a new storage domain must be specified explicitly, via oVirt UI or API during storage domain creation. StorageDomain.create verb needs to be extended with that information (actually a StorageDomainCreateArguments type needs to be extended)
+
+## Metadata
+
+VDSM needs to know size of block of the Storage domain and the best way to store it is metadata. Metadata for block device backed domains already includes corresponding fields:
+
+* DMDK_LOGBLKSIZE
+* DMDK_PHYBLKSIZE
+
+But metadata for the filesystem backed storage domains do not have those fields.  Block/FS domains also differ in two other fields:
+
+* DMDK_VGUUID - at block backed SD
+* REMOTE_PATH - at filesystem backed SD
+
+As those fields are unused, new version of storage domains needs to implemented, removing unused fields from the metadata and adding block size fields to the filesystems based SD, thus unifying metadata.
+
+For readability reasons DMDK_LOGBLKSIZE can be renamed to just BLOCK_SIZE.
+
+This also means, that manifests for both types of domains should include block size field. Block based domains already do that, but filesystem domains do not
+
+## Locking
+
+VDSM uses sanlock for all types of locking. At the moment domain metadata consists of following lock volumes:
+
+* ids - sanlock delta lease
+* leases - volume leases
+* xleases - external leases
+
+Those lease volumes use ~3.5GB for 512 block devices and will use ~25GB for 4096 block devices (worst case). Because of that huge overhead, a change to the locking scheme is proposed.
+
+### xleases
+
+Sanlock already supports add/remove/locate leases, based on vdsm index code, thus making VDSM implementation unneeded. Those new APIs needs to be implemented in Python bindings and VDSM should be provided with new APIs, that will be using those bindings. After that new storage domains could be switched to that new external leases
+
+### Dynamic locks
+
+With new external leases and new sanlock on-disk formats a new locking scheme, using dynamic lockspace allocations is proposed:
+
+* VM leases will be covered by new external leases (create exlease per vm lease)
+* Job leases will be covered by new external leases (Create xlease per job/volume when starting a job)
+* SPM lease needs to be moved to new external lease
+* HostedEngine volume lease should be switched to new external lease
+* Leases volume and old xleases volume will be dropped
+
+With that approach we may approximately need:
+
+* 1500 leases (1 lease per vm)
+* 50 concurrent jobs with leases
+
+This will be using about ~1.55Gb for 512 bytes block devices or 4096 bytes block devices with 250 hosts per cluster or ~3.1GB for 4096 block devices with 500 hosts per cluster.
+
+## Volumes
+
+BlockVolume and FileVolume implementations must use block size from the StorageDomain. That should be done in two steps - first move implementations to use hardcoded block size at StorageDomain and after that make block size adjustable at the StorageDomain side. To achieve that, Volume create must be moved to the StorageDomain, where it belongs.
+
+Image class also heavily relies on hardcoded block size. We can solve that by eliminating Image class at all and implementing it's functionality as a set of jobs under sdm/api
+
+## Mailboxes
+
+Desipte mailboxes code uses BLOCK_SIZE constant, it seems to be agnostic to actual BLOCK_SIZE and doesn't requires any modifications.
+
+We may need to replace 512 byte BLOCK_SIZE here with 4096 byte IO to be sure, that we will never issue an unaligned IO.
+
+# Implementation plan
+
+## Sanlock
+
+* Implement new on-disk formats
+* Add format/alignment selectors to the sanlock python bindings
+* Add new indexing APIs to the sanlock python bindings
+
+## VDSM 
+
+* Get rid of BLOCK_SIZE constant by moving it to the SD base class and making it a (read-only) property of that class
+* Make a quick PoC, proving, that 4k domain can be implemented
+* Reimplement Image class as a set of jobs under sdm/api
+* Move volume creation code to the StorageDomain 
+* Implement new external lease APIs
+* Make mailboxes 4k friendly
+--- at that stage we switch to the new storage domain format ---
+* Create new storage domain metadata format (see above) for V5 domains
+* Migrate to the new locking scheme (do we need a locking API, that will hide differences accessing old/new domains)
+* Add block size to the StorageDomain.create verb and make StorageDomain block_size property writable
+* Check, that 4k block support is working.
+
+## HostedEngine
+
+TBD
+
+## Engine
+
+Storage domain creation dialog should be extended with 512/4096 block size selector. In case of well known 4096 only deployments (hyperconverged on top of VDO for example), 4096 should be preselected and selector must be disabled.
+
+oVirt API should accept desired block size in a StorageDomain creation endpoint.
+
+# Benefit to oVirt
+
+* Support for 4k only SAN
+* Support for VDO native block size
+* Support 4k based file based storage
+
+# Open issues
+
+* [BUG-1592916: (https://bugzilla.redhat.com/1592916) -  Support device block size of 4096 bytes for file based storage domains]
+
+# External links
+[1] https://pagure.io/sanlock/c/d5e4def0d087dda942e7ac4330b288715603cc61?branch=master
--- a/data/authors.yml
+++ b/data/authors.yml
@@ -2713,3 +2713,7 @@ crutonjohn:
   email: cujo0224@gmail.com
   gravatar: fb9d43da7d8c95ccdcd36b82eb2d81ae
   
+
+dchaplyg:
+  name: Denis Chaplygin
+  email: dchaplyg@redhat.com


source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+VDSM heavily relies on 512 block support and it is hardcoded to a lots of places.
+
+The general approach is to refactor all block size dependent code to use block size from the storage domain, while keeping it hardcoded to 512b in it and in a second step, implement block size detection at the SD level and replace hardcoded value with actual value.
+
+Subsystem specific notes are below:
+
+## Sanlock
+
+Sanlock is used by oVirt for locking and fencing shared resources. Latest released sanlock have native support for 4k devices, but, unfortunately, at the moment it work only with block devices, not with filesystems. Speaking precisely, sanlock on filesystems always assumes, that block size of the device is 512b. With patch [1] it will be possible, to specify desired block size, but it is still not able to detect block size of the filesystem backing device.
+
```
sabose: Is the patch merged? What's the impact of not being able to auto-detect?

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+## Blocksize detection
+
+VDSM have to detect block size during storage domain creation. For block device backed storage domains sanlock or blkid could be used. For filesystem backed storage domains VDSM can actually try to make 512b atomic operation and switch to 4k only if it fails. Something like dd if=/fs/under/test/file of=/dev/null bs=512 count=1 iflag=direct
+
```
sabose: Isn't it better to default to 4K - since most storage would support 4K writes? Is there a benefit in defaulting to 512?

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
+---
+title: 4k block device support for the storage domains
+category: feature
+authors: dchaplyg
+feature_name: 4k block device support for the storage domains
+feature_modules: vdsm
+feature_status: Planning
```
sandrobonazzola: @duck-rh is above header ok? or is ti still something inherited from wiki conversion we should drop?

gregsheremeta: > should drop?

if we don't drop it now, we will be soon :)

source/develop/release-management/features/storage/4k-blocks.md
akashihi: It is merged, but not yet released.

source/develop/release-management/features/storage/4k-blocks.md
gregsheremeta: The header. The things between the ---

gregsheremeta: @akashihi i looked into this. Indeed you can delete these 3 lines

feature_name: 4k block device support for the storage domains
feature_modules: vdsm
feature_status: Planning

gregsheremeta:
not sure why travis didn't run. It should run on next commit

gregsheremeta:
> As there are conflicts to solve anyway, let's see how it goes when these fixes are pushed

several PRs have run successfully since this one. It was some kind of temporary blip :)

source/develop/release-management/features/storage/4k-blocks.md
gregsheremeta: I'm pretty sure we'll remove that features listing thing. Maybe even move 'features' (design documents) into the new ovirt-design repo. Not 100% sure yet.

so @akashihi you can leave the header.

gregsheremeta:
cool, travis worked this time

source/develop/release-management/features/storage/4k-blocks.md
duck-rh: I was not aware of this new repo. It's your decision to make but I would suggest using the domain name instead of the GH domain. I think it is now possible to using GH pages with HTTPS and custom domains. We can also build/publish on OSAS servers the same way the website is done (with Jekyll instead of Middleman).

In this case I guess these headers can be removed but doing this while migrating the page instead of breaking the feature listing seems a better option.

gregsheremeta: > breaking the feature listing

I wouldn't agree with the word "break." There are tons of problems with feature pages that I'm not going to go into here. I'll soon send a comprehensive email to devel about my plans to improve the site's information architecture.

duck-rh: I'm not a developer, so not on this list, so could you Cc/forward this to me please?

duck-rh: maybe even the infra list could be interested

gregsheremeta: Yes, when I get around to sending it, I will include you.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+# 4k block device support for the storage domains
+
+# Summary
+
+VDSM assumes, that 512b is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, case while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM.
```
nirs: 512b -> 512 bytes

 You need to backup this claim some reference:
    
    During virt-v2v, we can use qemu-img measure on the overlay. It gives a conservative
    estimate that will never fail, based on the number of allocated blocks and additional
    metadata required for the destination file format.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+feature_status: Planning
+---
+
+# 4k block device support for the storage domains
+
+# Summary
+
+VDSM assumes, that 512b is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, case while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM. 
+
+Unfortunately, some 4k block storage devices either refuse to read/write smaller blocks atomically (some SAN appliances) or have a performance impact while emulating 512b block device (like VDO).
```
nirs: We need more info on this issue. How widespread are drives or storage systems that do not support 512 bytes emulation?

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+# 4k block device support for the storage domains
+
+# Summary
+
+VDSM assumes, that 512b is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, case while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM. 
+
+Unfortunately, some 4k block storage devices either refuse to read/write smaller blocks atomically (some SAN appliances) or have a performance impact while emulating 512b block device (like VDO). 
+
+Therefore without 4k block device support oVirt either can't use those devices as a storage domain or will have performance degradation on those domains.
```
nirs: You need to backup this with performance results on real world scenarios.

Interesting tests would be:
- some "normal" workload in a vm, once on a 4k drive, and once on 4k drive in 512k emulation.
- copying disks using qemu-img

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Current status
+
+*   Target Release: 4.3
+*   Status: Planning
+*   Last updated: July 31, 2018
+
+# Implementation details
+
+VDSM heavily relies on 512 block support and it is hardcoded to a lots of places.
```
nirs: Vdsm relies heavily on hard-coding the value of block size :-)

source/develop/release-management/features/storage/4k-blocks.md
nirs: It does not expose the capability via the python binding yet.

source/develop/release-management/features/storage/4k-blocks.md
nirs: Using 4096 may loose atomic writes if the actual sector size is 512. We don't want to go there.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+## Blocksize detection
+
+VDSM have to detect block size during storage domain creation. For block device backed storage domains sanlock or blkid could be used. For filesystem backed storage domains VDSM can actually try to make 512b atomic operation and switch to 4k only if it fails. Something like dd if=/fs/under/test/file of=/dev/null bs=512 count=1 iflag=direct
+
+## Metadata
+
+VDSM needs to know size of block of the Storage domain and the best way to store it is metadata. Metadata for block device backed domains already includes corresponding fields:
+
+* DMDK_LOGBLKSIZE
+* DMDK_PHYBLKSIZE
```
nirs: I don't think we need the physical block size. We need only the logical block size used for alignment when using direct I/O.

Maybe the physical block size is used in lvm, we reject adding a PV with different block size, not sure if we check the logical or physical block size in this case.

For file storage we probably use only DMKD_LOGBLKSIZE, but I prefer to use something more readable like BLOCK_SIZE. Anyway we can handle the details later, lets not mention it here.

The only think that matters at this point is:
- Detect underlying block size
- Keep block size in storage domain metadata

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+But metadata for the filesystem backed storage domains do not have those fields.  Block/FS domains also differ in two other fields:
+
+* DMDK_VGUUID - at block backed SD
+* REMOTE_PATH - at filesystem backed SD
```
nirs: This is not related to the feature, lets remove this.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+As those fields are unused, new version of storage domains needs to implemented, removing unused fields from the metadata and adding block size fields to the filesystems based SD, thus unifying metadata.
+
+This also means, that manifests for both types of domains should include block size field. Block based domains already do that, but filesystem domains do not
+
+## Locking
+
+xlease modules uses BLOCK_SIZE constant and replacing it with externally supplied block size should make it 4k friendly.
```
nirs: This should work, but we want to use the new sanlock rindex apis to create leases instead of this code.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+SDM and domain locking use hardcoded block size and it should be fixed.
+
+## Volumes
+
+VolumeManifest have support for variable block sizes, but implementations hard code it to 512b blocks. SD volume creation code should set block size at the Volume.
```
nirs: It should use the domain block size.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+Volume creation flow is also handled by sdm.VolumeArtifacts, that uses BLOCK_SIZE constants, which should be replaced with block size from the volume manifest.
```
nirs: This class is not used, this is dead code left from the work on removing the SPM. You can ignore this code and remove it from the page.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+SDM and domain locking use hardcoded block size and it should be fixed.
+
+## Volumes
+
+VolumeManifest have support for variable block sizes, but implementations hard code it to 512b blocks. SD volume creation code should set block size at the Volume.
+
+Volume creation flow is also handled by sdm.VolumeArtifacts, that uses BLOCK_SIZE constants, which should be replaced with block size from the volume manifest. 
+
+BlockVolume and FileVolume implementations must use block size from the VolumeManifest. That should be done in two steps - first move implementations to use hardcoded block size at VolumeManifest and after that make block size adjustable at the VolumeManifest side.
```
nirs: The other way around, volume code should use block size from the domain. We should probably move volume creation code to the storage domain, where it belongs.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+Image class also heavily relies on hardcoded block size, should take it from the SD manifest.
```
nirs: Image class should be eliminated. All the code there should become storage jobs under sdm/api/.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+Volume creation flow is also handled by sdm.VolumeArtifacts, that uses BLOCK_SIZE constants, which should be replaced with block size from the volume manifest. 
+
+BlockVolume and FileVolume implementations must use block size from the VolumeManifest. That should be done in two steps - first move implementations to use hardcoded block size at VolumeManifest and after that make block size adjustable at the VolumeManifest side.
+
+Image class also heavily relies on hardcoded block size, should take it from the SD manifest. 
+
+## Mailboxes
+
+Desipte mailboxes code uses BLOCK_SIZE constant, it seems to be agnostic to actual BLOCK_SIZE and doesn't requires any modifications.
```
nirs: Correct, but we should probably eliminate the use of BLOCK_SIZE, or replace with 4k to make sure this code does not generate unaligned I/O when using 4k block size.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Benefit to oVirt
+
+* Support for 4k only SAN
```
nirs: We are not sure we want to provide this, since Cinder support will provide this capability. We don't want to invest in legacy block storage.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
...
+* Support for VDO native block size
```
nirs: We can add "Support 4k based file based storage"

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,93 @@
+---
+title: 4k block device support for the storage domains
```
nirs: This is not good name, since may never support 4k for block storage. Lets call this "4k drive support"

source/develop/release-management/features/storage/4k-blocks.md
akashihi: 512b fixed

I mostly talking on a 512 emulation details, not performance.

source/develop/release-management/features/storage/4k-blocks.md
akashihi: I requestd such measurements from the RHHI team, waiting for data.

source/develop/release-management/features/storage/4k-blocks.md
akashihi: VDO:)

Will try to find more brands/specific systems details

source/develop/release-management/features/storage/4k-blocks.md
akashihi: I would like to have same metadata for both block domains and same domains, so we will have same code for both.

In case PHYBLKSIZE isn't used anywhere, we can drop it safely.

source/develop/release-management/features/storage/4k-blocks.md
akashihi: We are going to create new storage domain, let's unify metadata

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Yes, i replaces whole locking section with the new approach

source/develop/release-management/features/storage/4k-blocks.md
akashihi: yep

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Done

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Done

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Done

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Future is not clear yet, let's keep it here for a while

source/develop/release-management/features/storage/4k-blocks.md
akashihi: right

source/develop/release-management/features/storage/4k-blocks.md
akashihi: Done

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+feature_status: Planning
+---
+
+# 4k drive support
+
+# Summary
+
+VDSM assumes, that 512 bytes is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, cause while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM. 
+
+Unfortunately, some 4k block storage devices either refuse to read/write smaller blocks atomically (some SAN appliances) or have a performance impact while emulating 512 bytes block device (like VDO).
```
nirs: We discussed previously that VDO does not support atomic writes in emulation mode. Do we have new info about this or you just forgot to mention it?

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+category: feature
+authors: dchaplyg
+feature_name: 4k drive support
+feature_modules: vdsm
+feature_status: Planning
+---
+
+# 4k drive support
+
+# Summary
```
nirs: Summary is something that come at the end, I think we want "Overview" here.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Owner
+
+*   Name: [Denis Chaplygin](https://github.com/akashihi)
+*   Email: <dchaplyg@redhat.com>
+
+## Current status
+
+*   Target Release: 4.3
+*   Status: Prototyping
```
nirs: This is in design stage - we don't have a design yet. After we learn from the prototype, we will have more info to design this.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+feature_name: 4k drive support
+feature_modules: vdsm
+feature_status: Planning
+---
+
+# 4k drive support
+
+# Summary
+
+VDSM assumes, that 512 bytes is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, cause while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM.
```
nirs: There is no such thing called VDSM. We call it "Vdsm" or "vdsm".

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+
+# 4k drive support
+
+# Summary
+
+VDSM assumes, that 512 bytes is a minimally accessible block on the storage device, which can be written or read atomically. That was true for a long time, but now most of the storage devices changed their block size to the 4096 bytes (4 kilobytes). It is still compatible with VDSM, cause while we are requesting 512 atomic read or write, underlying storage will atomically read 4096 bytes, throw away 3584 bytes and return just 512. As it still acts same way, it has no impact on the VDSM. 
+
+Unfortunately, some 4k block storage devices either refuse to read/write smaller blocks atomically (some SAN appliances) or have a performance impact while emulating 512 bytes block device (like VDO). 
+
+Therefore without 4k block device support oVirt either can't use those devices as a storage domain or will have performance degradation on those domains.
```
nirs: Important stuff missing here, or in the first part of the feature page.

- We want to support 4k for file based storage. This is important for hyperconverge solution
- We don't plan to support 4k drives for legacy block based storage. This will be supported only when using Cinder based storage, planned for 4.3. (This is our official goal, according to Yaniv Lavi).

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+*   Name: [Denis Chaplygin](https://github.com/akashihi)
+*   Email: <dchaplyg@redhat.com>
+
+## Current status
+
+*   Target Release: 4.3
+*   Status: Prototyping
+*   Last updated: August 22, 2018
+
+# Implementation details
```
nirs: I think we should avoid implementation details at this stage of development. What we need instead is description of what will be supported and what will not supported with 4k drives. Information that users will find useful.

For example, how a user will create a storage domain using 4k drives. How the system will be changed to support 4k drives should be at the end, under "implementation" section. At this point it is not important.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+
+ - ALIGN1M | SECTOR512: max_hosts 2000
+ - ALIGN1M | SECTOR4K:  max_hosts 250
+ - ALIGN2M | SECTOR4K:  max_hosts 500
+ - ALIGN4M | SECTOR4K:  max_hosts 1000
+ - ALIGN8M | SECTOR4K:  max_hosts 2000
+
+Thus VDSM may use ALIGN1M | SECTOR512 format for 512 bytes devices, consuming 1MB for each sanlock file. On the 4k block devices VDSM could use ALIGN1M | SECTOR4K or ALIGN2M | SECTOR4K, thus consuming either 1MB or 2MB for each sanlock file. In case of ALIGN2M | SECTOR4K usage, all lockspace files will be bigger, comparing to 512 bytes format. 
+
+With patch [1] it will be possible, to specify desired alignment, but it is not yet available in the python bindings (needs to be implemented).
```
nirs: This patch is not relevant now.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Blocksize detection
+
+There is no reliable way to detect size of the logical block of the underlying device in case of a network filesystem. Both NFS (v3, v4, v4.1, v4.2) and Ceph have no aligment restrictions enforced, therefore they can't be probed for a mimally allowed block IO size.
```
nirs: But we can detect reliably with GlusterFS, so maybe we should take advantage of this to provide better user experience when using this storage type.

Also, we can detect reliably if we cannot detect - if we can 1 byte direct I/O request, it means the storage capabilities cannot be detect, and the user must configure the block size. If the storage support detection, we  can tell the user what are the storage capabilities.

source/develop/release-management/features/storage/4k-blocks.md
```diff
@@ -0,0 +1,165 @@
...
+
+Thus VDSM may use ALIGN1M | SECTOR512 format for 512 bytes devices, consuming 1MB for each sanlock file. On the 4k block devices VDSM could use ALIGN1M | SECTOR4K or ALIGN2M | SECTOR4K, thus consuming either 1MB or 2MB for each sanlock file. In case of ALIGN2M | SECTOR4K usage, all lockspace files will be bigger, comparing to 512 bytes format. 
+
+With patch [1] it will be possible, to specify desired alignment, but it is not yet available in the python bindings (needs to be implemented).
+
+## Blocksize detection
+
+There is no reliable way to detect size of the logical block of the underlying device in case of a network filesystem. Both NFS (v3, v4, v4.1, v4.2) and Ceph have no aligment restrictions enforced, therefore they can't be probed for a mimally allowed block IO size. 
+
+Therefore desired block size for a new storage domain must be specified explicitly, via oVirt UI or API during storage domain creation. StorageDomain.create verb needs to be extended with that information (actually a StorageDomainCreateArguments type needs to be extended)
```
nirs: We may need to extend the api to return info abut the storage domain, so we can provide detection results. The UI can use the results to limit impossible configuration.

source/develop/release-management/features/storage/4k-blocks.md
nirs: We don't want to change block storage in any way, this is risky. So we should have new options for file storage. If we add support for 4k to block storage in the future, we will probably replace the old names with the new names.

source/develop/release-management/features/storage/4k-blocks.md
nirs: We don't need to unify this for supporting 4k for file storage, so not yet.

gregsheremeta:
[ rebased ]

sandrobonazzola:
@nirs any progress on this?

nirs:
@sandrobonazzola we will update this page soon, it requires a rewrite. Maybe it is better to close the PR for now since the partial and possibly stale content may confuse people.

PR #1724, opened 2018-08-17
Security group feature

maiqueb:
Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>

- add a new feature page for supporting security groups on external networks

@maiqueb
@dominikholler


source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+### Direction attribute
+
+The ***direction*** attribute in the ACL has a very important impact in
+the openflow pipeline - in which stage of the pipeline processing will
+the rule be evaluated.
+
+As can be seen below, the OpenFlow pipeline consists of two
+different stages: *ingress* and *egress*.
+
+![of-pipeline](../../../../images/features/network/security-groups/openflow-flow.png)
```
maiqueb: I ought to somehow mention this image was looted from the openvswitch specification.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+every networking-api security group in the list.
+
+![add-port](../../../../images/features/network/security-groups/add_ports.png)
+
+## Provider start-up security groups
+
+### drop-all-ip
+
+To achieve the intended deny-all behavior, a port group will be created
+when the ovirt-provider-ovn starts.
```
maiqueb: well, this can also be done when a port is added....

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+
+* The security group feature should be configurable, and activated by
+default - this means an option for using oVirt's external networks
+without security group support should be possible.
+
+* If the external network has a subnet defined, DHCP traffic is allowed to
+/ from the VM.
+
+### Requirements
+
```
dominikholler: The following requests will be implemented:

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+traffic is allowed.
+
+* The security group feature should be configurable, and activated by
+default - this means an option for using oVirt's external networks
+without security group support should be possible.
+
+* If the external network has a subnet defined, DHCP traffic is allowed to
+/ from the VM.
+
+### Requirements
```
dominikholler: API Requests

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+default, and the user adds rules that allow traffic. As long as *one*
+rule is matched, the traffic is allowed.
+
+Security group rules can match on L3 and L4 parameters, specifically on
+the protocol, the IP protocol, and destination ports. The remote IP
+address is also a subject for the match conditions, and rules can be
+added to allow *incoming* and *outgoing* traffic.
+
+More information on the API of security group rules can be found
+[here](https://developer.openstack.org/api-ref/network/v2/#security-group-rules-security-group-rules).
```
dominikholler: I expected the link already above, but maybe this is just a matter of personal expectations, since the link feels here well positioned, too.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Activating the feature
+
+Two alternatives are possible:
+
+1. Add a configuration flag to the ovirt-provider-ovn configuration file.
+
+2. Use the [Allowed address pairs](https://developer.openstack.org/api-ref/network/v2/#allowed-address-pairs) networking API extension.
+Using this extension, we would be able to individually specify to which ports we would want to activate the 'deny-all' behavior, instead activating it for all ports.
```
dominikholler: There is a third option: to add allow-all rules to existing networks during the upgrade.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+### Encoding the security group information
+
+The security group information will be encoded in the port group object.
+
+The security group has the following attributes:
+
+| name                 | type   | description |
+|----------------------|--------|-------------|
+| tenant_id       | string | The ID of the tenant.                                                      |
+| project_id      | string | The ID of the project.                                                     |
```
dominikholler: Are `tenant_id` and `project_id` aliases and referring to the same data?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+of the corresponding port group.
+
+### Encoding the security group rule information
+
+The security group information will be encoded in the ACL object, to
+which it naturally corresponds in the OVN-world.
+
+A security group rule has the following attributes:
+
+| name                 | type   | description |
```
dominikholler: From my point of view, there is no value in replicating the table from the OpenStack specification. I would be interested in the difference to the OpenStack spec, e.g. information about which attributes are implemented, and which not.

source/develop/release-management/features/network/networking-api-security-groups.html.md
dominikholler: Please ensure that this does not violate copyright.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Relevant events
+
+The relevant for the security group feature are listed and explained
+below.
+
+### Add security group
+
+The *cloud management system* - e.g. ovirt - issues a request to create
+a security group.
```
dominikholler: s/ovirt/oVirt Engine/

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+
+Not implemented:
+* [Tag](https://developer.openstack.org/api-ref/network/v2/#id381)
+
+Not sure:
+* [Allowed address pairs](https://developer.openstack.org/api-ref/network/v2/#allowed-address-pairs) networking API extension.
+
+## Testing
+
+TODO
```
dominikholler: the API could be tested by using an openstack client, e.g. ansible os_security_group or the command line client.
I wonder if there is a way using less effort to check the functionality?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: In openstack, afaik, a tenant and a project are interchangeable terms. 

Why feature both ?.. Good question; probably it is something they want to change in the future.

Never-the-less, I agree that duplicating the tables from the API doesn't add any value; I did it simply to make life easier for the reader.

If you think it isn't easier, say so and I'll get rid of all tables.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,352 @@
...
+| security_group_id    | string | The security group ID to associate with this security group rule.                                                                                                |
+| remote_ip_prefix (*) | string | The remote IP prefix that is matched by this security group rule.                                                                                                |
+| description (*)      | string | A human-readable description for the resource. Default is an empty string.                                                                                       |
+
+(*) - optional attributes
+(**) - optional attributes that ***will not*** be implemented.
+
+### OVN ACL table
+
+An OVN ACL has the following attributes:
```
maiqueb: I think this table is useful, since it is hard to refer the user to the ovn-nb documentation - it is a txt, where - afaik - I can't point to a specific anchor in the file

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I don't think so; ansible looks to be *the* simplest possible way of interacting with this feature.

Having said that, we need to make sure our docs specify very well which subset of the API we support - I think everything except **tags** should be handled.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I removed the list of requests, since, all in all, having it was just dumb; those were just the methods allowed in the networking api specification.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I'm trying to figure out how can I do that...

How can I be sure that linking to the copyright - Copyright © 2015; Open Networking Foundation - is enough?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,468 @@
...
+The user is meant to provision security group(s) and rule(s) through ansible,
+or directly using a REST client.
+
+This greatly reduces the oVirt-engine impact, since there's no need to 
+implement the REST client side of the API, nor implement these changes in the
+GUI.
+
+Throughout the remainder of the document, ansible will be assumed as the client, given its descriptive yaml syntax. The [os_security_group](https://docs.ansible.com/ansible/2.5/modules/os_security_group_module.html) and [os_security_group_rules](https://docs.ansible.com/ansible/2.5/modules/os_security_group_rule_module.html) will be leveraged for this.
+
+### Mandatory group creation
```
maiqueb: i'm considering adding plantuml diagrams for this flow. 

do you think the text alone is easy to understand?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,468 @@
...
+    state: present
+    name: allow-dhcp-<subnet_uuid>
+    description: security group to allow dhcp-traffic
+~~~~~
+
+***NOTE:*** the **deny-all** security group can be provisioned through ansible,
+during the engine-setup stage. the **allow-dhcp** groups will be created as 
+part of the subnet creation workflow.
+
+### Initial rule creation
```
maiqueb: i'm considering adding plantuml diagrams for this flow. 

do you think the text alone is easy to understand?

dominikholler:
I have no idea, I just avoided such situations in the past.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,471 @@
...
+rule is matched, the traffic is allowed.
+
+Security group rules can match on L3 and L4 parameters, specifically on
+the protocol (L3), the IP protocol (L4), and destination ports. The remote IP
+address is also a subject for the match conditions, and rules can be
+added to allow *incoming* and *outgoing* traffic.
+
+The API of security group rules can be found [here](https://developer.openstack.org/api-ref/network/v2/#security-group-rules-security-group-rules).
+
+### Activating the feature
```
mmirecki: Do we need additional configuration for this?
Can the "port_security_enabled" attribute of the port not be used for this? With this we can decide at network/port level.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,471 @@
...
+    description: security group to allow dhcp-traffic
+~~~~~
+
+***NOTE:*** the **deny-all** security group can be provisioned through ansible,
+during the engine-setup stage. the **allow-dhcp** groups will be created as 
+part of the subnet creation workflow.
+
+### Initial rule creation
+
+Since the Openstack Networking API is a white list, the **deny-all** related
```
mmirecki: They could be triggered by setting 'port_security_enabled'

source/develop/release-management/features/network/networking-api-security-groups.html.md
mmirecki: tenant_id - user
project_id - one of many projects belonging to one user

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,471 @@
...
+---
+
+
+# Networking API security group support
+
+
+## Summary
+
+This feature aims to implement OpenStack Networking API [security group](https://developer.openstack.org/api-ref/network/v2/index.html#security-groups-security-groups)
+support to external networks.
```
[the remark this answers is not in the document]
yanivlavi: +1

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,471 @@
...
+* When possible, Openstack neutron's behavior should be mimicked: by
+default, all VM incoming IP *and* outgoing traffic is dropped. This behavior
+is better specified in the [requirements](#requirements) section.
+
+* The security group feature should be configurable, and activated by
+default - this means an option for using oVirt's external networks
+without security group support should be possible.
+
+* If the external network has a subnet defined, DHCP traffic is allowed to
+/ from the VM.
```
yanivlavi: Isn't this the OVN default behaviour.
Seems like a bug to me on OVN if they don't allow this implicitly.

source/develop/release-management/features/network/networking-api-security-groups.html.md
yanivlavi: +1
We should also consider the Neutron behaviour in this.
Meaning before and after you apply the first SG.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: @mmirecki don't think so; a 'user' is a user; a tenant & project are [interchangeable](https://docs.openstack.org/operations-guide/ops-projects-users.html#projects-or-tenants) terms

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,512 @@
+---
+title: Networking API security groups support
```
dankenigsberg: I'd prefer to call this

 `ovirt-provider-ovn security groups`

because "Networking API" is a very clear term in the OpenStack world, but it is very vague in oVirt.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+level. Afterwards, it can be updated through the API - at port *or* network
+levels.
+
+The default behavior for the network's **port-security** attribute would be
+defined in the configuration file, through the ***port-security-enabled***,
+property, located in a new section, called **NETWORK**. This value will be set
+during *engine-setup* - refer to [this section](#engine-setup) - and the
+default choice will be 'yes', which will result in all traffic being dropped to
+*and* from the VMs.
+
```
dominikholler: If security groups are enabled during engine-setup, shouldn't this result in something like the [default config of OpenStack](https://imgur.com/a/2UWPgau)

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+Existent ports can later be updated, disabling the **port-security** attribute, which would remove the ACLs that drop all IP traffic to and from the VM.
+
+As defined in the [network port security attribute definition](https://developer.openstack.org/api-ref/network/v2/#port-security), updating the **port-security** attribute on a network object **does not** cascade the value to ports attached to that network - meaning that effect would only apply to ports created *after* that update.
+
+## User workflow
+
+Since the feature is active by default, the intended workflow - from the oVirt
+engine user's perspective - is:
+
+* create an external network through the provider - its default *port_security_enabled* value is true.
```
dominikholler: "on the ovirt-provider-ovn" ?

dominikholler: default value is true, if activated during engine-setup ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+* create VMs attached to this external network - they will inherit the
+*port_security_attribute* from the network, which will result in all IP traffic
+to *and* from that VM being dropped.
```
dominikholler: I don't think that dropping all traffic from VM is a good default

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+* the intended traffic will have to be white-listed. Assuming the user is interested in allowing ssh traffic *and* all egress IP traffic, the following ansible playbook should be used:
```
dominikholler: I would expect that just incoming request has to be whitelisted.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+    protocol: tcp
+    port_range_min: 22
+    port_range_max: 22
+
+- os_security_group_rule:
+    cloud: ovirt
+    state: present
+    direction: egress
+    ethertype: IPv4
+~~~~~~
```
dominikholler: due lack of knowledge: how is this applied to a port or set as a default of a network?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,582 @@
...
+
+### Installation/Upgrade
+
+The install time required updates are described in the section [above](#engine-setup).
+
+On upgrades with with existing VMs - having attachments to external networks -
+the ports will **not** be updated, and it will be up to the administrator to
```
dominikholler: Are the defaults of existing networks kept?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I understand the concern, and actually even share it. I just went with the flow and mimicked networking-ovn behavior, which is to drop everything *if* port_security is enabled (which it is by default).

The question is a bit of mimicking the networking-ovn behavior versus surprising the user.

I would rather **not surprise** existing users - e.g. set the default to **false** - but would like @yanivlavi opinion on this matter.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Understandable; according to https://wiki.openstack.org/wiki/Neutron/SecurityGroups, in the 'Behavior' section, user created security groups will have that rule automatically created. 

Which would  brings us to the behavior you mention above. 

Take into account that both options are not the same - until the user creates a security group, the vms egress traffic would be blocked vs having it allowed.

Again the question is: what do we want to achieve ? 

@yanivlavi your opinion is highly appreciated here.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I agree with that observation - but propose a twist to it.

Looking at the rules you provide, it looks like **all egress traffic is allowed**,whereas **all ingress** from instances ***belonging*** to the **'default' security group is allowed**.

In this feature page, the implementation of the 'remote_security_group' parameter is discarded, since it would bloat the feature a lot - and this is actually the first motive for including it I've found. Never-the-less, I agree that's a good motive to include it.

I suggest we mimic the egress behavior - allow outgoing traffic - **but** drop all incoming traffic by default.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+## Summary
+
+This feature aims to implement the OpenStack Networking API [security group](https://developer.openstack.org/api-ref/network/v2/index.html#security-groups-security-groups)
+in the ovirt-provider-ovn.
+
+By doing so, we allow fine-grained access control to - and from - the oVirt
+VMs attached to external networks.
+
+The Networking API v2 defines security groups as a white list of rules.
+That means, that by default, neither incoming nor outgoing traffic is
```
MosheSheena: I didn't understand how this relates to the previous line..
I think you can state it as a stand alone fact

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+allowed (from the VMs perspective).
+
+A security group is applied to a logical port - a *logical switch port*
+in the OVN model.
+
+### Owner
+
+* Feature Owner: [Miguel Duarte Barroso](https://github.com/maiqueb) (mbarroso on IRC)
```
MosheSheena: what channel in IRC?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+* Security group: container for security group **rules**; networking API
+entity
+* Security group rules: specify the access to a network resource;
+networking API entity
+
+### Benefit to oVirt
+
+Currently, the only security mechanisms present for oVirt VMs attached to
+external networks are quite crude L2 and L3 protections - MAC
+spoofing, and IP spoofing - implemented directly by Open vSwitch. It is worth
```
MosheSheena: "It is worth mentioning that IP spoofing protection is only applied when a subnet is defined
￼ on top of the Networking API network"
I don't think it's related - it's enough to leave it as is: "implemented directly by Open vSwitch".

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+on top of the Networking API network.
+
+Security groups, which are a traffic white list, complement provided security
+by specifying which traffic is allowed to *and* from the resources - e.g. ports - using high level abstraction L3 and L4 semantics.
+
+Furthermore, security groups allow these rules to be applied in fine
+grained fashion - e.g. the user can specify which rules apply to which
+ports.
+
+Below, you can find some examples of goals for this feature:
```
MosheSheena: I think perhaps "use-cases" can replace "goals" and it will be more concise

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+* **only** allow incoming traffic to a specific VM from a specific
```
MosheSheena: Is the "only" necessary?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+for [security group rules](https://developer.openstack.org/api-ref/network/v2/#security-group-rules-security-group-rules). 
+They are applied to logical ports, and thus are used to associate a set of 
+rules to a port.
+
+They are are indeed a white list - e.g. everything is *denied* by
+default, and the user adds rules that allow traffic. As long as *one*
+rule is matched, the traffic is allowed.
+
+Security group rules can match on L3 and L4 parameters, specifically on
+the protocol (L3), the IP protocol (L4), and destination ports. The remote IP
```
MosheSheena: "the protocol" ? what protocol?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Since the rules are expressed as a white-list, you have to specify which traffic is allowed. In short, it literally is a 'list of allowed stuff'. As opposed to a black-list, where you list the things you block.

That means that if you don't white-list anything, your 'list of allowed stuff' is empty, thus making everything blocked by default.

Makes sense to you ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: hm, I think so; it we don't implement this feature, 'we allow incoming traffic to a specific VM'. 

We want to allow when a set of conditions are met, and reject everything else.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+
+L2 protection is **not in scope** of the Networking API security groups.
+
+## Objectives
+
+* Provide a security group implementation, using OVN as the backend.
+No additional backend / database should be used other than ovn's
+northbound database.
+
+* When possible, Openstack neutron's behavior should be mimicked: by
```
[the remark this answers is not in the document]
maiqueb: The motivation is twofold: we don't want to confuse users that are / have used the Openstack Networking API, which have expectations on the API itself; and we want to refer the users to the openstack documentation whenever possible. And it's only possible if the behaviors are the same.

You want me to add that to the document? I really feel it's unnecessary.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+default, all VM incoming IP *and* outgoing traffic is dropped. This behavior
+is better specified in the [requirements](#requirements) section.
+
+### Requirements
+
+* For ingress traffic (to an instance)
```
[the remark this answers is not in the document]
maiqueb: Correct; I'll use either guest or VM. Instance is the openstack 'term' for it.... I'm mixing stuff :)

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+  + Only traffic matched with security group rules are allowed.
+  + When there are no rules defined, all ingress traffic is dropped.
+* For egress traffic (from an instance)
+  + When there are no rules defined, all egress traffic is allowed.
+* The DHCP traffic traffic for OVN subnets will have to be explicitly allowed,
```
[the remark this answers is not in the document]
maiqueb: Thanks for spotting it.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I agree. 

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,588 @@
...
+otherwise, it will be automatically dropped, as a consequence to the above requirements.
+
+### Security groups description
+
+Security groups, in the general sense, are nothing more than containers
```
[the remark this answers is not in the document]
maiqueb: Symbolic term, you've assumed correctly. 

I'll rework the sentences.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: is is better now?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+otherwise, it will be automatically dropped, as a consequence to the above requirements.
+
+### Security groups description
+
+Security groups, in the general sense, are meant to group together a list of
+[security group rules](https://developer.openstack.org/api-ref/network/v2/#security-group-rules-security-group-rules). 
+They are applied to logical ports, and thus are used to associate a set of 
+rules to a port.
+
+They are are indeed a white list - e.g. everything is *denied* by
```
MosheSheena: Please remove one "are"

MosheSheena: This is not an example, but an initial state of the "list"

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+default, and the user adds rules that allow traffic. As long as *one*
+rule is matched, the traffic is allowed.
+
+Security group rules can match on L3 and L4 parameters, specifically on
```
MosheSheena: What do you mean by "L3 and L4 parameters" ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+added to allow *incoming* and *outgoing* traffic.
+
+The API of security group rules can be found [here](https://developer.openstack.org/api-ref/network/v2/#security-group-rules-security-group-rules).
+
+Security group filters are **only applied to** ports whose
+**port_security_enabled** attribute is set. By default, when a new port is
+created, and the user does not specify this attribute, it is created set
+to 'True', meaning that security group filters will be applied to it.
+
+This causes all incomming traffic to the port to be dropped, until the user
```
MosheSheena: What about outgoing traffic?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+white-lists the intended traffic for that port.
+
+### User interface
+
+The user is meant to provision security group(s) and rule(s) through ansible,
```
MosheSheena: "through ansible" - how? contacting the host? provider?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+levels.
+
+The default behavior for the network's **port-security** attribute would be
+defined in the configuration file, through the ***port-security-enabled***,
+property, located in a new section, called **NETWORK**. This value will be set
+during *engine-setup* - refer to [this section](#engine-setup) - and the
+default choice will be 'yes', which will result in all traffic being dropped to
+the VMs.
+
+Existent ports can later be updated, disabling the **port-security** attribute,
```
MosheSheena: "Existing ports.."

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+    state: present
+    name: <port_name>
+    security_groups:
+        - "{{ my_app_default_security_group.id }}"
+
+~~~~~~
+
+### Installation/Upgrade
+
+The install time required updates are described in the section [above](#engine-setup).
```
MosheSheena: "above" scrolls down and not up.. Not critical but please notice

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+### Engine-Setup
+
+A new ovirt-provider-ovn related question will be added, asking the user if
+port-security is enabled at network level. The default answer will be 'yes'.
+
+This will lead to all the created ports having port-security active, leading to
+all IP traffic being dropped.
+
+## Design feature descrition
```
MosheSheena: Please fix  "descrition" to "description"

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+### Mandatory security group creation
+
+There are 2 types of required security groups, that are created automatically
+by the ovirt-provider-ovn, when required. They are:
```
MosheSheena: I'm confused with "created automatically￼ by the ovirt-provider-ovn" and "when required".. Is it the user's call?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+1. the **deny-all** security group, provisioned when the first port with the
+**port-security** flag activated is created. This port will be the container for
```
MosheSheena: "This port.." - did you mean "this group"?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+Since the Openstack Networking API is a white list, the **deny-all** related 
+rules cannot be triggered from the API itself.
+
+To avoid the need to create the rules beforehand, these ACLs will be
+provisioned when the first port having the *port-security* attribute enabled is
+created.
+
+The DHCP related rules will be provisioned after the respective group is
+created, and as part os the [create subnet](#creating-a-subnet-on-top-of-a-network)
```
MosheSheena: Please change "part os.." to "part of.."

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+The [ansible module](https://docs.ansible.com/ansible/2.5/modules/os_port_module.html) currently does not support updating the *port_security_enabled* attribute for a port.
+
+To provide the user an integrated experience, ansible should allow for that attribute to be updated.
+
+A bug has been created to track this feature request. Its state can be followed
+[here](https://bugzilla.redhat.com/1623993).
+
+## Mapping networking API to OVN model objects
+
+The security group data will be model as a [port group](https://github.com/openvswitch/ovs/blob/master/ovn/ovn-nb.xml#L926).
```
MosheSheena: Please change to "will be modeled.."

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+* create an external network on the ovirt-provider-ovn - its default
+*port_security_enabled* value is true - which can be overriden during engine-setup.
+* create VMs attached to this external network - they will inherit the
+*port_security_attribute* from the network, which will result in dropping all
+IP traffic to that VM.
+* the intended traffic will have to be white-listed. Assuming the user is
+interested in allowing ssh traffic, the following ansible playbook should be
+used, which creates one security group - for which a security group rule
+allowing all **egress** traffic is automatically created -, plus one security
```
MosheSheena: "-," - please remove one of them

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+| match     | string | The match criteria for the ACL. All the L3-L4 information (protocol, ether protocol, ports, ip prefix) is encoded here. Directly relates to OVS matches. |
+| action    | string | Either 'allow', 'allow-related', 'drop', or 'reject'. Currently, 'reject' behaves as 'drop'.                                                             |
+
+### Mapping the rules
+
+Mapping is a two way game: it is needed to translate *from* a networking
+API security group rule *to* an OVN ACL, ***and*** from an OVN ACL to a
+networking API security group rule.
+
+To achieve the latter, the information that is not **directly** encoded
```
MosheSheena: I would empahsize the "not" instead / with the "directly"

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+![add-subnet](../../../../images/features/network/security-groups/create_subnet.png)
+
+### Attach a VM to an external network.
+
+Please refer to the sequence diagram below to better understand this
+flow.
+
+When a VM is attached to an external network, an ovn logical switch port
+is created. This port concept is already implemented in the
+ovirt-provider-ovn project, *but* does not allow it to be bound to
```
MosheSheena: I would emphasize the "not", but it's not a must

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+security groups. That support will need to be implemented.
+
+Whenever a port is created, *if* it has the **port-security** flag
+activated, the ovirt-provider-ovn will have to update the **deny-all**
+port group data, adding the new port to its list of ports. *If* that
+port group does not exist, it will have to be created, and the
+corresponding ACLs added to it. This **port-security** attribute is
+currently not supported in the ovirt-provider-ovn, and thus, will also
+have to be added to the API.
```
MosheSheena: Is there a Bug? or an RFE? can the progress be tracked somewhere?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+*If* the port is attached to a network featuring a subnet, the
+corresponding **allow-dhcp** port group port list will also be updated
+with the new port.
+
+It will also need to update the list of ports for each of the port
```
MosheSheena: It refers to 'ovirt-provider-ovn'?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+  * direction: to-lport
+  * match: outport == @<port_group> && ip
+
+**NOTE:** these rules will be filtered out by the API, because they are simply
+not featured in the Networking API. This means that a user listing all security
+group rules **will not** see the rules corresponding to the ACLs defined above.
+
+### allow-dhcp
+
+To allow a VM to get an IP address from the subnet on top the network
```
MosheSheena: "on top of the network"

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+    * priority: ALLOW_PRIORITY
+    * action: allow
+    * direction: to-lport
+    * match: outport == @allow-dhcp<subnet_uuid> && ip4 && ip4.src == <subnet_cidr> && udp && udp.src == 67 && udp.dst == 68
+
+***Note:*** the '@' sign is a port group tag understood by OVN
+
+## ovirt-provider-ovn API update
+
+The following updates are required to the current API, listed by entities:
```
MosheSheena: "...required for the..."

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+* network
+  + add the **port_security_enabled** boolean attribute
+* port
+  + implement the **port_security_enabled** boolean attribute
+  + implement the **security_groups** array attribute
```
MosheSheena: Specifying a data struct such as 'array' might be limiting.. perhaps it will be better to say the attribute needs to be implemented.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+* security group rules
+    + implement this entity, as described [here](#encoding-the-security-group-rule-information)
+
+## Networking API extensions implemented / not implemented
+
+Implemented:
+* [Resource timestamps](https://developer.openstack.org/api-ref/network/v2/#id367)
+* [Port security](https://developer.openstack.org/api-ref/network/v2/#id35)
+
+Not implemented:
```
MosheSheena: Do we want to leave it as is, or state if we intend on implementing / not supporting?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+* [Tag](https://developer.openstack.org/api-ref/network/v2/#id381)
+
+## Feature dependencies
+
++ Open vSwitch 2.10 - port group table
++ python-ovsdbapp 0.12.1 - port group support, with a nice API, without
```
MosheSheena: "With a nice API" is not objective - I think it should either be removed or placed within parentheses

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+openstack dependencies
+
+## Testing
+
+The recommended tools to check this feature are ansible and ncat.
```
MosheSheena: Perhaps refer to http://man7.org/linux/man-pages/man1/ncat.1.html about ncat

source/develop/release-management/features/network/networking-api-security-groups.html.md
MosheSheena: Now it does - but it's not so trivial to understand that having an entity such a security groups that is in fact a white list means that by default all traffic is not allowed.

source/develop/release-management/features/network/networking-api-security-groups.html.md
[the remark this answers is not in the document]
MosheSheena: you're right I don't think it will add any value to the paper

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I'm literally listing them in the next sentence

*the L3 protocol, the L4 protocol, and destination ports. The remote IP
address is also a subject for the match conditions*

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: As per the **requirements** section: 

* For egress traffic (from an instance)
  + Only traffic matched with security group rules is allowed.
  + When there are no rules defined, all egress traffic is dropped.
  + **Rules allowing all egress IP traffic are automatically added to all groups.**

I'll add a sentence referring to it; hopefully it'll make it clearer.

Is this better ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Good point; assumptions are really bad.

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+Since the feature is active by default, the intended workflow - from the oVirt
+engine user's perspective - is:
+
+* create an external network on the ovirt-provider-ovn - its default
+*port_security_enabled* value is true - which can be overriden during engine-setup.
+* create VMs attached to this external network - they will inherit the
+*port_security_attribute* from the network, which will result in dropping all
+IP traffic to that VM.
+* the intended traffic will have to be white-listed. Assuming the user is
+interested in allowing ssh traffic, the following ansible playbook should be
```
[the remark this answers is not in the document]
maiqueb: used where ever you want:; just point it at the ovirt-provider-ovn. 

That can be deployed any way the user wants - but in the ovirt assisted deploy, we assume it is deployed in the same host as the engine.

I don't think this belongs here, since the user could be using this without ovirt.

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+    state: present
+    security_group: "{{ my_app_default_security_group.id }}"
+    direction: ingress
+    protocol: tcp
+    port_range_min: 22
+    port_range_max: 22
+
+~~~~~~
+
+* finally, the user is required to update the ports, indicating which security
```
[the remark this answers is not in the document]
maiqueb: Updated it from playbook to tasks, which is what those actually are.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Nice spotting this one; I have moved the section to a different place and didn't realize that.

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Hm, no. It is done automatically by the ovirt-provider-ovn,when required. 

As indicated in the bullets below:
**deny-all group**: the **deny-all** security group, provisioned when the first port with the 
**port-security** flag activated is created

**allow-dhcp** group: Creating a subnet on top of an external
network will provision an **allow-dhcp** security group **for that particular
network**

I'll rework this a bit, see if it becomes clearer.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Yes. 

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I agree. 

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+The newly created ACL is then stored in the ovn-nb database, and its
+data is afterwards translated back into the networking-api syntax.
+
+The security group rules also feature the resource timestamps extension,
+but the revision number will always be *one*, since we do not allow a
+rule to be updated.
+
+A sequence diagram where a user adds - through ansible - a security
+group rule meant to white list ssh traffic is added to the security
```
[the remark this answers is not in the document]
maiqueb: it is; thanks for spotting it.

Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+group 'allow-ssh' is shown below.
+
+![add-rule](../../../../images/features/network/security-groups/create_security_group_rule_ansible.png)
+
+### Creating a subnet on top of a network
+
+Creating a Networking API subnet leads to the creation of its
+corresponding DHCP options object in the OVN north database.
+
+If the feature is activated - e.g. the **deny-all** port group is
```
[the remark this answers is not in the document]
maiqueb: Yes and yes.

Reworked this sentence; is it better now?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+present in the OVN northbound database - the DHCP related traffic will
+be dropped by Open vSwitch. To prevent that, thus assuring that the VMs
+can get an IP address through DHCP, a security group meant to hold the
+rules white listing DHCP traffic is created, and ACLs allowing that
+traffic added to it.
```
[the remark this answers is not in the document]
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+
+The following sequence diagram depicts the flow described above.
+
+![add-subnet](../../../../images/features/network/security-groups/create_subnet.png)
```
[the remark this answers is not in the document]
maiqueb: I've reworked the text; is this understandable now?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I prefer it as is.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: it's merged already. 

no bug has been filed; I do have a bug for the security group feature itself, and this is the neutron way of checking if security groups apply. 

Want me to refer to that bug - sec. group feature RFE ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Yes, but that's not important, I've updated the text a bit; is it better now?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I disagree, never the less have to agree that it sounds weird. 

Is it better now?

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Removed all types.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
dominikholler: Sounds good

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+~~~~~~
+
+### Installation/Upgrade
+
+The install time required updates are described in the section [below](#engine-setup).
+
+On upgrades with existing VMs - having attachments to external networks -
+the ports will **not** be updated, and it will be up to the administrator to
+manually activate port security in the existent ports.
+
```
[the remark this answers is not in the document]
maiqueb: Very valid point; I think that since at the moment neither [os_network](https://docs.ansible.com/ansible/2.5/modules/os_network_module.html) nor [os_port](https://docs.ansible.com/ansible/2.6/modules/os_port_module.html) ansible modules allow updating the **port_security_enabled** attribute, we should provide an upgrade script that uses the ovn-nb CLI to set those values (port_security at network level) to false.

sounds  like a feasible option ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,597 @@
...
+or using a REST client, targeting the ovirt-provider-ovn.
+
+This greatly reduces the oVirt-engine impact, since there's no need to 
+implement the REST client side of the API, nor implement these changes in the
+GUI.
+
+Throughout the remainder of the document, ansible will be assumed as the client, given its descriptive yaml syntax. The [os_security_group](https://docs.ansible.com/ansible/2.5/modules/os_security_group_module.html)
+and [os_security_group_rules](https://docs.ansible.com/ansible/2.5/modules/os_security_group_rule_module.html)
+will be leveraged for this.
+
```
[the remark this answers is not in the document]
maiqueb: I'll update this section referring ManageIQ as the GUI.

source/develop/release-management/features/network/networking-api-security-groups.html.md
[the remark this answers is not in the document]
dominikholler: Why not just let the ovirt-provider-ovn interpret 'undefined' for networks in the db as port_security_enabled=`false` ?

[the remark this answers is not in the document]
maiqueb: Even nicer.

To sum it up, when new networks are created - and port security is not provider - I'll set the value from the configuration file (created from the user answers to engine-config); for existing networks, a default of False will be assumed.

This will lead to OVN networks created without using the ovirt-provider-ovn using this 'False' default, meaning no security groups will apply to their ports.

Implementing it is a matter of simply changing [this line](https://gerrit.ovirt.org/#/c/93935/11/provider/neutron/neutron_api.py@1158) back to the 'False' default.

Do you foresee any issue with that approach ?

source/develop/release-management/features/network/networking-api-security-groups.html.md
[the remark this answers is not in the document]
maiqueb: Done

source/develop/release-management/features/network/networking-api-security-groups.html.md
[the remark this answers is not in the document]
dominikholler: Sounds good.

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I've reworked it a bit; I am forced to agree I missed a premise there in the explanation.

Let me know if it's OK now.

source/develop/release-management/features/network/networking-api-security-groups.html.md
MosheSheena: Yes, much better +1

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,621 @@
...
+## Documentation & External references
+
+[OpenStack Networking API specification](https://developer.openstack.org/api-ref/network/v2/)
+
+[oVirt OVN integration](https://ovirt.org/develop/release-management/features/network/external-network-provider/)
+
+[oVirt Provider OVN](https://www.ovirt.org/develop/release-management/features/network/ovirt-ovn-provider/)
+
+[Track feature status in Trello](https://trello.com/b/2WpFZANY/security-groups)
```
dominikholler: outdated

source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -0,0 +1,619 @@
...
+
+The security group information will be encoded in the ACL object, to
+which it naturally corresponds in the OVN-world.
+
+A security group rule has the following attributes:
+
+| name                 | type   | description |
+|----------------------|--------|-------------|
+| remote_group_id (**) | string | The remote group UUID to associate with this security group rule. You can specify either the  remote_group_id or remote_ip_prefix attribute in the request body. |
+| direction            | string | Ingress or egress, which is the direction in which the security group rule is applied.                                                                           |
```
dominikholler: Why will be the `remote_group_id` not implemented, how is this solved in neutron?

source/develop/release-management/features/network/networking-api-security-groups.html.md
mmirecki: true, I was not specific enough
tenant_id - organization
user - user
project_id - one of many projects belonging to tenant

source/develop/release-management/features/network/networking-api-security-groups.html.md
maiqueb: I'll make another PR to the feature page, updating it with information about the ```remote_group_id``` attribute.

PR #1731, opened 2018-08-28
remove obsolete RH tracking

duck-rh:


duck-rh:
as decided with the infra team

PR #1769, opened 2018-10-05
Add information about implementing the remote_group_id parameter

maiqueb:
Since that was the only misalignment we had to the security group
rules networking API, the tables can be removed.

Address set information is added.

@maiqueb 

@mmirecki 
@dominikholler

### DIFF
--- a/source/develop/release-management/features/network/networking-api-security-groups.html.md
+++ b/source/develop/release-management/features/network/networking-api-security-groups.html.md
@@ -291,41 +291,17 @@ group of ports. More information can be found [here](https://docs.openstack.org/
 
 The security group information will be encoded in the port group object.
 
-The security group has the following attributes:
+The description of the security group parameters is located at the [networking api](https://developer.openstack.org/api-ref/network/v2/index.html#security-groups-security-groups).
 
-| name                 | type   | description |
-|----------------------|--------|-------------|
-| tenant_id       | string | The ID of the tenant.                                                      |
-| project_id      | string | The ID of the project.                                                     |
-| description (*) | string | A human-readable description for the resource. Default is an empty string. |
-| name            | string | Human-readable name of the resource.                                       |
-
-(*) - optional attributes
-
-All of the aforementioned attributes will be stored in the external_ids
-of the corresponding port group.
+All the security group attributes will be stored in the corresponding port
+group external ids.
 
 ### Encoding the security group rule information
 
 The security group information will be encoded in the ACL object, to
 which it naturally corresponds in the OVN-world.
 
-A security group rule has the following attributes:
-
-| name                 | type   | description |
-|----------------------|--------|-------------|
-| remote_group_id (**) | string | The remote group UUID to associate with this security group rule. You can specify either the  remote_group_id or remote_ip_prefix attribute in the request body. |
-| direction            | string | Ingress or egress, which is the direction in which the security group rule is applied.                                                                           |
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/network/networking-api-security-groups.html.md
```diff
@@ -389,6 +365,66 @@ changed through an ACL - since it would have a lower priority.
...
+
+The address sets referenced above would look like:
+```
+ovn-nbctl create Address_Set name=set1 addresses='10.0.0.1 10.0.0.2 10.0.0.3'
+ovn-nbctl create Address_Set name=set2 addresses='2001:db8:0:0:0:0:2:1 2001:db8:0:0:0:0:3:1 2001:db8:0:0:0:0:14:1'
+```
+
+A reference to the associated port group will be stored on the address set's
+external ids.
+
```
dominikholler: Does this mean you plan to map the remote_group_id only on L3 addresses?
What would happen if there are no OpenStack subnets?
Would be a possible implementation to define the remote_group_id on L2/eth addresses?

dominikholler: OK, you convinced me, that security groups without subnets are quite useless, so L3 appropriate here.

maiqueb: That might indeed by a better solution; it is supported (address sets allow eth addresses), and simplifies the scenarios where IP addresses are assigned through DHCP.

One thing that can probably render this possibility useless is that the **remote_group_id** is meant to be used across different networks, and our current provider implementation does not assure that mac addresses are unique (when statically assigned by the user):
```shell
 ✘ mduarted@localhost ~/github/sec_group_demo master ●  vagrant ssh -c 'sudo ovn-nbctl show '
switch cad16023-91ae-4c90-8e8d-1bbf54b58686 (ovirt-demo_net-3724ecff-4cfa-40f3-bec5-6f5aeceaa263)
    port be3f0ce9-f6da-428b-822b-1eb839342e5c
        addresses: ["00:00:aa:00:dd:01 192.168.10.20"]
    port dd7ea492-a7cd-42ea-988c-c80ef33b20e7
        addresses: ["00:00:aa:00:dd:01 192.168.10.10"]
```
The user can assign the same mac to different ports in same networks - which would render L2 networks unusable, but, if he assigns same macs to different ports, in **different** networks, the problem would be located at security group level - we wouldn't be able to discern between both of them.

maiqueb:
There are a set of issues here, that might constrain the actual solution. I'll list the alternatives below:
- Use address sets L3 - one address set for ipv4, another for ipv6
Having a solution based on subnets has the obvious problem of *relying* on subnets. This wouldn't be an issue since security groups - as defined in the networking API - only allow L3 and L4 semantics.
Another issue, that worries me more than that, is that the entire solution relies on the provider being able to track the port IPs. When VMs get the IP address asynchronously through DHCP, I'm not sure on how to track that change - polling every X seconds? Not cool. Have a daemon to that? Even worse. Or not...

- Use L2 address sets
Since our provider implementation has a single L2 address pool, when the MAC address is not statically provided, we're safe. Problem comes when the user statically provisions the same MAC address in **different** networks - which should be perfectly fine. It would allow traffic from unwanted sources here, which, is a problem. 

- Match directly on port groups - do not use address sets
This would make things a lot simpler, and we wouldn't have to introduce address sets at all. 
It wouldn't rely on subnets *at all*, and would work independently of the IP version type.
Not 100% this is possible - **should be**. Off to testing if it is.

PR #1899, opened 2019-02-10
dbutils: Document the secret dbutils directory

nirs:
The page mention various tools but forget to tell where they are hidden.

### DIFF
--- a/source/develop/developer-guide/db-issues/helperutilities.html.md
+++ b/source/develop/developer-guide/db-issues/helperutilities.html.md
@@ -5,6 +5,10 @@ authors: emesika
 
 # Helper Utilities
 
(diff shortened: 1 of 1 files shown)


danielerez:
Looks good.

didib:
Thanks for the patch. Both docs and code need more love, including making all of them use generate-pgpass.sh/generatePgPass and clarify that credentials default to configured ones...

Also it's worth noting that dbutils is part of setup, but generate-pgpass.sh is part of dbscripts. I guess this is why dbfunc-base.sh is copied to here and not used from dbscripts. The reason is that if you setup version X, and then upgrade the setup packages, dbutils will be from the new version, but dbscripts from the current (old).

PR #1921, opened 2019-02-19
Update chap-Storage.html.md - firmer suggestion to put data under /var

mattsimonsen:
Given the warning about having data outside of "/" and in /var for oVirt node, I recommend we make it harder for somebody to poorly read the documentation and put data in / by accident, if the LV based install is in /var/local

Also a note on what will happen seems prudent, given how bad it would be for oVirt node images to disappear.

This pull request needs review by: @rollandf


gregsheremeta:
@rollandf Fred, your thoughts on this change?
@tnisan if you're interested :)

rollandf:
@yuvalturg Can you take a look?

yuvalturg:
I don't think the VMs should not disappear upgrade - there's a test in the postinstall script of the image-update rpm that fails the upgrade if it finds__DIRECT_IO_TEST__ on the root filesystem [1].  Also, regarding /var/local, if the local VMs reside on a different partition (in the example it's under a seperate LV in /data), the update mechanism shouldn't touch those.

If there's a bug in any of these places, we should probably file a bug report.

[1] https://gerrit.ovirt.org/#/c/88313/

srihosting:
I'm glad that there is now a check to prevent images from being lost- as this happened to us early in testing. My changes may not be needed, though personally I'd still be most comfortable pointing folks to /var/local rather than anything within the root file-system, just in case that check failed in the future.

mattsimonsen:
Doesn't appear this was needed

PR #1963, opened 2019-03-24
Update vdsm build instructions for developers

nirs:
This change depends on the new --enable-timestamp configure option added
in https://gerrit.ovirt.org/c/98830.

With this option, a developer can use "dnf upgrade *.rpm" to install the
latest build.

@dankenigsberg @rollandf please review.


nirs:
No objection so far, time to merge.

PR #1965, opened 2019-03-27
Cinderlib: Add manual steps for installation

rollandf:

- Add manual steps needed to install the cinderlib libraries

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @rollandf 

This pull request needs review by: @bennyz @shenitzky

### DIFF
--- a/source/develop/release-management/features/storage/cinderlib-integration.html.md
+++ b/source/develop/release-management/features/storage/cinderlib-integration.html.md
@@ -267,6 +267,50 @@ Cinderlib will be packaged in RPM.
 The Cinder drivers and Cinderlib needs to be installed in the Engine.
 OS-Brick needs to installed on the hosts
 
+#### Manual Install
+
+Since that some of the required libraries are currently not part of oVirt installation, the following steps are required:
+
+On Engine:
+
+```
+$ yum install -y centos-release-openstack-pike
+$ yum-config-manager --enable openstack-pike
+$ yum install -y openstack-cinder python-pip
+$ pip install cinderlib
+```
+
+On all the hosts: (Vdsm needs to be restarted after installing the packages)
+
+```
+$ yum install -y centos-release-openstack-pike
+$ yum-config-manager --enable openstack-pike
+$ yum install -y python2-os-brick
+```
+
+
+For Ceph backend, the following package is needed in the Engine and in all the hosts:
+
+```
+$ yum install -y ceph-common
+```
+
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/storage/cinderlib-integration.html.md
```diff
@@ -267,6 +267,32 @@ Cinderlib will be packaged in RPM.
...
+$ yum-config-manager --enable openstack-pike
+$ yum install -y openstack-cinder python-pip
+$ pip install cinderlib
+```
+
+On all the hosts: (Vdsm needs to be restarted after installing the packages)
+
+```
+$ yum install -y centos-release-openstack-pike
+$ yum install -y python2-os-brick
```
nirs: Don't we need to enable the repo like we do on engine?

If we have packages in yum, why we don't require them in vdsm?

source/develop/release-management/features/storage/cinderlib-integration.html.md
```diff
@@ -285,4 +311,5 @@ OS-Brick needs to installed on the hosts
 
 ### Documentation / External references
 [1] [Cinder Library](https://github.com/Akrog/cinderlib)
+
```
nirs: Looks like unrelated and unintended change

source/develop/release-management/features/storage/cinderlib-integration.html.md
```diff
@@ -267,6 +267,32 @@ Cinderlib will be packaged in RPM.
...
+$ yum install -y centos-release-openstack-pike
+$ yum install -y python2-os-brick
+```
+
+
+For Ceph backend, the following package is needed in the Engine and in all the hosts:
+
+```
+$ yum install ceph-common
+```
```
nirs: What about Fedora? maybe packaging is in better shape on Fedora?

avihaie: Looks good , just 2 comments:

1) About installing 'ceph-common' on the hosts, do we need any restart (VDSM) afterwards? if so please add that restart is required.
2) Perhaps its best to change all 'install' to 'yum install -y'.

source/develop/release-management/features/storage/cinderlib-integration.html.md
rollandf: The package is available in yum only with the openstack repo installed.
We will need to have them available via oVirt channel.

PR #1985, opened 2019-04-17
Add feature page for network check

dominikholler:
reviewers:
@tiraboschi
@mburman5
mavital
@mmirecki


source/develop/release-management/features/sla/hosted-engine-network-check.html.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Alternative Network Check for Self Hosted Engine
```
doron-fediuck: Why alternative? DNS should be the default and ICMP a fallback.

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
```diff
@@ -0,0 +1,115 @@
...
+~~~~
+
+The related variables in the answer file will be
+`OVEHOSTED_NETWORK/network_test`, `OVEHOSTED_NETWORK/network_test_tcp_address`
+and `OVEHOSTED_NETWORK/network_test_tcp_port`.
+
+## Modification of cockpit-ovirt
+
+cockpit-ovirt is going to be extended to ask and check the additional
+parameters like this:
```
doron-fediuck: 1. Please make sure DNS is the default.
2. Please remove the "todo" from the image.

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
```diff
@@ -0,0 +1,115 @@
...
+## Owner
+
+*   Name: [Dominik Holler](https://github.com/dominikholler)
+
+*   Email: <dholler@redhat.com>
+
+
+## Detailed Description
+
+Additional ways to check the network liveliness to be added are DNS, TCP and
```
mykaul: Both DNS and TCP need to be supported with IPv6.

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
```diff
@@ -0,0 +1,115 @@
...
+none.
+
+DNS means that an NS query for "." (the root) will be performed, if this
+succeeds, network is considered as alive.
+
+TCP means, that a TCP connection to a given host and port combination will be
+created. If this succeeds, network is considered as alive.
+
+None means, that the network is always considered alive.
```
sabose: Is that ever the case, I mean is this safe to be provided as a choice?

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
```diff
@@ -0,0 +1,115 @@
...
+~~~~
+          Please specify which way the network connectivityshould be checked (ping, dns, tcp, none) [dns]: tcp
+          Please specify the desired destination IP addressof the TCP connection test: sdf
+          Please specify the desired destination TCP portof the TCP connection test: 434
+[ ERROR ] Failed to connect via TCP
+~~~~
+
+The related variables in the answer file will be
+`OVEHOSTED_NETWORK/network_test`, `OVEHOSTED_NETWORK/network_test_tcp_address`
+and `OVEHOSTED_NETWORK/network_test_tcp_port`.
```
sabose: is port mandatory? What's the value in case of ICMP?

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
dominikholler: I added the two sentences "Like ping, the DNS and TCP check should work for IPv4 and IPv6." and "Even there is nothing special about IPv6 in this feature, it might be useful to
verify on IPv6."

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
dominikholler: @tiraboschi can you please jump in here?

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
dominikholler: I added "`DNS` has to be the pre-selected default value." and updated the image.

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
dominikholler: I added "while the latter two are only relevant if the value of OVEHOSTED_NETWORK/network_test is tcp."

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
tiraboschi: sanlock (volume lease) is used to protect the engine VM disk at storage level: we are sure that two copies of the same VM acting on the same disk could not be started even if the two hosts cannot communicate over the network so from this point of view it's absolutely safe.

On the other side, if the network on the host is not properly working (various possible issues here with different effects) nothing will be going to restart the engine VM over a different host and so we are not 100% sure that the user can really reach the engine VM from his laptop.

tiraboschi:
One missing section is about how to configure this on existing deployments:
```
[root@c76he20190415h1 ~]# hosted-engine --set-shared-config network_test tcp --type=he_local
[root@c76he20190415h1 ~]# hosted-engine --set-shared-config tcp_t_port 192.168.1.115 --type=he_local
[root@c76he20190415h1 ~]# hosted-engine --set-shared-config tcp_t_port 22 --type=he_local
[root@c76he20190415h1 ~]# systemctl restart ovirt-ha-broker ovirt-ha-agent
[root@c76he20190415h1 ~]# tail -f /var/log/ovirt-hosted-engine-ha/broker.log
Listener::INFO::2019-04-17 14:13:02,712::monitor::78::ovirt_hosted_engine_ha.broker.monitor.Monitor::(start_submonitor) Started submonitor storage-domain
Thread-5::INFO::2019-04-17 14:13:02,913::engine_health::94::engine_health.EngineHealth::(action) VM not on this host
Listener::WARNING::2019-04-17 14:13:03,029::status_broker::153::ovirt_hosted_engine_ha.broker.status_broker.StatusBroker::(lock_host_id) Host id is not locked, trying to lock
Thread-3::INFO::2019-04-17 14:13:03,041::mem_free::51::mem_free.MemFree::(action) memFree: 5084
Thread-2::INFO::2019-04-17 14:13:03,637::mgmt_bridge::62::mgmt_bridge.MgmtBridge::(action) Found bridge ovirtmgmt with ports
Thread-1::INFO::2019-04-17 14:13:04,894::network::88::network.Network::(action) Successfully verified network status
Thread-4::INFO::2019-04-17 14:13:12,725::cpu_load_no_engine::110::cpu_load_no_engine.CpuLoadNoEngine::(calculate_load) VM not on this host
Thread-4::INFO::2019-04-17 14:13:12,726::cpu_load_no_engine::126::cpu_load_no_engine.CpuLoadNoEngine::(calculate_load) System load total=0.0380, engine=0.0000, non-engine=0.0380
Thread-5::INFO::2019-04-17 14:13:12,927::engine_health::94::engine_health.EngineHealth::(action) VM not on this host
Thread-2::INFO::2019-04-17 14:13:13,585::mgmt_bridge::62::mgmt_bridge.MgmtBridge::(action) Found bridge ovirtmgmt with ports
Thread-1::INFO::2019-04-17 14:13:15,072::network::88::network.Network::(action) Successfully verified network status
```
with
`hosted-engine --set-shared-config network_test dns --type=he_local`
we can change the test type on a specific host.

with
`hosted-engine --set-shared-config network_test dns --type=he_shared`
we can change the test type as configured on the shared storage (it will be used a template for the hosts that are going to be added in the future).

dominikholler:
> One missing section is about how to configure this on existing deployments:
> 

Thanks, you told about this and I fogot to mention this!
Done.

nsednev:
What will happen if ping, dns or tcp will fail, should deployment report with an error and ask to provide another working option, or it'll break out from the deployment?

There are 4 options that will be implemented now (none, ping, dns and tcp), I have a questions regarding TCP session option:

"TCP means, that a TCP connection to target host and port combination will be created. If this succeeds, network is considered as alive." 

What component will establish TCP session, would it be hosted-engine-ha-agent component on source ha-host? 
How destination host will "know" that it has to establish the TCP session and which application will "listen" on destination port?

"Please specify which way the network connectivityshould be checked (ping, dns, tcp, none) [dns]: tcp
          Please specify the desired destination IP addressof the TCP connection test: sdf
          Please specify the desired destination TCP portof the TCP connection test: 434"

Why not to ask customer to simply provide one question with 2 variables separated by ":" (destination host and destination port), like "1.1.1.24:88 or someone.com:88?
Would destination IP address of the TCP connection be tested for canonical FQDN if provided? "sdf" is not a canonical FQDN, hence it should have been reported as an error.

Customer will provide socket during deployment. What exactly the code should check for socket, TCP state of the session? Which state should be considered as matching for "alive" connection? 
I understand that TCP session status will also checked after deployment is complete, as network keep alive mechanism for network status on ha-host. 
If TCP will make re-transmissions, in cases when connectivity was interrupted? How host's status will be influenced by TCP re-transmissions?

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
nsednev: > is port mandatory? What's the value in case of ICMP?
ICMP is not an application, its a layer 3 protocol, it's protocol number is 1 and its acting at network layer, hence it doesn't has any port. Ports are for transport layer 4.

dominikholler:
> What will happen if ping, dns or tcp will fail, should deployment report with an error and ask to provide another working option, or it'll break out from the deployment?
> 

I added a paragraph to "Modification of ovirt-hosted-engine-setup" to
answer this:
"The related input is validated instantly. On invalid input, an error is
reported. If the invalid input is read from the answer file, the deployment will
stop, if the invalid input is provided interactively, the wizard repeats the
question after presenting the error message."

> There are 4 options that will be implemented now (none, ping, dns and tcp), I have a questions regarding TCP session option:
> 
> "TCP means, that a TCP connection to target host and port combination will be created. If this succeeds, network is considered as alive." 
> 
> What component will establish TCP session, would it be hosted-engine-ha-agent component on source ha-host? 

yes, in project ovirt-hosted-engine-ha

> How destination host will "know" that it has to establish the TCP session and which application will "listen" on destination port?
> 

Not at all.
I added
"The configuration if this given host is out of scope of this feature."
to the detailed description to highlight this.

> "Please specify which way the network connectivityshould be checked (ping, dns, tcp, none) [dns]: tcp
>           Please specify the desired destination IP addressof the TCP connection test: sdf
>           Please specify the desired destination TCP portof the TCP connection test: 434"
> 
> Why not to ask customer to simply provide one question with 2 variables separated by ":" (destination host and destination port), like "1.1.1.24:88 or someone.com:88?

From my personal point of view, this is more complex that asking two
simpler questions with two simpler answers.
This would mean that we have to explain the user the desired input
format and parse the input to extract the values.
E.g. think how two input an TCP port on an IPv6 address.

> Would destination IP address of the TCP connection be tested for canonical FQDN if provided? "sdf" is not a canonical FQDN, hence it should have been reported as an error.
> 

No, "sdf" might be a valid host name.

> Customer will provide socket during deployment. 

Do you mean a hostname and TCP port combination?

> What exactly the code should check for socket, TCP state of the
> session? 

TCP means, that a TCP connection to a given host and port combination will be created.

> Which state should be considered as matching for "alive"
> connection? 

If the connection is successfully created.

> I understand that TCP session status will also checked after deployment is complete, as network keep alive mechanism for network status on ha-host. 
> If TCP will make re-transmissions, in cases when connectivity was interrupted? How host's status will be influenced by TCP re-transmissions?
> 

The idea is just the check if a TCP connection can be successfully
created, which avoids all this questions.

nsednev:
Still not clear to me, what component is responsible for establishing TCP session and to where? TCP session requires client-server connectivity, where client initiates session to target socket (socket is combination of IP address and port). hosted-engine-ha-agent component on source ha-host will create TCP session to where?
Is it possible to interactively ask customer, instead of stopping deployment, in case of wrong data provided during deployment from answer file? It'll be better to continue, instead of end up with dirty environment.
I'm not sure if usage of hostnames is a good idea, hostnames are names of the servers or VMs within the same LAN, but they're local names and won't be reachable from external networks, hence I think that DNS option should be in FQDN format, hostname+domain name.

dominikholler:
> Still not clear to me, what component is responsible for establishing TCP session and to where?

Without this feature, ovirt-hosted-engine-ha broker pings cyclically
the gateway to check network connectivity.
This feature is about to replace this ping with alternatives.


> TCP session requires client-server connectivity, where client initiates
> session to target socket (socket is combination of IP address and
> port). hosted-engine-ha-agent component on source ha-host will create
> TCP session to where?

To the destination IP address
(OVEHOSTED_NETWORK/network_test_tcp_address)
and the desired destination TCP port
(OVEHOSTED_NETWORK/network_test_tcp_port)
of the TCP connection test, if the user wants to use the TCP check.

> Is it possible to interactively ask customer, instead of stopping deployment, in case of wrong data provided during deployment from answer file?

This would be not helpful in automated usage and inconsistent the the
other inputs.

> It'll be better to continue, instead of end up with dirty environment.

This will not end up with a dirty environment, because the inputs are
checked before the hosted engine is deployed.

> I'm not sure if usage of hostnames is a good idea,

The user does not have to use textual hostnames, the user can use plain IPv4 or
IPv6 addresses, too.

> hostnames are names of the servers or VMs within the same LAN, but
> they're local names and won't be reachable from external networks,

The hostname is validated on the first host, and has to be reachable by
the other hosts which might host engine VM.
The hostname doesn't has to be reachable by external networks, only by
the other hosts which might host the engine VM.

> hence I think that DNS option should be in FQDN format, hostname+domain
> name.
>


I am against creating artificial limitations, if there are possible
useful usages.

dominikholler:
@nsednev Are your questions answered in a sufficient way?

source/develop/release-management/features/sla/hosted-engine-network-check.html.md
dominikholler: Done

dominikholler:
@mmirecki @sandrobonazzola From my point of view this PR is ready to be merged.

sandrobonazzola:
Thanks!

Issue #2005, opened 2019-06-02
Issue: /source/documentation/admin-guide/appe-oVirt_and_SSL.html.md


mkowalski:
#### Type of problem?
Documentation issue

Point (10) in the SSL certificate replacement procedure references to the `/etc/ovirt-imageio-proxy/ovirt-imageio-proxy.conf` file, however sample code just a line below references `/etc/ovirt-engine/ovirt-websocket-proxy.conf.d/10-setup.conf`. This looks like a simple copy-paste bug.

PR #2006, opened 2019-06-02
Fix incorrect example for ovirt-imageio-proxy.conf

mkowalski:
This patch fixes wrong filepath in the example code
from 10-setup.conf to ovirt-imageio-proxy.conf

Closes-Bug: #2005

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @mkowalski

### DIFF
--- a/source/documentation/admin-guide/appe-oVirt_and_SSL.html.md
+++ b/source/documentation/admin-guide/appe-oVirt_and_SSL.html.md
@@ -94,7 +94,7 @@ The internal CA stores the internally generated key and certificate in a **P12**
 
 10. Edit the **/etc/ovirt-imageio-proxy/ovirt-imageio-proxy.conf** file:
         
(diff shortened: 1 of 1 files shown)


didib:
I don't mind merging, because it obviously fixes an error, but this page still requires some love:

1. Compare to RHV docs and see what we forgot to update here. IIUC everything there is relevant to oVirt.

2. Consider also currently-open bugs, and either update the docs when fixed, or update them now as needed and then update again when fixed. E.g.:

https://bugzilla.redhat.com/show_bug.cgi?id=1385617

3. Also, we hope to get rid of this page almost entirely, replacing it with an ansible playbook:

https://bugzilla.redhat.com/show_bug.cgi?id=1336873

That's definitely on my todo list, but volunteers are always welcome :-)

PR #2113, opened 2019-10-06
Create coreos-ignition-support-design-discussion

stvnrsnbrg:
This file is intended to be used in order to discuss design issues for enabling Ignition support within oVirt. This may include initial support as well as discussion for future implementation of features supported and not supported by Cloud-Init.

- Design discussions for the coreos-ignition-support integration with oVirt.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): (please @stvnrsnbrg  yourself to sign)

This pull request needs review by: (please @rgolangh  the reviewer if relevant)

### DIFF
--- a/source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
+++ b/source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
@@ -1,0 +1,69 @@
+---
+title: CoreOS Ignition Support Design Review
+category: feature
+authors: srosenbe
+feature_name: CoreOS Ignition Support Design Review
+feature_modules: engine
+feature_status: Tech Preview 4.3
+---
+
+## Overview
+Design Proposals based upon CoreOS Ignition Support Documentation: 
+
+https://github.com/oVirt/ovirt-site/edit/master/source/develop/release-management/features/virt/coreos-ignition-support.md
+
+The purpose of this document is to define which features shall be supported by CoreOS Ignition for RHCOS and FCOS within oVirt as per the current Cloud-Init implementation. As per the document referenced, certain features may not be included at least within the initial release and therefore an agreement on what may be added initially and what can be implemented in later releases is required as per the "Limitations" and "Future work" sections therein.
+
+
+## Initial Run for Cloud-Init 
+
+Cloud-Init supports the following top level functions:
+
+- VM Hostname
+- Configure Time Zone
+- Authentication
+- Networks
+- Custom Script
+
+From the coreos-ignition-support document:
+> Network configuration which are currently cloud-init specific protocol will not work with ignition. Ignition simply
+> manipulate the disk therefor the desired network state will be achieved through laying configuration files.
+
+Therefore, we may need to consider the impact of removing the Network configuration for Ignition or discuss other options.
+
+
+Authentication Section:
+
+- User Name
+- Use already configured password
+- Password
+- Verified Password
+- SSH Authorized Keys
+- Regenerate SSH Keys
+
+From the coreos-ignition-support document:
+> authentication/ssh-authorized-keys: that would be simply a copy to an ignition config snippet
+
+Current support for SSH Authorized Keys may be postponed for future development due to Ignition limitations or to discuss other options.
+
+Network Section:
+
+- Network Protocol
+- DNS Servers
+- DNS Search Domains
+- In-guest Network Interface Name
+- List Boot Protocols:
+  - IPv4 Boot Protocol: None/DHCP/Static
+  - IPv6 Boot Portocol: None/DHCP/Static
+  
+Defining the Network interface via "layering configuration files" can be defined for future development.
+
+Custom Script Section: 
+
+- Allows the user to enter a script that shall be run on the Virtual Machine when it starts. Format is in Custom YAML sections which may address the SSH Authentication and Networking requirements. Future implementation may allow for automatically sending custom scripts for various configurations. EXamples can be found here:
+
+https://coreos.com/ignition/docs/latest/examples.html
+
+Note: Just as SysPrep sends the configuration within files by creating the Unattend.xml file, we may consider a similiar solution for Ignition for future consideration.
+
+


source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
```diff
@@ -0,0 +1,69 @@
...
+feature_modules: engine
+feature_status: Tech Preview 4.3
+---
+
+## Overview
+Design Proposals based upon CoreOS Ignition Support Documentation: 
+
+https://github.com/oVirt/ovirt-site/edit/master/source/develop/release-management/features/virt/coreos-ignition-support.md
+
+The purpose of this document is to define which features shall be supported by CoreOS Ignition for RHCOS and FCOS within oVirt as per the current Cloud-Init implementation. As per the document referenced, certain features may not be included at least within the initial release and therefore an agreement on what may be added initially and what can be implemented in later releases is required as per the "Limitations" and "Future work" sections therein.
```
rgolangh: It should be noted that the 'support' scope here is UI and API. It needs to be clearly understood that we support 
passing the ignition config v2 as is already,  and the added features here meant to generate the ignition config themselves, rather than just passing it to custom_script.

source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
```diff
@@ -0,0 +1,69 @@
...
+- Use already configured password
+- Password
+- Verified Password
+- SSH Authorized Keys
+- Regenerate SSH Keys
+
+From the coreos-ignition-support document:
+> authentication/ssh-authorized-keys: that would be simply a copy to an ignition config snippet
+
+Current support for SSH Authorized Keys may be postponed for future development due to Ignition limitations or to discuss other options.
```
rgolangh: what limitations?

source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
```diff
@@ -0,0 +1,69 @@
...
+- In-guest Network Interface Name
+- List Boot Protocols:
+  - IPv4 Boot Protocol: None/DHCP/Static
+  - IPv6 Boot Portocol: None/DHCP/Static
+  
+Defining the Network interface via "layering configuration files" can be defined for future development.
+
+Custom Script Section: 
+
+- Allows the user to enter a script that shall be run on the Virtual Machine when it starts. Format is in Custom YAML sections which may address the SSH Authentication and Networking requirements. Future implementation may allow for automatically sending custom scripts for various configurations. EXamples can be found here:
```
rgolangh: EXamples/Examples

source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
```diff
@@ -0,0 +1,69 @@
...
+
+https://coreos.com/ignition/docs/latest/examples.html
+
+Note: Just as SysPrep sends the configuration within files by creating the Unattend.xml file, we may consider a similiar solution for Ignition for future consideration.
```
rgolangh: What do you mean?

source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
stvnrsnbrg: The issues we discussed with SSH Authorization Keys support.

source/develop/release-management/features/virt/coreos-ignition-support-design-discussion
stvnrsnbrg: Generating the Ignition configuration file and updating it automatically so that we can overcome the limitations discussed.

evol262:
What we really need for complete support is an additional OS (for RHCOS), and extensions of the RunOnce fields for that ostype to populate ignition configs with username/password/ssh key similar to cloud-init

liranr23:
Network settings are not similar to cloud-init implementation.
The network for ignition works with systemd.
For now, I think we should wait with the network settings (the user will be able to set it in the custom script section).

rgolangh:
We need to figure out how are we going to translate this widget into a valid ignition config:
image

Essentially this should be it:
```
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDGdByTgSVHq......."
    - name: elroy
      password_hash: "$6$5s2u6/jR$un0AvWnqilcgaNB3Mkxd5yYv6mTlWfOoCYHZmfi3LDKVltj.E8XNKEcwWm..."
      ssh_authorized_keys:
        - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDGdByTgSVHq......."
```


Liran / Steven please check that and see this is feasible and we don't have something which we can't implement.

For example setting timezone would have to be implemented by setting a file. This is harder maybe to get right.

evol262:
Templating this in the same way as sysprep may be the best method for now, since, AFAIK, the YAML format can't actually be used without config transpiling

rgolangh:
What do you mean by templating?

evol262:
I mean, since CoreOS's config is more or less a JSON derivative, stubbing it out in exactly the same way as the XML for sysprep instead of trying to compile a new config every time throught he CoreOS Config Transpiler or building custome JSON may be safer considering the versioning issues with Ignition

rgolangh:
I can't see how it is safer then generating JSON using code. In what way it
is safer?
Also for supporting the ssh keys for example. this would mean we would have
a template just for the ssh snippet
that we would stitch the json. I prefer structuring a json using an API. I
can't make mistakes there. at least no
structural mistakes, and the json will be in complete control, no residual,
empty/dummy parts from a template.

And back to the main discussion, the templating is irrelevant in
implementing the network configuration widget.
What do you think about the network, should we do it or not?

evol262:
For network, I would say no, since CoreOS has its own mechanisms for this, and users who want a complex networking setup can write custom Ignition.

For the scripts themselves, I'm primarily in favor of templating vs generating JSON since we have seen problems in testing and other places due to differing Ignition config versions. I'd rather stick to one than have an ongoing maintenance point where we may be asked to support multiple, different versions of Ignition's format, but I don't have a strong opinion either way

rgolangh:
On Tue, 19 Nov 2019 at 17:43, Ryan Barry <notifications@github.com> wrote:

> For network, I would say no, since CoreOS has its own mechanisms for this,
> and users who want a complex networking setup can write custom Ignition.
>

+1

Also, our main use case, and the general use case of ignition is for
OCP/OKD, and there the ignition config is generated by the
Machine-Config-Opetator anyway, so
I don't see users really needing abstraction for creating network
configuration for ignition.


For the scripts themselves, I'm primarily in favor of templating vs

liranr23:
An update on this discussion, I went with JSON implementation.
In the future if we will add another UI options, creating a relevant JSON should be enough.
We left with:
Authentication, including user name, password and ssh keys, and custom script.
It added to UI in the admin portal and runonce.
**The ignition version must be added!** We need to think whether we wish to add it by the OS type or let the user set it.
The related patch: https://gerrit.ovirt.org/#/c/105098/

sandrobonazzola:
any update?

sandrobonazzola:
any update @stvnrsnbrg?

sandrobonazzola:
Closing due to unresponsiveness

Issue #2086, opened 2019-09-02
Enable site support for AsciiDoc format


pmkovar:
A larger part of the ovirt user docs ( https://www.ovirt.org/documentation/) is now outdated, still describing 4.2, and out of sync with the downstream RHV product docs (https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/) that are actively maintained. 

With the site migration to Jekyll (#1968), we can now enable support for the AsciiDoc format (the Asciidoctor variant) that is a prerequisite to keep ovirt docs updated and in sync with downstream.

pmkovar:
@dominikholler @sandrobonazzola

duck-rh:
I can help for the Jekyll settings but I'll need some example asciidoc sources to integrate; I could not find the sources to the linked doc.

pmkovar:
Many thanks @duck-rh 

@rolfedh Can you help here with supplying a good stand-alone RHV adoc file as a test doc? Doesn't matter what the content is, we are interested in markup only.

sandrobonazzola:
+1 for enabling asciidoc

rolfedh:
> ith supplying a good stand-alone RHV adoc file as a test doc? 
Here's the upgrade guide...
[doc-Upgrade_Guide.zip](https://github.com/oVirt/ovirt-site/files/3569607/doc-Upgrade_Guide.zip)

rolfedh:
I have some questions about the proposed approach. The oVirt doc repo is open source. How do we incorporate contributions that others have made directly to that repo? Merging may not be enough. If RHV delivers a subset of oVirt's features, how do we address the features that are not covered in the RHV doc?

pmkovar:
@rolfedh Thanks, I think the archive is missing files from the `common` directory, though.

I agree simple merging may not be enough. My idea is to carefully merge one title at a time, making sure we do not override upstream-only changes that might have been made. Every PR should go through a thorough SME/peer review. 

That said, looking at the git history, the number of upstream-only contributions seems low for most titles under https://github.com/oVirt/ovirt-site/tree/master/source/documentation. 

I'd leave it up for the community contributors to decide what upstream-only features they want to cover in user docs, and how.

Thoughts on improving the workflow welcome, though.

duck-rh:
[...]

PR #2129, opened 2019-11-01
Add asciidoc support and import intro-vm-portal

pmkovar:
Fixes issue #2086

* Add support for font-based admonitions & coderay highlighter

  Based on 0a135ac2f93505274d133aa797371b1e9fae0c4e and a successful jekyll-asciidoc deployment at:
https://github.com/eclipse/che-docs/

* Import asciidoc version of intro-vm-portal guide for ovirt 4.3 as a proof of concept

  Based on the common attributes setup by @rolfedh.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @pmkovar

This pull request needs review by: @sandrobonazzola @duck-rh


pmkovar:
Can confirm that the `listen` gem keeps complaining about symlinks (https://github.com/guard/listen/wiki/Duplicate-directory-errors) at build time. Can we deploy anyway?

Tried to workaround that problem with 484d0cdef99605378f7cf5330bf8b27428f8442d, but interestingly enough that triggers another problem, probably related to the fact `asciidoctor` won't allow includes from the index file's parent directory for security reasons even in unsafe mode:

```
ERROR Errno::ECONNRESET: Connection reset by peer @ io_fillbuf - fd:16 
	/usr/share/ruby/webrick/httpserver.rb:82:in `eof?'
	/usr/share/ruby/webrick/httpserver.rb:82:in `run'
	/usr/share/ruby/webrick/server.rb:307:in `block in start_thread'
```
We could try placing a master index file under ` source/documentation/` and include all guides from that file, but we don't want to deviate too much from the downstream setup because of content sharing. 

Other ideas?

sandrobonazzola:
@duck-rh can you help answering @pmkovar questions?

duck-rh:
@sandrobonazzola when I come back from event and vacation on 18th I will, sorry for the delay

sandrobonazzola:
Hey @duck-rh any update?

_config.yml
```diff
@@ -44,6 +44,23 @@ kramdown:
   parse_block_html: true
   parse_span_html: true
 
+plugins:
+  - jekyll-asciidoc
+
+asciidoc: {}
```
duck-rh: This is not needed / legacy, the `asciidoctor` entry is sufficient.

pmkovar:
> If you move `source/documentation/common` out of the source directory (like a `adoc-common` top directory or the like and update the symlink accordingly), then it fixes your problem. Also Jekyll will generate any subpage in common, so it's better to move it aside to avoid broken generation and having unwanted pages.

@duck-rh  Done, thanks for the suggestion.
 
> In fact I would suggest finding a way to move all includes away and keep only the index.adoc unless we want split pages.

Understood, but I want to keep the changes to the asciidoc sources to a bare minimum.

sandrobonazzola:
@duck-rh can you please review again?

duck-rh:
@sandrobonazzola  Same remarks as #2177 apply.

Both PRs embed the asciidoc activation and coderay addition, so one should go in first and the other be rebase on top of it.

So once it is done in either of them I can do a final review.

pmkovar:
> @sandrobonazzola Same remarks as #2177 apply.
> 
> Both PRs embed the asciidoc activation and coderay addition, so one should go in first and the other be rebase on top of it.

@sandrobonazzola @duck-rh Let's merge this one first. #2177 is a follow up.

PR #2142, opened 2019-11-13
Added debugging steps for vdsm

ezio-auditore:
I have added steps to debug vdsm using pdb(python debugger):

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): (please @ezio-auditore  yourself to sign)

This pull request needs review by: @nirs @sandrobonazzola

### DIFF
--- a/source/develop/developer-guide/vdsm/developers.md
+++ b/source/develop/developer-guide/vdsm/developers.md
@@ -594,6 +594,23 @@ You have no tags in your repo. If you want to push your repo from another machin
 
 We have a partial code walk through of the [virt subsystem](/develop/developer-guide/vdsm/vm-startup/) (virtual machine life cycle, creation, migration) (more to come...)
 
+## Debugging 
+
+You can debug VDSM using the standard Python debugger [pdb] (https://docs.python.org/3/library/pdb.html)
+Add `import pdb; pdb.set_trace()` at the point where you want a breakpoint. Make sure you have both vdsmd and supervdsmd
+services stopped. Now you have to run vdsm and supervdsm interactively. 
+First run supervdsm as in one console
+
+       /usr/bin/python3 /usr/share/vdsm/supervdsmd --sockfile /var/run/vdsm/svdsm.sock
+       
+Make sure you need to run vdsm as user vdsm and `vdsm:x:36:36::/home/vdsm:/bin/bash` is set in your `/etc/passwd` and `/etc/passwd-`. Login as root and issue the following command in another console
+       
+       $ sudo su vdsm - 
+
+       /usr/bin/python3 /usr/share/vdsm/vdsmd
+
+Now at some point when vdsm stops at breakpoint you can use the pdb commands to navigate and debug.      
+
 ## Performance and scalability
 
 To analyse the performance and the scalability of the VDSM, you first need to [set up the tools](/develop/developer-guide/vdsm/profiling-vdsm/). Then you may want to [run common scenarios](/develop/developer-guide/vdsm/benchmarks/), or write your own using a [template](/develop/developer-guide/vdsm/benchmarks/), to make sure the results are easily shareable.
--- a/source/develop/developer-guide/vdsm/developers.md
+++ b/source/develop/developer-guide/vdsm/developers.md
@@ -601,13 +601,13 @@ Add `import pdb; pdb.set_trace()` at the point where you want a breakpoint. Make
 services stopped. Now you have to run vdsm and supervdsm interactively. 
 First run supervdsm as in one console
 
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
Thanks for submitting this PR! @nirs @mwperina can you please review?

mwperina:
Marcin, could you please review?

@tinez

source/develop/developer-guide/vdsm/developers.md
```diff
@@ -601,13 +601,13 @@ Add `import pdb; pdb.set_trace()` at the point where you want a breakpoint. Make
 services stopped. Now you have to run vdsm and supervdsm interactively. 
 First run supervdsm as in one console
 
-       /usr/bin/python3 /usr/share/vdsm/supervdsmd --sockfile /var/run/vdsm/svdsm.sock
+        $ /usr/bin/python3 /usr/share/vdsm/supervdsmd --sockfile /var/run/vdsm/svdsm.sock
```
tinez: This would suggest that you should run 'supervdsm' as a regular user, while it needs to be run with root privileges. I think it'd be better to change '$' to '#'

source/develop/developer-guide/vdsm/developers.md
```diff
@@ -594,6 +594,23 @@ You have no tags in your repo. If you want to push your repo from another machin
...
+## Debugging 
+
+You can debug VDSM using the standard Python debugger [pdb] (https://docs.python.org/3/library/pdb.html)
+Add `import pdb; pdb.set_trace()` at the point where you want a breakpoint. Make sure you have both vdsmd and supervdsmd
+services stopped. Now you have to run vdsm and supervdsm interactively. 
+First run supervdsm as in one console
+
+       /usr/bin/python3 /usr/share/vdsm/supervdsmd --sockfile /var/run/vdsm/svdsm.sock
+       
+Make sure you need to run vdsm as user vdsm and `vdsm:x:36:36::/home/vdsm:/bin/bash` is set in your `/etc/passwd` and `/etc/passwd-`. Login as root and issue the following command in another console
```
tinez: > Make sure you need to run vdsm as user vdsm

This doesn't make a nice sentence

> `vdsm:x:36:36::/home/vdsm:/bin/bash` is set in your `/etc/passwd` and `/etc/passwd-`

Providing the whole line from '/etc/passwd' is not a good idea. It should be sufficient to mention that you need to run 'vdsm' as 'vdsm' user, but by default 'vdsm' user has its shell set to '/sbin/login'. To make logging in possible you need to change it with `usermod -s /bin/bash vdsm`

source/develop/developer-guide/vdsm/developers.md
```diff
@@ -594,6 +594,23 @@ You have no tags in your repo. If you want to push your repo from another machin
...
+       
+       $ sudo su vdsm - 
+
+       /usr/bin/python3 /usr/share/vdsm/vdsmd
+
```
tinez: > Now at some point when vdsm stops at breakpoint you can use the pdb commands to navigate and debug. 

Here it's worth mentioning, that debugging 'vdsm' is not that simple. Since it's multithreaded you should expect some logging messages occuring in your console even though you hit a breakpoint. It's also possible that the thread you've reached with a breakpoint may be killed due to timeout handling code.

nirs:
I never debug vdsm using pdb. I don't find it useful for several reasons:
- pdb requires changing the source to create breakpoints
- you can debug python programs using gdb (requires debuginfo-install pythonX). This gives give access both to python code (using py-bt) and the underlying C code (using bt)
- debugging vdsm is not very useful since it is very helpful since it is multi-threaded and stopping all thread will create a big mess. The only reason to stop all threads is to understand why it is stuck, and we have better tools to understand this like https://github.com/oVirt/vdsm/commit/ce265f7d6de80a725f1fd0aed0b6dbb826d18c75
- If you cannot understand what vdsm is doing by watching the logs you have bad logging and you will not be able to solve issues in the field. Better to invest in improving the logs, so you don't need a debugger
- if you cannot understand the code without a debugger the code is too complex and you need to simplify it
- if you need a debugger you don't good enough tests, and you should improve them instead
- Sometime you need access to vdsm internal state while it is running - we have integrated manhole support, see: https://github.com/oVirt/vdsm/commit/c67dfa80081f5fdfb1016c593dac65a16e867861

Having said that, if you find pdb useful, adding a document about debugging vdsm with pdb in vdsm source can be a good idea (e.g. doc/debugging-with-pdb.md). I don't think that ovirt site is a good place for developer documentation for single ovirt project. It is a good place for user documentation or documenting topics related to many ovirt projects (e.g. engine, vdsm, imageio).

sandrobonazzola:
closing based on @nirs comment

PR #2240, opened 2020-05-28
automation: add internal links checking

sandrobonazzola:

- Once we finish to fix the 4418 broken internal links we have
  right now, this automation should help us keeping the website
  in a better shape.


Gemfile.lock
```diff
@@ -154,7 +173,10 @@ DEPENDENCIES
   kramdown
   liquid
   listen
+  rake
 
 RUBY VERSION
    ruby 2.5.5p157
 
+BUNDLED WITH
```
duck-rh, line 181: This may break the website generation on the builder. I don't recall the available version but most probably inferior compared the one on your system. Also it is not needed (but added automatically by bundler comands), so please remove it.

Gemfile
```diff
@@ -42,3 +42,6 @@ gem 'chronic'
 # for dot access to Hash members in HAML pages
 gem 'hash_dot', '~> 2.4.2'
 
+# for checking broken links
+gem 'rake'
```
duck-rh, line 46: We do not run the tests using Rakefile, so it's unnecessary.

duck-rh:
Another note: on my system, nokogumbo (dependency of html-proofer) required nokogiri 1.10.9 so I had to update the Gemfile.lock accordingly. it may be because I'm using Ruby 2.7 on my system but I think it would be harmless to bump a minor version to get the compatibility.

Gemfile.lock
sandrobonazzola: ok

Gemfile
sandrobonazzola: ok

sandrobonazzola:
About the .htaccess rewrite rule, I searched a bit but didn't find anything for making the analysis be aware of it.

duck-rh:
> About the .htaccess rewrite rule, I searched a bit but didn't find anything for making the analysis be aware of it.

Clearly the tool focus on static analysis, which makes sense because these rules are HTTP server specific. The other way to check is to make HTTP requests like webcheck did, but it takes way longer and this tools has more tests.

I also found there are 4 (1 is obsolete) redirections in the httpd vhost config and I guess having all the rules in one place and in control of the team would be better, so I propose to move them in the .htaccess file.

I also think we need to get all the things fixed (or ignored if too pedantic) before merging or PRs won't pass the tests and then would not be able to be merged anymore,

sandrobonazzola:
I totally agree, this PR must not be merged before it passes CI

sandrobonazzola:
Abandoning this one after #2294

PR #2270, opened 2020-06-30
net: Add limitation documentation to copy-host

bellle:
@dominikholler

### DIFF
--- a/source/develop/release-management/features/network/copy-host-networking.html.md
+++ b/source/develop/release-management/features/network/copy-host-networking.html.md
@@ -102,6 +102,11 @@ the feature cannot determine what to do and the operation will fail.
 
 * The cluster's management network stays on the original interface which is excluded from the operation. 
 
+* Copying QoS, DNS and custom_properties is not supported.
+
+* Copying labels (i.e., label attached to network interface on source host, will not be copied to network
+interface on target host).
+
 
 ### Implementation
 


source/develop/release-management/features/network/copy-host-networking.html.md
```diff
@@ -102,6 +102,10 @@ the feature cannot determine what to do and the operation will fail.
 
 * The cluster's management network stays on the original interface which is excluded from the operation. 
 
+* QoS, DNS and custom_properties are currently unsupported.
```
MosheSheena, line 105: 'currently unsupported' suggest there is a plan somehow somewhere to make it supported - if there are already bz's to support this please add them so that whoever is interested will have a way to track their progress, or if there aren't I think it's perfectly fine to write they are simply unsupported - if in the future that will change then obviously this document will change as well.

source/develop/release-management/features/network/copy-host-networking.html.md
```diff
@@ -102,6 +102,10 @@ the feature cannot determine what to do and the operation will fail.
...
+
+* Multi-host copying with labels.
```
MosheSheena, line 107: So this is listed as a limitation, yet I failed to realize what's the meaning of this sentence.
Could you maybe elaborate on where the limitation stands regarding labels?

bellle: @MosheSheena I think it means copying the networks's label. So I guess the network will be copied but the label will not.
Do you think something like "Multi-host network labels copying" will be better?

MosheSheena: @almusil could you please help us out on the label's part? to my knowledge labels are not supported, and this was not covered by QE.

Bell I believe these limitations should explicitly and clearly state what is not supported and in what way meaning if copying labels isn't supported then it should be written as plain as possible so that users could quickly be informed of it.

almusil: AFAIR labels were never intended to be copied because they are kinda separate unit. So it is fine to mention that copying of labels is not supported but I don't quite understand the multi-host part.

MosheSheena: @bellle I believe it should be strictly written that the following will not be copied from the source host to the target host: QoS, DNS, custom_properties (without hints to future support).
Labels attached to network interfaces on the source host will not be attached to network interfaces on the target host.

source/develop/release-management/features/network/copy-host-networking.html.md
bellle: Since this is in limitation, I just listen the features that are unsupported. I hope it is ok, wanted it to be black and white.

source/develop/release-management/features/network/copy-host-networking.html.md
bellle: Ok I made it verbose to be able to understand.

source/develop/release-management/features/network/copy-host-networking.html.md
```diff
@@ -102,9 +102,10 @@ the feature cannot determine what to do and the operation will fail.
 
 * The cluster's management network stays on the original interface which is excluded from the operation. 
 
-* QoS, DNS and custom_properties are currently unsupported.
+* Copying QoS, DNS and custom_properties.
```
MosheSheena, line 105: @bellle could you please add "are not supported" to the end of the line?
Just to make it super clear.

bellle: Done

dominikholler:
@sandrobonazzola can you please merge?

PR #2298, opened 2020-07-22
Add TPM feature page

mz-pdm:

- Initial version of TPM feature page. It's WIP, will be updated as needed.

This pull request needs review by: @ahadas

### DIFF
--- a/source/develop/release-management/features/virt/tpm-device.md
+++ b/source/develop/release-management/features/virt/tpm-device.md
@@ -1,0 +1,161 @@
+---
+title: TPM devices
+category: feature
+authors: mzamazal
+feature_name: TPM devices
+feature_modules: engine,vdsm
+feature_status: In development
+---
+
+# TPM devices
+
+## Summary
+
+This feature allows adding emulated TPM (Trusted Platform Module) devices to VMs.
+
+## Owner
+
+*   Name: Milan Zamazal
+*   Email: mzamazal@redhat.com
+
+## Description
+
+TPM devices are useful to perform certain cryptographic operations (generating cryptographic keys, random numbers, hashes, etc.) or to store data that can be used to verify current software configuration securely.  QEMU and libvirt implement, among other, support for emulated TPM 2.0 devices, which is what oVirt uses to add TPM devices to VMs.
+
+Once an emulated TPM device is added to the VM, it can be used as a normal TPM 2.0 device in the guest OS.
+
+## Prerequisites
+
+There are no special prerequisites for this feature.  Software TPM emulation in QEMU is implemented with the help of [swtpm](https://github.com/stefanberger/swtpm) package, which is installed on the host automatically together with Vdsm.
+
+## Security
+
+TPM device data is managed by swtpm and stored in the host file system (until the VM stops running there) and in the Engine database (TODO: update according to the selected implementation).  Whoever has access to those locations has access to the TPM data.  This is necessary to provide the data to the currently running VM or on its next start.  libvirt provides an option to encrypt the data with a key, but the key still must be stored somewhere in Engine and in the domain XML, so it's more obfuscation rather than real security and it's currently not used by oVirt.
+
+Availability of data stored in TPM shouldn't be critical.  Its loss can cause operational problems such as a need to initialize and set up TPM data again, to audit the operating system after the TPM data is reset (lost) or perhaps even to reinstall the whole system but it shouldn't cause real data loss.  When using bare metal, one can't rely on TPM data availability either because if the hardware stops working, the data is no longer accessible.  If one uses for example a private key stored only in TPM to encrypt and decrypt file system encryption keys, then the risk of losing data stored in the file system without having proper backups is at least as big with hardware TPM as with virtual TPM.
+
+## Limitations
+
+TPM devices can be used only with UEFI BIOS.
+
+Until <https://bugzilla.redhat.com/1855367> is implemented in libvirt, VMs with TPM devices cannot have snapshots with memory.  The problem is that we need swtpm data snapshot consistent with the VM snapshot, which is not possible to get without libvirt assistance.
+
+Since the emulated TPM is a software rather than a hardware device, its data updates can get lost, e.g. when a host suddenly crashes.  (TODO: Clarify according to the selected implementation.)
+
+There is apparently no guarantee that new swtpm versions can still read its old data.  That should be watched in testing.
+
+## User Experience
+
+Ideally, a TPM device would be added to each VM automatically and would be automatically available to the guest OS.  However, TPM support is a new feature and we cannot guarantee that everything is correctly supported in the underlying platform (for example, snapshots with memory are currently not working as expected).  For that reason, it's better to not add TPM devices to VMs by default for now and to make a VM option to enable or disable them per VM as needed, defaulting to the VM's template setting.
+
+TODO: Description of the UI actions once they are implemented.
+
+## Testing
+
+There are two areas to test: TPM presence and data persistence.
+
+TPM devices should be present in VMs running in clusters of versions >= 4.4 (TODO: re-check the version) when the corresponding VM option is checked.  A TPM device is present as ``<tpm>`` device in the domain XML and as a corresponding device in the guest OS (``/dev/tpm0`` on Linux systems).
+
+Newly created VMs, or VMs that didn't have a TPM device previously, get uninitialized TPM data, there is nothing to test about it.  Once TPM data is modified by the guest OS, the modified data must be present after different operations: VM stop & start, reboot, migration, export/import, cloning, start from snapshot, etc.  TPM data updates followed by common failure scenarios (disconnected host, crashed VM, etc.) can be tested in situations where it's reasonable to assume the data updates should be available.  The data can be checked in the guest OS (most reliably) or alternatively comparing swtpm data with expected contents.
+
+It should be also tested that VMs with TPM data created on hosts running older systems are unchanged when the VM is started on a host running the newest system.  This must be tested in the guest OS.
+
+## Changes to Engine and Vdsm
+
+There are no special changes needed in order to add a TPM device as such, it's just another kind of device added to Engine.
+
+### Data persistence
+
+A very different thing is ensuring TPM data persistence.  TPM data is managed using an external tool, swtpm.  Once the VM domain XML is deleted on the host, the TPM data is deleted as well by libvirt and it must be restored before the VM is run again, whether on this or another host (live host migrations are handled transparently and don't require any special actions on the oVirt side).
+
+There are several options how to provide TPM data on the VM start:
+
+- As a parameter to a VM.create API call.
+- In metadata section of the domain XML.
+- On a shared storage.
+
+The hard part is how to store updated TPM data from the host reliably, making sure that:
+
+- Data updates are not lost.
+- The updated data is consistent, i.e. it can be read by swtpm correctly on the next VM start.
+
+We cannot handle all erroneous situations (e.g. faulty hardware corrupting data) but a reasonable level of resistance against common failures should be granted.  Ideally, the following situations should be handled:
+
+- When a VM stops for any reason on a connected host, we store the latest state of the data.
+- When a host or a storage get disconnected, we store the latest known state of the data.
+- When a VM is restarted without Engine assistance, it uses the final data from the last run.
+- We don't store inconsistent data.  Currently used swtpm version [doesn't store changes atomically](https://github.com/stefanberger/swtpm/blob/609dfd873a31fc0402752c50d1d5cfa58949507f/src/swtpm/swtpm_nvfile.c#L492), so whether we copy swtpm data or let swtpm write to shared storage, we are at risk of race with swtpm writes on copying or when the storage gets suddenly disconnected.  This has been fixed upstream so once the update reaches Advanced Virt, there should be no longer problem with TPM data consistency within a single file.
+- Temporary network disconnections, lost events sent from Vdsm to Engine or temporary failures when delivering or processing Vdsm responses to Engine don't cause permanent loss of data updates.
+
+The options to deliver updated TPM data are similar to obtaining TPM data on VM start, with the important distinction that this time they are write instead of read operations, with the risk of data loss.  We can:
+
+- Send the updated data to Engine in events, current API call responses (stats, destroy) or using a new Vdsm API call for the given purpose.
+- Store the data in metadata, to be read and processed in Engine monitoring.
+- Attach the data to dumpxmls
+- Store the data to shared storage from Vdsm.
+- Let swtpm store the data directly to the shared storage.
+- Some combination of the options above.
+
+Different options are discussed in the following subsections.  We can probably assume that TPM data updates are not going to be very frequent.  We mustn't forget about reboots, snapshots, imports and exports.
+
+#### Using API calls
+
+TPM data can be stored in the Engine database and transferred in both the directions using Vdsm API calls.  It can be provided to the VM as an additional parameter to VM.create call.  Data changes can be reported back to Engine using:
+
+- Events sent from Vdsm.
+- As a return value parameter of VM.destroy call.
+- As an additional value in VM stats.
+- On a request from Engine using a newly introduced Vdsm API call.
+
+TPM data can be copied to swtpm data location before the VM is created.  Once VM is started, updated TPM data can be read from the swtpm data location.  A problem is how to read data from swtpm consistently, avoiding collisions with swtpm writes.  Considering TPM data is small, written locally and not updated frequently, a good enough approach should be to have a periodic Vdsm job watching for swtpm data changes.  If it detects a data change, it keeps watching the data and once it detects the same data in two consecutive job runs, it can report it as changed data relatively safely (unless there is some fatal condition such as a full local disk).  This mechanism would be common to all the approaches except for shared storage directly used by swtpm.
+
+There are some problems specific to this approach though:
+
+- Events are not reliable and Vdsm can't be sure Engine has received and processed the updated data.
+- The same applies to reporting the data in VM stats, unless we are willing to send the data, even when unmodified, on each VM stats query.
+- Data read after the VM stops can be considered stable and reported directly.  However, if it is attached to VM.destroy response and the response fails for any reason, it can't be reported again in a response to a repeated VM.destroy call, because the data is already deleted after the libvirt domain has been undefined.
+- A specific API call would solve most of the problems.  Vdsm could include TPM data into the computation of the VM device hash sent within VM stats and Engine could request TPM data itself if the device hash changes.  Alternatively, Engine could make the API call periodically, providing TPM data hashes as an argument, and Vdsm would respond only with data not matching the hashes.  An additional call would happen in either case just before calling VM.destroy to get the ultimate final version of TPM data.
+- It is fully dependent on Engine cooperation, it's not possible to start a VM without Engine explicitly providing the data.  But if we needed just to restart a VM on the same host without Engine then we could retain TPM data by either not undefining the domain XML or by making a local copy of the data between the restarts.
+
+This approach should be feasible with introducing a new Vdsm API call as outlined above but it has the drawback that Engine assistance is needed.
+
+#### Storing the data in domain XML metadata
+
+Using domain XML metadata, it's possible to transfer TPM data in all the directions.  TPM data can be read from the data on VM start, with or without Engine assistance, updates can be stored there atomically and Engine can update the data in monitoring.
+
+Ensuring data consistency read from swtpm is the same as with API calls.
+
+The only problem with using metadata is that it increases domain XML size significantly.  TPM data shouldn't be large although to know the size limits would require further investigation.  But initial swtpm data for a VM takes about 4 KB encoded.  It would "pollute" the domain XML and would have to be transferred each time Engine asks for the domain XML, even when there are no changes to data, which means it would be waste most of the time.
+
+This approach is a bit simpler and more flexible than using API calls but has the data size drawback.
+
+#### Attach the data to dumpxmls
+
+Since TPM data is not supposed to change frequently, it can be attached to Host.dumpxmls response.  The computation of the device hash can take TPM data into account.  When Engine calls dumpxml after the hash change, it will get not only the domain XML but also the TPM data, as another item in the dictionary.  (The same approach may possibly be used also for reporting the secure boot data in future.)
+
+Ensuring data consistency read from swtpm is the same as with API calls.
+
+This mechanism is similar to metadata but it puts the TPM data outside of the domain XML.  It shares the property of transferring the data each time Engine asks for the domain XML (in theory, it could be optimized by not sending unchanged data if a sufficiently robust mechanism was found).
+
+#### Using shared storage
+
+A somewhat similar approach to using domain XML metadata would be to store the data to a shared storage.  Instead of writing to or reading it from the metadata, it would be written to or read from the storage.  The benefit would be avoiding domain XML pollution and unnecessary data transfers.  Additionally, Engine wouldn't have to process the data and maintain it in the database and the data could be updated even when Engine was temporarily unavailable.
+
+However, there are also drawbacks:
+
+- A volume for TPM data must be created, arranged and maintained properly.
+- Vdsm is responsible for arranging atomic data updates, to avoid ending up with partially written data in case storage gets disconnected.
+- Unless direct I/O is used, Vdsm can't be sure the data was actually written.
+
+This approach would be more complicated than using metadata but it would have the advantages mentioned above.
+
+#### Making swtpm to use shared storage directly
+
+The swtpm data location is selected by libvirt automatically but we could probably make a symbolic link to the desired shared location.  This approach would avoid data handling on the Vdsm side and all the updates would be stored immediately.  But it shares some of the drawbacks of using shared storage and there are additional risks:
+
+- Data consistency is at risk when storage gets disconnected during swtpm write.
+- swtpm reacts to write failures [by deleting the data completely](https://github.com/stefanberger/swtpm/blob/609dfd873a31fc0402752c50d1d5cfa58949507f/src/swtpm/swtpm_nvfile.c#L563), which may fail as well, but a question is if the data is ever attempted to write again without further updates.
+- We can't depend on all the swtpm implementation details.
+- libvirt removes the data after domain gets undefined.  We would have to prevent that, perhaps by removing the symbolic link before undefining the libvirt domain.
+
+The possible data consistency race would have to be solved some way (how?) to be able to use this approach.  The overall benefit of this approach depends on whether resolving the data consistency problem is easier than copying the data.


source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,70 @@
...
+- Adding a new parameter to VM.destroy call response containing the updated data.
+
+There is some fragility with that approach.  If VM.destroy call fails for any reason after the libvirt domain is undefined (and the corresponding TPM data is deleted by libvirt), the updated data will be lost.  But we don't have many alternatives: While we could make a separate API call to retrieve TPM data or attach them to VM stats, that would mean retrieving live data from swtpm, which may not be consistent.  Only data retrieved after the VM stops can be considered stable.  And even with a separate API call, if Engine loses contact with the host and starts the VM elsewhere, updates to TPM data since the VM start will be lost.  We cannot guarantee, without introducing complicated implementations, up-to-date TPM data if the environment doesn't work as expected.  Considering TPM data typically serves forensic purposes rather than storing critical information, we may accept the fact that VM guest OS repair is needed in case the data got changed and the environment experienced unexpected trouble (i.e. VM.destroy call hasn't been successful).  It may still cause trouble for HA VMs.
```
mz-pdm, line 61: I'm not sure the approach with VM.destroy is reliable enough itself even when not considering other problems. Considering that TPM data doesn't change that often under normal circumstances, it could be complemented with stats reporting. A periodic job would watch for TPM data changes and if the data changed & remained the same in the next cycle then it would be reported as part of VM stats. I'd need more knowledge about how things work in Engine to implement that correctly but I think the basic idea could work. What do you think?

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,70 @@
...
+
+An alternative mechanism could be to use shared storage to store TPM data, either by copying the data there or (preferably, if possible) to let swtpm use the shared storage directly.  The differences to the solution proposed above would be:
+
+- More complicated implementation.
+- Need to connect stored data to the corresponding (disk-only) snapshots, either on the storage, or by copying them to and from the database.
+- Better data safety in case of direct use of the shared storage by swtpm, the data is updated immediately, not only once VM stops running (but there may in theory data consistency problems again).
+- It's unclear how swtpm would work with network storage and how to prevent libvirt from deleting the data once the domain XML is deleted on the host.
+
+Overall, storing TPM data in the database looks much easier and so also more robust in the end result.
```
sandrobonazzola, line 70: Can you please add also a section on how to test the feature?

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,70 @@
...
+
+An alternative mechanism could be to use shared storage to store TPM data, either by copying the data there or (preferably, if possible) to let swtpm use the shared storage directly.  The differences to the solution proposed above would be:
+
+- More complicated implementation.
+- Need to connect stored data to the corresponding (disk-only) snapshots, either on the storage, or by copying them to and from the database.
+- Better data safety in case of direct use of the shared storage by swtpm, the data is updated immediately, not only once VM stops running (but there may in theory data consistency problems again).
+- It's unclear how swtpm would work with network storage and how to prevent libvirt from deleting the data once the domain XML is deleted on the host.
```
ahadas, line 68: That's an interesting approach.
Maybe we can combine the two together? I mean, to let VDSM saves the data to those volumes and read from them when starting the VMs instead of sending the data to the engine?

mz-pdm: I discussed the idea of putting the data to storage with Nir and it doesn't look like a very simple thing. A volume would be needed with all the related arrangements and it needn't be absolutely reliable too -- we know storage problems can occur for many reasons. We came to the conclusion that a better idea could be to make a special Vdsm API call for retrieving TPM data that would be called periodically. Engine could ask for the data with a relatively long interval or whenever TPM data checksum reported by Vdsm changes (what would be the complications you mentioned above with storing VM data hashes on the Engine side and acting when they change?). The last call would be performed just before calling the final Destroy (to avoid problems with failures after the data is wiped on the host).
This way, Engine would be managing the data transfers, based on some information from Vdsm, which looks more robust than other solutions. Are there any obstacles on the Engine side to implement something like that?

ahadas: I think that storing this data (in vm_dynamic) and sending it on RunVm should be fairly simple. The complication I thought of was more about a mechanism to get the hash and detecting a change. We do that for the 'devices hash' - we have VmDevicesMonitoring that on engine startup loads the hash of each VM from the database and then stores that in-memory and compares it with the one we get from the stats call. It's not rocket science to introduce another hash that this module or an additional module would handle, but it can be simpler to use the same 'devices hash' for that as well if possible (that hash already covers some of the guest agent data afair).

What I like about storing this data on the storage though is that it sounds more aligned with our desire to be able to (re)start VMs without the engine being involved. I guess we wouldn't want to store such sensitive data in the metadata, right?

mz-pdm: > What I like about storing this data on the storage though is that it sounds more aligned with our desire to be able to (re)start VMs without the engine being involved. 

That's a good remark and an argument important enough to consider using storage.

> I guess we wouldn't want to store such sensitive data in the metadata, right?

The data is already accessible to anybody with root access to the host and also, in most of the discussed implementation scenarios, to Vdsm. So I think having them in metadata is no big additional risk. But the question is whether we want to store larger data in metadata. While TPM capacity is AFAIK very limited (I'm not sure about emulated TPM though), it's still going to be at least 4 KB of encoded data with swtpm.

Still, using metadata would simplify things a lot. When using storage, we would have to arrange the volume, to ensure data can't be left in inconsistent state there (e.g. when the storage gets suddenly disconnected while TPM data is being written), and to deliver (consistent) data there, either by arranging swtpm to use the storage directly or making Vdsm copy it there.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Done.

mz-pdm:
The document has been reworked, section on testing was added, text on data security was separated and expanded, the implementation section was completely rewritten and expanded, and some more changes have been made. Four implementation approaches are described, we need to decide which one to attempt.

sandrobonazzola:
Waiting for @ahadas approval before merging.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+Availability of data stored in TPM shouldn't be critical.  Its loss can cause operational problems such as a need to initialize and set up TPM data again, to audit the operating system after the TPM data is reset (lost) or perhaps even to reinstall the whole system but it shouldn't cause real data loss.  When using bare metal, one can't rely on TPM data availability either because if the hardware stops working, the data is no longer accessible.  If one uses for example a private key stored only in TPM to encrypt and decrypt file system encryption keys, then the risk of losing data stored in the file system without having proper backups is at least as big with hardware TPM as with virtual TPM.
+
+## Limitations
```
ahadas, line 37: Would we need to block operations like export/clone of running VMs with TCP devices (those operations take a snapshot without memory, only of the disks, and correlate it with the VM configuration at the time of the snapshot)?

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+This approach is a bit simpler and more flexible than using API calls but has the data size drawback.
+
+#### Using shared storage
+
+A somewhat similar approach to using domain XML metadata would be to store the data to a shared storage.  Instead of writing to or reading it from the metadata, it would be written to or read from the storage.  The benefit would be avoiding domain XML pollution and unnecessary data transfers.  Additionally, Engine wouldn't have to process the data and maintain it in the database and the data could be updated even when Engine was temporarily unavailable.
+
+However, there are also drawbacks:
+
+- A volume for TPM data must be created, arranged and maintained properly.
```
ahadas, line 139: Another drawback is that the size of these volume would likely be at lease 1G on block storage (like volumes that store vm configuration for snapshots with memory). This appears to be a reoccurring issue that also came up in the context of Q35 features..

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+It should be also tested that VMs with TPM data created on hosts running older systems are unchanged when the VM is started on a host running the newest system.  This must be tested in the guest OS.
+
+## Changes to Engine and Vdsm
+
+There are no special changes needed in order to add a TPM device as such, it's just another kind of device added to Engine.
+
+### Data persistence
+
+A very different thing is ensuring TPM data persistence.  TPM data is managed using an external tool, swtpm.  Once the VM domain XML is deleted on the host, the TPM data is deleted as well by libvirt and it must be restored before the VM is run again, whether on this or another host (live host migrations are handled transparently and don't require any special actions on the oVirt side).
```
nirs, line 71: How live migration is handled? how it does not require anything on ovirt side?

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+There are several options how to provide TPM data on the VM start:
+
+- As a parameter to a VM.create API call.
+- In metadata section of the domain XML.
```
nirs, line 76: First 2 looks the same, data sent from engine when starting a vm.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+The hard part is how to store updated TPM data from the host reliably, making sure that:
+
+- Data updates are not lost.
+- The updated data is consistent, i.e. it can be read by swtpm correctly on the next VM start.
+
+We cannot handle all erroneous situations (e.g. faulty hardware corrupting data) but a reasonable level of resistance against common failures should be granted.  Ideally, the following situations should be handled:
+
+- When a VM stops for any reason on a connected host, we store the latest state of the data.
+- When a host or a storage get disconnected, we store the latest known state of the data.
+- When a VM is restarted without Engine assistance, it uses the final data from the last run.
```
nirs, line 88: We cannot support this. Engine assistance is always required to run vms, for example for attaching storage to host.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- We don't store inconsistent data.  swtpm apparently [doesn't store changes atomically](https://github.com/stefanberger/swtpm/blob/609dfd873a31fc0402752c50d1d5cfa58949507f/src/swtpm/swtpm_nvfile.c#L492), so whether we copy swtpm data or let swtpm write to shared storage, we are at risk of race with swtpm writes on copying or when the storage gets suddenly disconnected.
```
nirs, line 89: We depend on swtmp here. If it does not ensure consistency we cannot ensure this.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- Temporary network disconnections, lost events sent from Vdsm to Engine or temporary failures when delivering or processing Vdsm responses to Engine don't cause permanent loss of data updates.
+
+The options to deliver updated TPM data are similar to obtaining TPM data on VM start, with the important distinction that this time they are write instead of read operations, with the risk of data loss.  We can:
+
+- Send the updated data to Engine in events, current API call responses (stats, destroy) or using a new Vdsm API call for the given purpose.
```
nirs, line 94: events are not good way to pass the data, but they can be used for signaling changes. Getting the data should be via new API for this purpose, or adding to existing API that returns information about a vm.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- Store the data in metadata, to be read and processed in Engine monitoring.
+- Store the data to shared storage from Vdsm.
```
nirs, line 96: This does not make sense, requires creating special volume per vm. If we do this then there is no need to keep tpm data elsewhere, it will simply be in shared storage. This requires support from swtpm, they should be able to use block device for the data, and use direct I/O to ensure consistency.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- Let swtpm store the data directly to the shared storage.
```
nirs, line 97: Yes, this is a valid option, but this means adding new disk for tpm and handling the extra disk in all flows copying vm disks, like live storage migration. This is a lot of work.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- When a host or a storage get disconnected, we store the latest known state of the data.
+- When a VM is restarted without Engine assistance, it uses the final data from the last run.
+- We don't store inconsistent data.  swtpm apparently [doesn't store changes atomically](https://github.com/stefanberger/swtpm/blob/609dfd873a31fc0402752c50d1d5cfa58949507f/src/swtpm/swtpm_nvfile.c#L492), so whether we copy swtpm data or let swtpm write to shared storage, we are at risk of race with swtpm writes on copying or when the storage gets suddenly disconnected.
+- Temporary network disconnections, lost events sent from Vdsm to Engine or temporary failures when delivering or processing Vdsm responses to Engine don't cause permanent loss of data updates.
+
+The options to deliver updated TPM data are similar to obtaining TPM data on VM start, with the important distinction that this time they are write instead of read operations, with the risk of data loss.  We can:
+
+- Send the updated data to Engine in events, current API call responses (stats, destroy) or using a new Vdsm API call for the given purpose.
+- Store the data in metadata, to be read and processed in Engine monitoring.
```
nirs, line 95: This looks like abuse of the metadata, and also not reliabe as libvirt API can block for serveal reason.  I would not use it for anything except vm metadata.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+TPM data can be stored in the Engine database and transferred in both the directions using Vdsm API calls.  It can be provided to the VM as an additional parameter to VM.create call.  Data changes can be reported back to Engine using:
+
+- Events sent from Vdsm.
+- As a return value parameter of VM.destroy call.
+- As an additional value in VM stats.
+- On a request from Engine using a newly introduced Vdsm API call.
+
+TPM data can be copied to swtpm data location before the VM is created.  Once VM is started, updated TPM data can be read from the swtpm data location.  A problem is how to read data from swtpm consistently, avoiding collisions with swtpm writes.  Considering TPM data is small, written locally and not updated frequently, a good enough approach should be to have a periodic Vdsm job watching for swtpm data changes.  If it detects a data change, it keeps watching the data and once it detects the same data in two consecutive job runs, it can report it as changed data relatively safely (unless there is some fatal condition such as a full local disk).  This mechanism would be common to all the approaches except for shared storage directly used by swtpm.
```
nirs, line 111: This is not so simple, since the data may change again after we detected a change, but it can work.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- Data read after the VM stops can be considered stable and reported directly.  However, if it is attached to VM.destroy response and the response fails for any reason, it can't be reported again in a response to a repeated VM.destroy call, because the data is already deleted after the libvirt domain has been undefined.
+- A specific API call would solve most of the problems.  Vdsm could include TPM data into the computation of the VM device hash sent within VM stats and Engine could request TPM data itself if the device hash changes.  Alternatively, Engine could make the API call periodically, providing TPM data hashes as an argument, and Vdsm would respond only with data not matching the hashes.  An additional call would happen in either case just before calling VM.destroy to get the ultimate final version of TPM data.
+- It is fully dependent on Engine cooperation, it's not possible to start a VM without Engine explicitly providing the data.  But if we needed just to restart a VM on the same host without Engine then we could retain TPM data by either not undefining the domain XML or by making a local copy of the data between the restarts.
```
nirs, line 119: We cannot start a vm on same host without enigne since vm xml and metadata is deleted when stooping a vm.

I think the only way that will work is to have new API called by engine to fetch tpm data, and 
split destroying a vm to 2 steps - stop the vm and undefine the vm. This way the vm data is not
lost.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+This approach should be feasible with introducing a new Vdsm API call as outlined above but it has the drawback that Engine assistance is needed.
+
+#### Storing the data in domain XML metadata
+
+Using domain XML metadata, it's possible to transfer TPM data in all the directions.  TPM data can be read from the data on VM start, with or without Engine assistance, updates can be stored there atomically and Engine can update the data in monitoring.
```
nirs, line 125: But vm metadata is lost when you stop a vm, I don't see how it helps.

source/develop/release-management/features/virt/tpm-device.md
nirs: No, we can create 128m disks.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+A somewhat similar approach to using domain XML metadata would be to store the data to a shared storage.  Instead of writing to or reading it from the metadata, it would be written to or read from the storage.  The benefit would be avoiding domain XML pollution and unnecessary data transfers.  Additionally, Engine wouldn't have to process the data and maintain it in the database and the data could be updated even when Engine was temporarily unavailable.
+
+However, there are also drawbacks:
+
+- A volume for TPM data must be created, arranged and maintained properly.
+- Vdsm is responsible for arranging atomic data updates, to avoid ending up with partially written data in case storage gets disconnected.
+- Unless direct I/O is used, Vdsm can't be sure the data was actually written.
+
+This approach would be more complicated than using metadata but it would have the advantages mentioned above.
```
nirs, line 143: This approach makes sense only if swtpm writes to shared storage. If vdsm has to read and write this will only complicate things.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Making swtpm to use shared storage directly
+
+The swtpm data location is selected by libvirt automatically but we could probably make a symbolic link to the desired shared location.  This approach would avoid data handling on the Vdsm side and all the updates would be stored immediately.  But it shares some of the drawbacks of using shared storage and there are additional risks:
```
nirs, line 147: In this case libivrt will have to accept a path to tmp data.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Data consistency is at risk when storage gets disconnected during swtpm write.
```
nirs, line 149: This is not an issue if swtpm is written properly, ensuring atomic updates.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- swtpm reacts to write failures [by deleting the data completely](https://github.com/stefanberger/swtpm/blob/609dfd873a31fc0402752c50d1d5cfa58949507f/src/swtpm/swtpm_nvfile.c#L563), which may fail as well, but a question is if the data is ever attempted to write again without further updates.
```
nirs, line 150: deleting on failure is not valid approach. 

The only way that will work is copy-on-write:

- Keep metadata sector pointing to the current data area
- keep 2 data areas

For each write, write to non-current data area. When the write is done
(fsync() succeeded), update the metadata sector to to point to this area.

On failure, the metadata sector still points to previous data area, no change
was done, and there nothing to cleanup.

On success, the metadata sector points to the new data.

If swtmp does not have something like this, there is no point to let it write
to shared storage since it will create inconsistent data.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- We can't depend on all the swtpm implementation details.
```
nirs, line 151: Of course we can, this is exactly what we should do - move the complexity down to the platform.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+- libvirt removes the data after domain gets undefined.  We would have to prevent that, perhaps by removing the symbolic link before undefining the libvirt domain.
```
nirs, line 152: We can ask libivrt to support external tpm data, so libvirt does not create or remove anything. This way we have
complete control over the data, and we don't need to change libvirt flows.

So we can do:

1. Store tpm data on the host tpm store
2. Start vm with tpm store path
3. Stop vm (undefine, etc.)
4. Pass current tpm data to engine

We use the same method for other stuff like scratch disks during backup. Libvirt supports EXTERNAL flags to let the user control something that libvirt cannot mange. This sounds like the best approach.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Did you mean "with TPM devices"? That's a good question, I'm afraid we can fall into trouble with all kinds of disk snapshots when the VM is running. A possible solution could be checking TPM data before starting the snapshot and before finalizing it and if it doesn't match then to let the snapshot fail.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Live migration is handled transparently by libvirt. The only trouble with it could be if swtpm was arranged to write to the same shared storage volume directly on both the source and the destination.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Yes, but not handled the same way later.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: If the storage is already attached (on the same host where the VM was running last time) then it may be possible. It's a completely different topic though, all we need here now is not to close the door to the functionality with TPM handling.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: We may be forced to depend on some heuristics, unless upstream is willing to implement or accept improvements, such as if the local data doesn't change for a sufficiently long time interval then it can be considered OK.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Did you mean "tpm"? Yes, adding such a feature would be an option, but probably only with other arrangements such as prevention of data removal by libvirt.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Unfortunately it's not and we should explore whether that could be changed (in a sufficiently close future).

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Yes.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: Ideally yes, but it may be a long run to fix upstream and get the fixes in downstream and it may not guarantee things won't break again in the future. Not that easy, we may still attempt that but other options for interim solutions must be also considered at the moment.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: This would be necessary only when swtpm was writing to the shared storage directly. I like that option the least of all. In the other cases it makes sense that libvirt removes related data when we undefine a domain -- it's up to us to delay cleanup if we need that.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: ... and is non-empty etc.

source/develop/release-management/features/virt/tpm-device.md
ahadas: Ack, so it seems that we converge here to not store the TPM data on the storage domain but it could be useful to create smaller disks for snapshot's metadata volumes - allocating 1g as we do today for xml and some other vdsm properties is a waste

source/develop/release-management/features/virt/tpm-device.md
ahadas: right, that's a bit risky. so another option that we discussed off-line is to treat the TPM data as yet another data which does not change frequently, like the VM devices. so the computation of the reported hash will also take the TPM data into account and when the engine calls `dump` it will not only get the domain xml but also the TPM data as another in the dictionary. this mechanism may also possibly be used for reporting the secure boot's data later on.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+Availability of data stored in TPM shouldn't be critical.  Its loss can cause operational problems such as a need to initialize and set up TPM data again, to audit the operating system after the TPM data is reset (lost) or perhaps even to reinstall the whole system but it shouldn't cause real data loss.  When using bare metal, one can't rely on TPM data availability either because if the hardware stops working, the data is no longer accessible.  If one uses for example a private key stored only in TPM to encrypt and decrypt file system encryption keys, then the risk of losing data stored in the file system without having proper backups is at least as big with hardware TPM as with virtual TPM.
+
+## Limitations
+
+TPM devices can be used only with UEFI BIOS.
+
+TPM devices are currently not available on POWER (TODO: re-check).
+
+Until https://bugzilla.redhat.com/1855367 is implemented in libvirt, VMs with TPM devices cannot have snapshots with memory.  The problem is that we need swtpm data snapshot consistent with the VM snapshot, which is not possible to get without libvirt assistance.
```
sandrobonazzola, line 43: please use <https://bugzilla.redhat.com/1855367> so it will also be available as link

mz-pdm: Done.

source/develop/release-management/features/virt/tpm-device.md
mz-pdm: I added it as another option to the document.

source/develop/release-management/features/virt/tpm-device.md
```diff
@@ -0,0 +1,154 @@
...
+
+Since the emulated TPM is a software rather than a hardware device, its data updates can get lost, e.g. when a host suddenly crashes.  (TODO: Clarify according to the selected implementation.)
+
+There is apparently no guarantee that new swtpm versions can still read its old data.  That should be watched in testing.
+
+## User Experience
+
+Ideally, a TPM device would be added to each VM automatically and would be automatically available to the guest OS.  However, TPM support is a new feature and we cannot guarantee that everything is correctly supported in the underlying platform (for example, snapshots with memory are currently not working as expected).  For that reason, it's better to not add TPM devices to VMs by default for now and make a cluster option to enable them for the given cluster, similar to memory ballooning.
```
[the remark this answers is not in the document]
ahadas: so let's remove that part?

ahadas (commented):
from my side, let's just remove the cluster-level setting part and get this in

mz-pdm:
Replaced TPM cluster setting with a per-VM setting.

ahadas (approved):

sandrobonazzola:
There are a few conversation not resolved and a pending review request for @nirs . Can you please confirm if we can merge this or if we need further discussion?

mz-pdm (commented):
Minor changes:

- Removed unavailability on POWER (it should be available in 8.3).
- Upstream swtpm fix mentioned.
- Cluster->VM option in Testing section.

I believe the document describes the current state of our thoughts. We can and will change it later as needed.

ahadas:
> I believe the document describes the current state of our thoughts. We can and will change it later as needed.

+1
lgtm

sandrobonazzola:
Merging based on @ahadas approval.

PR #2319, opened 2020-08-25
features warning: add spaces around strong text

vjuranek:
Current spaces around strong text are ignored. Use &nbsp; to add space
around strong text in feature pages warning.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @vjuranek

### DIFF
--- a/source/_includes/feature.haml
+++ b/source/_includes/feature.haml
@@ -3,16 +3,16 @@
   Feature pages are design documents that developers have created while collaborating on oVirt.
   %br
   %br
-  Most of them are 
+  Most of them are&nbsp;
   %strong
     outdated
   , but provide historical design context.
   %br
   %br
-  They are 
+  They are&nbsp;
   %strong
     not 
-  user documentation and should not be treated as such.
+  &nbsp;user documentation and should not be treated as such.
   %br
   %br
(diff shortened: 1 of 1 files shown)


source/_includes/feature.haml
```diff
@@ -3,16 +3,16 @@
   Feature pages are design documents that developers have created while collaborating on oVirt.
   %br
   %br
-  Most of them are 
+  Most of them are&nbsp;
   %strong
     outdated
   , but provide historical design context.
   %br
   %br
```
nirs, line 11: This is not the right direction. Remove the "strong" tag instead. If a message need to use
"strong" the style should be set using css, not by this kind of ugly inline markup.

I would also remove this warning, it is not needed.

nirs (changes requested):

source/_includes/feature.haml
sandrobonazzola: Warning is there for a good reason, not going to be removed. the strong tag here is used in conformance to its definition:https://www.w3schools.com/tags/tag_strong.asp ; the fact it shows things in bold is merely a css configuraton, I agree.

PR #2320, opened 2020-08-25
cd-change: add design doc for CD management

vjuranek:
Add desing doc for functions and flows related to management fo CD-ROM:
for inserting, ejecting and changing CD.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @vjuranek 

This pull request needs review by: @nirs

### DIFF
--- a/source/develop/release-management/features/storage/cd-block-storage.md
+++ b/source/develop/release-management/features/storage/cd-block-storage.md
@@ -1,0 +1,309 @@
+---
+title: CD-ROM support for block storage
+category: feature
+authors: vjuranek, nsoffer
+feature_name: CD-ROM support for block storage
+feature_modules: engine,vdsm
+feature_status: Design
+---
+
+# CD-ROM support for block storage
+
+
+## Summary
+
+This feature adds new implementation for CD-ROM management no block storage domains,
+namely for loading, ejecting and changing CD-ROM.
+
+## Owner
+
+- Vojtech Juranek (<vjuranek@redhat.com>)
+
+## Overview
+
+### What was available before this feature?
+
+Before this feature, oVirt engine provided functions for changing and ejecting CD-ROM.
+However, non of these functions worked properly with block based storage.
+Changing CD failed on block storage as the appropriate volume wasn't activated.
+Ejecting CD failed to deactivate the volume.
+For more details, see following bugs.
+
+#### CD-ROM related XML
+
+Currently each VM contains CD-ROM (without CD), e.g.
+
+```xml
+    <disk type='file' device='cdrom'>
+      <driver name='qemu' error_policy='report'/>
+      <source startupPolicy='optional'/>
+      <target dev='sdc' bus='sata'/>
+      <readonly/>
+      <alias name='ua-79287c04-4eea-4db7-a376-99a9f85ad0ed'/>
+      <address type='drive' controller='0' bus='0' target='0' unit='2'/>
+    </disk>
+```
+
+When VM is booted with CD, it chages e.g. to
+
+```xml
+    <disk type='block' device='cdrom'>
+      <driver name='qemu' type='raw' error_policy='report'/>
+      <source dev='/rhev/data-center/mnt/blockSD/88252cf6-381e-48f0-8795-a294a32c7149/images/89f05c7d-b961-4935-993f-514499024515/626a493f-5214-4337-b580-96a1ce702c2a' index='2'>
+        <seclabel model='dac' relabel='no'/>
+      </source>
+      <backingStore/>
+      <target dev='sdc' bus='sata'/>
+      <readonly/>
+      <alias name='ua-79287c04-4eea-4db7-a376-99a9f85ad0ed'/>
+      <address type='drive' controller='0' bus='0' target='0' unit='2'/>
+    </disk>
+```
+
+so that it contains path to the ISO.
+
+Besides this libvirt XML, oVirt adds its own XML for each disk to be able to identify
+the volume by PDIV (pool, domain, image volume) coordinates and includes other useful info like lease path. e.g.
+
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc">
+        <ovirt-vm:domainID>88252cf6-381e-48f0-8795-a294a32c7149</ovirt-vm:domainID>
+        <ovirt-vm:imageID>89f05c7d-b961-4935-993f-514499024515</ovirt-vm:imageID>
+        <ovirt-vm:poolID>13345997-b94f-42dd-b8ef-a1392f65cebf</ovirt-vm:poolID>
+        <ovirt-vm:volumeID>626a493f-5214-4337-b580-96a1ce702c2a</ovirt-vm:volumeID>
+        <ovirt-vm:volumeChain>
+            <ovirt-vm:volumeChainNode>
+                <ovirt-vm:domainID>88252cf6-381e-48f0-8795-a294a32c7149</ovirt-vm:domainID>
+                <ovirt-vm:imageID>89f05c7d-b961-4935-993f-514499024515</ovirt-vm:imageID>
+                <ovirt-vm:leaseOffset type="int">105906176</ovirt-vm:leaseOffset>
+                <ovirt-vm:leasePath>/dev/88252cf6-381e-48f0-8795-a294a32c7149/leases</ovirt-vm:leasePath>
+                <ovirt-vm:path>/rhev/data-center/mnt/blockSD/88252cf6-381e-48f0-8795-a294a32c7149/images/89f05c7d-b961-4935-993f-514499024515/626a493f-5214-4337-b580-96a1ce702c2a</ovirt-vm:path>
+                <ovirt-vm:volumeID>626a493f-5214-4337-b580-96a1ce702c2a</ovirt-vm:volumeID>
+            </ovirt-vm:volumeChainNode>
+        </ovirt-vm:volumeChain>
+    </ovirt-vm:device>
+```
+
+When CD is not loaded, this element is still present (as CD-ROM is always present), but is empty, e.g.:
+
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc"/>
+```
+
+### Bugs
+
+- [Bug 1589763: Error changing CD for a running VM when ISO image is on a block domain](https://bugzilla.redhat.com/1589763)
+- [Bug 1868643: Volume not deactivated after ejecting CDROM](https://bugzilla.redhat.com/1868643)
+
+
+### Functionality added with this feature
+
+This feature reworks existing vdsm API for loading, ejecting and
+changing CD to be able to work on block storage domains.
+From implementation point of view it is only one function,`ChangeCD`.
+This function will handle all aforementioned cases - loading CD, ejecting CD and changing CD.
+It  will also ensure, that the CD will be changed atomically.
+This means that the change either happens or previous state will be preserved,
+e.g. if user wants to change CD by another and activation of new CD fails, CDROM will still hold
+old CD once the operation finishes.
+
+
+#### Vdsm
+
+- Rework `VM.changeCD` API implementation.
+- Updated recovery flow to check `status` flag for CDs.
+- If ISO is not specified by PDIV but by path, fall back to old behaviour.
+
+
+### Flows
+
+#### VM with empty tray, insert CD
+
+1. engine sends `changeCD` command, specifying ISO by PDIV format.
+1. add new drive spec to VM metadata with new attribute `state="loading"`
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc" state="loading">
+```
+1. prepare new drive spec
+  - if failed:
+      1. remove drive spec (log errors)
+      1. fail the flow with original error
+1. loads CD into VM
+  - if failed:
+      1. tear down the volume (log error)
+      1. remove the drive spec (log error)
+      1. fail the flow with the original error
+1. remove `state` attribute from VM metadata
+  - if failed: log the error and succeed
+
+
+#### VM with loaded CD, eject CD
+
+1. engine sends `changeCD` command with empty PDIV format.
+1. vdsm updates VM metadata with new attribute `state="ejecting"`:
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc" state="ejecting">
+```
+1. vdsm eject CD from VM (calling libvirt `updateDeviceFlags()`)
+  - if failed (e.g. timeout?):
+      1. remove the `state` attribute (log error)
+      1. fail with original error
+1. tear down drive spec
+  - if tearing down volume fails: log error if volume used by another VM and succeed
+  - if error is anything else except device in use error, remove `state` attribute from metadata, re-throw the exception and fail
+1. remove drive spec from VM metadata
+  - if failed: log error and succeed
+
+
+#### VM with loaded CD, changeCD
+
+1. engine sends `changeCD`, specifying new ISO path by PDIV format and also path.
+1. update exiting drive spec for loaded CD with `state="ejecting"`:
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc" state="ejecting">
+```
+1. add another drive spec to VM metadata for CD being loaded with attribute `state="loading"`
+```xml
+    <ovirt-vm:device devtype="disk" name="sdc" state="loading">
+```
+1. prepare new drive spec
+  - if failed:
+      1. remove `state="ejecting"` from loaded CD metadata
+      1. remove newly added drive spec (log errors)
+      1. fail the flow with original error
+1. loads new CD into VM (by calling libvirt `updateDeviceFlags()`)
+  - if failed:
+    1. tear down new CD volume (log error)
+    1. remove the new CD drive spec (log error)
+    1. remove `state="ejecting"` from old drive spec metadata
+    1. fail the flow with the original error
+1. remove the `state="loading"` attribute from new drive spec metadata
+  - if failed: log the error and continue
+1. tear down old drive spec
+  - if failed (e.g. timeout?):
+      1. log the error and continue (tearing down new CD would probably fail
+      as well and we would end up with unused activated volume anyway, so
+      there's no reason to fail it now)
+1. remove old CD drive spec
+  - if failed: log the error and succeed
+
+
+#### VM recovery flow
+
+Try to cleanup storage leftovers from loading and ejecting operations without changing VM state.
+
+
+##### VM with empty tray, insert CD
+
+1. metadata `state="loading"`, there are no other matadata for CDROM, volume is prepared, CD is loaded
+  - remove `state="loading"` from metadata from XML
+1. metadata `state="loading"`, there are no other matadata for CDROM, volume is prepared, CD isn't loaded
+  - tear down the volume and remove drive spec from XML
+1. metadata `state="loading"`, there are no other matadata for CDROM, volume is torn down, CD isn't loaded
+  - remove drive spec metadata
+
+
+##### VM with loaded CD, eject CD
+
+1. metadata `state="ejecting"`, there are no other matadata for CDROM, volume is prepared, CD is loaded
+  - remove `state="ejecting"` from metadata from XML
+1. metadata `state="ejecting"`, there are no other matadata for CDROM, volume is prepared, CD not loaded
+  - tear down the volume and remove `state="ejecting"` from metadata from XML
+1. metadata `state="ejecting"`, there are no other matadata for CDROM, volume is torn down, CD not loaded
+  - remove `state="ejecting"` from metadata from XML
+
+
+##### VM with loaded CD, changeCD
+
+1. metadata `state="ejecting"`, there are no other matadata for CDROM
+  - see previous paragraph
+1. there are two metadata for CDROM, one with `state="ejecting"`, second with `state="loading"`, volume for loading is torn down
+  - remove metadata for loading from VM XML
+  - remove `state="ejecting"` from ejecting metadata
+1. there are two metadata for CDROM, one with `state="ejecting"`, second with `state="loading"`, volume for loading is prepared, CDROM still contains ejecting CD
+  - tear down loading CD volume
+  - remove metadata for loading CD
+  - remove `state="ejecting"` from metadata
+1. there are two metadata for CDROM, one with `state="ejecting"`, second with `state="loading"`, volume for loading is prepared, CDROM still contains loading CD
+  - tear down ejecting CD volume
+  - remove metadata for ejecting CD
+  - remove `state="loading"` from metadata
+1. there are two metadata for CDROM, one with `state="ejecting"`, second without `state` attribute, volume for ejected CD is prepared
+  - tear down ejecting CD volume
+  - remove metadata for ejecting CD
+1. there are two metadata for CDROM, one with `state="ejecting"`, second without `state` attribute, volume for ejected CD is torn down
+  - remove metadata for ejecting CD
+
+
+`state` attribute is an optimization for easy cleanup.
+E.g. in case of hot disk plug, this is not used and if vdsm fails between activation of the volume
+and pluging disk into VM, unused active disk is left in the storage.
+
+
+### Changes in Engine
+
+Engine will only use PDIV in ChangeCD command and there will be no changes visible to the user.
+
+- Update ChangeCD to use PDIV for CD specification on 4.5 cluster version
+
+
+### Changes in SDK
+
+`VmCdromResource` used SDK uses `ChangeDisk` command.
+Once internal engine implementation will switch to using PDIV, this will be available also in SDK.
+Therefore, similarly to engine, there will be no changes in public SDK API visible to user.
+
+
+## Installation/Upgrade
+
+Changed implementation of `ChangeCD` we delivered with new vdsm packages containing these changes.
+Changes in engine will require appropriate vdsm version to use new vdsm APIs.
+Once user upgrades to this engine version, changes will be available to the user.
+
+New `ChangeCD` implementation will be used for ISOs on DATA domains.
+Old  `ChangeCD` implementation will be used for ISOs on old ISO domains.
+
+
+## Testing
+
+### Test setup
+
+- create block storage domain
+- upload an ISO into block storage domain
+- create VM with CD
+
+### Preposed testing scenarios
+
+- eject CD from CD-ROM
+  - verify CD was removed in engine UI
+  - verify that metadata related to this CD-ROM was removed from VM XML
+  - verify ISO volume was deactivated on the host
+
+- load CD into CD-ROM
+  - verify CD was loaded in engine UI
+  - verify that metadata related to this CD-ROM was added into VM XML
+  - verify ISO volume was activated on the host
+  - verify ISO is not visible to VM (e.g by running `virsh domblklist`)
+
+- eject CD from CD-ROM
+  - verify CD was removed in engine UI
+  - verify that metadata related to this CD-ROM was removed from VM XML
+  - verify ISO volume was deactivated on the host
+  - verify ISO is visible to VM (e.g by running `virsh domblklist`)
+
+- change CD in CD-ROM
+  - verify new CD was loaded in engine UI
+  - verify that metadata related to old CD-ROM was removed from VM XML
+  - verify that metadata related to new CD-ROM was added into VM XML
+  - verify old ISO volume was deactivated on the host
+  - verify new ISO volume was activated on the host
+  - verify old ISO is not visible to VM (e.g by running `virsh domblklist`)
+  - verify ISO is visible to VM (e.g by running `virsh domblklist`)
+
+
+### Related links
+
+- [Discussion in ovirt devel mailing list.](https://www.mail-archive.com/devel@ovirt.org/msg16082.html)
+- [Bug 1725915: Vdsm tries to tear down in-use volume of ISO in block storage domain](https://bugzilla.redhat.com/1725915)
+
+


source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
+---
+title: CD-ROM Management
+category: feature
+authors: vjuranek, nsoffer
+feature_name: CD-ROM Management
+feature_modules: engine,vdsm
+feature_status: Design
+---
+
+# CD-ROM Management
```
nirs, line 10: CD-ROM support for block storage?

We already support CD management.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+## Owner
+
+- Vojtech Juranek (<vjuranek@redhat.com>)
+
+## Overview
+
+### What was available before this feature?
+
+Before this, feature oVirt engine provided functions for changing and ejecting CD-ROM.
+However, non of these functions worked properly.
```
nirs, line 26: with block based storage.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+Changing CD failed on block storage as the appropriate volume wasn't activated.
+Ejecting CD failed to deactivate the volume.
+For more details, see following bugs.
+
+### Bugs
+
+- [Error changing CD for a running VM when ISO image is on a block domain](https://bugzilla.redhat.com/1589763)
+- [Volume not deactivated after ejecting CDROM](https://bugzilla.redhat.com/1868643)
```
nirs, line 34: Including the bug number in the title can help (e.g. Bug xxxyyy: text ...)

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+### Related links
+
+- [Discussion in ovirt devel mailing list.](https://www.mail-archive.com/devel@ovirt.org/msg16082.html)
+- [Vdsm tries to tear down in-use volume of ISO in block storage domain](https://bugzilla.redhat.com/1725915)
```
nirs, line 40: Bug 1725915: ...

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+Ejecting CD failed to deactivate the volume.
+For more details, see following bugs.
+
+### Bugs
+
+- [Error changing CD for a running VM when ISO image is on a block domain](https://bugzilla.redhat.com/1589763)
+- [Volume not deactivated after ejecting CDROM](https://bugzilla.redhat.com/1868643)
+
+
+### Related links
```
nirs, line 37: I would move this to the end of the page.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+
+- [Discussion in ovirt devel mailing list.](https://www.mail-archive.com/devel@ovirt.org/msg16082.html)
+- [Vdsm tries to tear down in-use volume of ISO in block storage domain](https://bugzilla.redhat.com/1725915)
+
+
+### Functionality added with this feature
+
+This feature propose new functions to both engine and vdsm for loading, ejecting and changing CD.
```
nirs, line 45: propose -> add
functions -> API

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+#### Vdsm
+
+- Add VM.insertCD API
+- Add VM.ejectCD API
+- Updated recovery flow to check `in_progress` flag for CDs
+
+#### Engine
+
+- Add InsertCD sub-command
+- Add EjectCD sub-command
```
nirs, line 57: These are internal commands, not a new public API.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+- Update ChangeCD to use InsertCD and EjectCD on 4.5 cluster version
```
nirs, line 58: This is not new API. I think this section belongs in the detailed design part.

The new APIs are only in vdsm, this is not user visible change.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Flows
+
+#### VM with empty tray, insert CD
+
+1. engine sends `insertCD` command, specifying ISO by PDIV format
+1. add new drivespec to VM metadata with new attribute `in_progress`
+```
+    <ovirt-vm:device devtype="disk" name="sdc" in_progress="yes">
+```
```
nirs, line 69: indentation not needed when using ``` style. If you want to indent you can remove the back-quotes.

source/develop/release-management/features/storage/cd-management.md
```diff
@@ -0,0 +1,117 @@
...
+  - remove metadata from XML
+1. metadata in progress, volume active, device exists
+  - remove attribute `in_progress` from metadata
+
+
+#### VM with loaded CD, changeCD
+
+1. engine sends ejectCD, specifying ISO path by PDIV format
+1. engine sends insetrCD command, specifying ISO by PDIV format
+
```
nirs, line 117: The internal changes in the system should be here.

Missing info:
- How this feature can be tested?
- Backward compatibility? when new APIs can be used?
- Does it affects upgrades?

Check the feature template for other missing info.

nirs:
@mz-pdm, @ahadas, please review this.

source/develop/release-management/features/storage/cd-management.md
vjuranek: for some reason without ``` it doesn't work (at least on my local instance). So ``` are there for proper rendering as HTML and indentation is for reading as a plain text

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+```
+    <ovirt-vm:device devtype="disk" name="sdc" in_progress="yes">
+```
+1. prepare new drivespec
+  - if failed:
+    1. remove drivespec (log errors)
+    1. fail the flow with original error
+1. loads CD into VM
+  - if failed:
+    1. teardown the volume (log error)
```
ahadas, line 125: do we prevent tearing down a volume that is used by other vms at the same time?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Functionality added with this feature
+
+This feature adds new APIs to vdsm for loading, ejecting and changing CD.
+With these functions, all aforementioned functionality will work properly.
+
+#### Vdsm
+
+- Add VM.insertCD API
+- Add VM.ejectCD API
```
ahadas, line 106: few years ago we started changing the virt-related parts in the engine<->vdsm api to be closer to that of libvirt
we changed create-vm but also attach/detach nic (see HotPlugNicVDSCommand and HotUnplugNicVDSCommand) and didn't get to change-cd
the nice parts in this approach:
1. VDSM doesn't need to generate the XML
2. we can reuse the same code that generates the XML for create-VM
3. by adding the xml to the dictionary of the existing verb we don't need new cluster-level or host capability or anything like that - new VDSM respect the XML if it is provided and otherwise use the other provided parameters. old VDSM ignores the XML.

can we do something similar in this case?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+
+Try to cleanup storage leftovers from insert and remove operations without changing VM state.
+
+1. metadata in progress, volume is not active
+  - remove metadata from XML
+1. metadata in progress, volume active, no device
+  - deactivate the volume
+  - remove metadata from XML
+1. metadata in progress, volume active, device exists
+  - remove attribute `in_progress` from metadata
```
ahadas, line 159: what does the `in_progress` field cover in addition to the state where the medata exists and the device doesn't exist?

ahadas (commented):
@vjuranek, @nirs  thanks for pulling this off!

source/develop/release-management/features/storage/cd-block-storage.md
michalskrivanek: +1 
it should be as close to it as possible. That's our "API" to modify VMs.
We don't have the path to pass to libvirt here as it's going to be made up by prepareVolume, but still I don't see a reason to differ here.

Also, empty and inserted CD are represented in libvirt by an "empty path" and "some path". We can keep it the same here and have some PDIV vs empty.

michalskrivanek: Also, can you clarify if you mean to keep the API synchronous? I.e. do all those actions, activation, etc, are completed before the API call returns.

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+1. vdsm updates VM metadata with new attribute `in_progress`:
+```
+    <ovirt-vm:device devtype="disk" name="sdc" in_progress="yes">
+```
+1. vdsm eject CD from VM (calling libvirt `updateDeviceFlags()`)
+  - if failed (e.g. timeout?):
+    1. remove the `in_progress` flag (log error)
+    1. fail with original error
+1. tear down drivespec
+  - if tearing down volume fails: log error (volume used by another VM)
```
mz-pdm, line 144: And succeed, right?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Overview
+
+### What was available before this feature?
+
+Before this, feature oVirt engine provided functions for changing and ejecting CD-ROM.
+However, non of these functions worked properly with block based storage.
+Changing CD failed on block storage as the appropriate volume wasn't activated.
+Ejecting CD failed to deactivate the volume.
+For more details, see following bugs.
```
mz-pdm, line 29: I think the guest OS will also be more happy if there is no CD change without "opening the tray" anymore.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: Is "volume used by another VM" the only error that can happen here?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+  - if tearing down volume fails: log error (volume used by another VM)
+1. remove drivespec from VM metadata
+  - if failed: log error and succeed
+
+
+#### VM recovery flow
+
+Try to cleanup storage leftovers from insert and remove operations without changing VM state.
+
+1. metadata in progress, volume is not active
```
mz-pdm, line 153: How exactly is "active" related to the flows? Is it "not active" -> "prepare new drivespec" -> "active" -> "teardown" -> "not active"? It would be worth to clarify and/or using common terminology to avoid any confusion.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm, line 159: Can there be active asynchronous operations in progress that could do something during or after the cleanup? For example, could it happen that CD removal starts, Vdsm  restarts, identifies the case "metadata in progress, volume active, device exists" and then the CD removal in libvirt finishes?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+    1. teardown the volume (log error)
+    1. remove the drive spec (log error)
+    1. fail the flow with the original error
+1. remove `in-progress` attribute from VM metadata
+  - if failed: log the error and succeed
+
+
+#### VM with loaded CD, eject CD
+
+1. engine sends `ejectCD`, specifying ISO path by PDIV format
```
mz-pdm, line 134: Why is ISO path needed to eject a CD? Isn't / shouldn't the information be available in the device and metadata parts of the domain XML?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+On engine side also two new internal commands will be added:
+
+- Add InsertCD sub-command
+- Add EjectCD sub-command
+
+
+## Installation/Upgrade
+
+New vdsm APIs will we delivered with new vdsm packages contaning these changes.
+Changes in engine will require approprate vdsm version and use new vdsm APIs.
```
mz-pdm, line 182: " .. and use .." or " ... to use ..."? Will the new functionality be activated depending on the cluster version?

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,228 @@
...
+  - verify old ISO is not visible to VM (e.g by running `virsh domblklist`)
+  - verify ISO is visible to VM (e.g by running `virsh domblklist`)
+
+
+### Related links
+
+- [Discussion in ovirt devel mailing list.](https://www.mail-archive.com/devel@ovirt.org/msg16082.html)
+- [Bug 1725915: Vdsm tries to tear down in-use volume of ISO in block storage domain](https://bugzilla.redhat.com/1725915)
+
+
```
mz-pdm, line 228: Please proofread (spell checking, trivial grammar errors) before submitting the final version.

source/develop/release-management/features/storage/cd-block-storage.md
ahadas: @michalskrivanek about the 'prepared' path, the engine already knows how to predict it on create-vm (see [LibvirtVmXmlBuilder#writeDiskSource](https://github.com/oVirt/ovirt-engine/blob/8203fad23f1de20c6c800a2e4352464bb230ab3d/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java#L2116)) - we can call the same code

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: as mentioned bellow, metadata always exists, as CD is always present. We can skip it, but simplifies the check in which state we are, so unless there are good reason not to have, I'd keep it.

source/develop/release-management/features/storage/cd-block-storage.md
michalskrivanek: oh, right, thank you, i forgot that we actually finished that

source/develop/release-management/features/storage/cd-block-storage.md
michalskrivanek: metadata exists, <disk> definition exists. 
But the path in source file is what matters.
It does seem redundant to have an extra field. And it requires more metadata updates when used as a persistent store for the flag...We should try to minimize the amount of updates needed, that's an expensive operation.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: I'm not sure I understand here. By active I mean active volume. On block devices, we need to active LVM volume before use and deactivate once not used, so this covers a situation when we load CD, add `in_progress` attribute into metadata, but vdsm fails/is killed/whatever before it activates the volume.

As for terminology, activate/deactivate volume is common terminology, unless I'm missing something or both of us speak about different things.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: I need to get more familiar with virt vdsm part to be able to answer this, but IIRC handling async livirt events is not properly done in vdsm and this can happen. However, IMHO fixing it is far behind the scope of this feature and if we want to handle it, we need propre async handling of libvirt events.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: yes, will use new API only for cluster version 4.5 (mentioned in Changes in engine par)

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: path in the source doesn't express whole state, volume can be active/inactive and path also doesn't say anything about the state when vdsm restarts - if there was some operation in progress which needs cleanup or the CD was already inserted, but we e.g. fail to activate the volume.

michalskrivanek: and how does it help you to have a simple flag in metadata? That's another operations which may fail or vdsm can crash in between and you're have the same problem. Or rather more - now you also have this potentially wron "in_progress" you have to reconcile. (e.g. "you add in_progress to metadata, vdsm crashes right afterwards before you activate")

Yes, you need to find out if the volume is active/inactive when you see a path in disk. But you have to do it anyway. And it's better to check the actual state rather then rely on a flag which suffers from the same problem.

michalskrivanek: if CD is already there (<disk> has a path) but you fail to activate then it's not valid and doesn't really work, right? I don't think it's anything new...right now you can also just use a non-existant path on iso domain (or remove the file in the middle). We don't have that many complaints on that....if it's at VM start then it's dropped (startupPolicy optional). If it's later then it's less nice...but again, we're no handling that today either. You don't have to fix whole vdsm in this effort:)

vjuranek: well, so it's fine to remove CD e.g. if the volume is not active no matter if it was successfully inserted before vdsm restart? If only current state matter during recovery, we don't need it.

michalskrivanek: In recovery we always just hope for the best, for normal disks too.
VM runs, it had that path before, presumably working(otherwise it would be dropped due to startupPolicy). If we failk in prepareVolume during recovery we ignore it...because what else can you do, it may in fact keep running fine if your activation error is bogus. Or not. But then again, what would you do...on VM start you fail to start, that's ok, but if it's already running you shouldn't be killing it.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: I'm not sure I follow. vdsm generate XML just for vdsm to have PDIV stored for the image. The issue is that vdsm activate the volume only when engine specifies volume using PDIV. If engine sends just path, vdsm assume it exists. So you propose to keep everything and just activate the volume in case of engine sends path? (IIRC we excluded this approach in previous discussions).

As for the API, yet, it's meant to be synchronous.

michalskrivanek: vdsm doesn't generate xml. engine generates all the xml except for some parts which we try to minimize. The reesulting path where that volume is mounted is defined by engine already ([here](https://github.com/oVirt/ovirt-engine/blob/8203fad23f1de20c6c800a2e4352464bb230ab3d/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java#L1176))
I don't think Arik meant to send only the path. You still need additional parameters in metadata, that's fine. It's just that the path doesn't have to be replaced/generated in vdsm.
So in mind the API should be similar to other hotplugs - a verb like "changeCD" and an xml updating the <disk>'s path and corresponding metadata changes (PDIV, other properties, whatever you need). Both ways for insert and eject
API - ack. that makes it simpler

vjuranek: so you mean replace e.g. [this](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/virt/vm.py#L4375) where we change path in VM XML by using XML will be sent by engine and on vdsm just make sure volume defined by sent PDIV is activated?

ahadas: to be concrete, here's an example of xml that the engine generates for hotplug-nic:
```xml
<?xml version="1.0" encoding="UTF-8"?><hotplug>
  <devices>
    <interface type="bridge">
      <model type="virtio"/>
      <link state="up"/>
      <source bridge="ovirtmgmt"/>
      <alias name="ua-15f4ba44-8e3b-4bbe-817a-f118cd2c88c7"/>
      <mac address="56:6f:a7:1a:00:0a"/>
      <mtu size="1500"/>
      <filterref filter="vdsm-no-mac-spoofing"/>
      <bandwidth/>
    </interface>
  </devices>
  <metadata xmlns:ovirt-vm="http://ovirt.org/vm/1.0">
    <ovirt-vm:vm>
      <ovirt-vm:device mac_address="56:6f:a7:1a:00:0a">
        <ovirt-vm:custom/>
      </ovirt-vm:device>
    </ovirt-vm:vm>
  </metadata>
</hotplug>
```
it's ok to have both `device` and `metadata` sections.
in case of change-cd we can set the PDIV details within the metadata section like we do in create-vm, for instance:
```xml
<ovirt-vm:device devtype="disk" name="vda">
   <ovirt-vm:poolID>542d569a-74a1-4847-9c14-08c20182e7f3</ovirt-vm:poolID>
   <ovirt-vm:volumeID>1d82ee44-884b-4681-9fa3-5dcf515c1e67</ovirt-vm:volumeID>
   <ovirt-vm:imageID>152440e6-f383-466f-9724-73b3877c4e2c</ovirt-vm:imageID>
   <ovirt-vm:domainID>dde147c1-60ec-4e99-9a6e-6a709975b65b</ovirt-vm:domainID>
</ovirt-vm:device>
```

nirs: I don't see how sending xml from engine helps. This just complicates everything and we don't do this for any storage API.

The main issue here is supporting block storage for cdrom.
- We need to always keep PDIV information in vm metadata, this is done now only when starting VM with a cdrom on block device
- We need to tear down the volume after ejecting a CD
- We need to prepare a volume before inserting a CD
- For handling PDIV, vdsm need to accept dict with the uuids. Accepting xml with the uuids does no make sense.
- We need new APIs to make the flow easier to manage
- For new APIs we need new cluster version to make it sane to manage on engine side
- Enigne cannot predict block volume path, what it does today is wrong and unsupported by vdsm. Volume path is internal implementation detail and any vdsm client that "predict" this is going to fail sooner or later. I warned about this many times.

source/develop/release-management/features/storage/cd-block-storage.md
nirs: The purpose of the in_progress flag is to allow cleanup during recovery. Without this we will have to check every drive spec and device to do the needed cleanup.

If removing it fails after cleaning up stale volumes, we will try agian in the next recovery, there is no hard.

source/develop/release-management/features/storage/cd-block-storage.md
nirs: What do you mean?

source/develop/release-management/features/storage/cd-block-storage.md
nirs: In vdsm API we use prepare and teardown. The underlyng term in lvm level is activation/deactivation.
In this document we better use prepare/teardown.

source/develop/release-management/features/storage/cd-block-storage.md
nirs: We cannot base anything on libvirt events since we may miss events during
restart.

If a remove operation finish after vdsm recover, the cd will remain removed with
active volume. To handle this we need support from libvirt, like having a block job
so we can detect an operation in progress in libvirt side.

To have this libvirt must support job id set by the caller. We will start a removal
process by storing the job id in the vm metadata. Then if vdsm recover during
the operation we can check the status of the job properly, and register a callback
for completing the cleanup when the job completes.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: Sorry for commenting on this without knowing all the details and please correct me as needed. A general approach since introducing Engine XML is using XML for device handling (although it must be amended or completed on the host side in some cases). Ideally, we should be consistent and use XML wherever possible, which is unfortunately not true yet. If we can provide a CD at VM start (I guess we need that at least when booting from the CD) in the form of XML (including metadata), it makes sense to provide it in the same form when changing the CD. Sure, extracting the info from the XML is a "useless" extra step on the Vdsm side (not sure about the Engine side) but we do that on VM startup anyway. When introducing a new API, which we don't want to change in the future, we should consider the general aim of Engine XML API consistency, even if it means a bit of extra work now. It may return the investment later, especially if we ever manage to cleanup storage device handling and make another new shiny motorbike from the old rusty one. On the Vdsm side, It's about replacing dictionary queries with XPath queries and replacing a complete XML output processing with selective DOM changes. Which shouldn't be *that* big deal. Let's consider which of the two options (XML x dictionary) will make future maintenance easier from the global point of view and whether using XML could help start fixing the current storage device handling mess or would make it even worse.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: I tend to agree with Vojta and Nir here. Using in_progress flag looks like making it more rather than less robust. We don't have such a flag on disk hot plug/unplug though, right?

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: OK, please clarify success/failure in the document.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: If a CD change is implemented as eject + insert then it is what the guest OS is used to. If a CD is changed in-place, without ejecting the old CD first, it can confuse the guest OS and lead to various problems. We had such issues in the past.

source/develop/release-management/features/storage/cd-block-storage.md
mz-pdm: I don't know how it works in libvirt and whether what I write above can actually happen. But it's probably worth to check and document possible problems if there can be any.

source/develop/release-management/features/storage/cd-block-storage.md
ahadas: yeah, and from the engine side it enables us to reuse the code that creates the cd-rom device on create-vm also for creating the device with the different CD - if one ever changes the properties of the cd-rom device on create-vm, it will auto-magically apply in the case of change-cd without any further changes in VDSM.

at the more conceptual level I wrote back then [a post on why we took that direction](https://ahadas.com/engine-xml/). it may look like an additional overhead now but as we're about to change the VDSM<->engine API, we can take this opportunity to better align it with other verbs in VDSM in which the engine XML is used.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: Based on https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainUpdateDeviceFlags the call returns 0 in case of success, -1 in case of failure. So my impression (don't know how to find it out for sure without reading the source code) is that this call is synchronous and scenarios outlined above cannot happen.

vjuranek: It seems there's more agreement to keep it. I clarified it little bit more in the new version and added example from hot disk plug, which doesn't have it and it can result into active unused volumes.

source/develop/release-management/features/storage/cd-block-storage.md
vjuranek: It seems there's no agreement on this as @nirs is strongly against it. Nevertheless we at least agree that we will add two new APIs into vdsm. We can discuss further once the initial patches for API will be post, what should it accecpt (xml vs. dict)

michalskrivanek: I'm not sure there is even agreement on 2 new APIs. @ahadas brought up (ok, maybe only in our conversation and not here) a point that we can easily use the same API verb just extended the dictionary with "xml" attribute, carrying the new xml representation with pdiv.

michalskrivanek: > I don't see how sending xml from engine helps. This just complicates everything and we don't do this for any storage API.

we do that vdsm for every aspect of VM configuration since 4.2. Yes, storage code is behind because the storage team decided to ignore that change. It doesn't complicate anything, it gets the storage code aligned with the rest of vdsm.

> The main issue here is supporting block storage for cdrom.

yes

>     * We need to always keep PDIV information in vm metadata, this is done now only when starting VM with a cdrom on block device

yes, no argument there. It should be up to date

>     * We need to tear down the volume after ejecting a CD

after ejecting from the last guest using the same CD

>     * We need to prepare a volume before inserting a CD

yes (supposing doing it twice in case it's already used by other guest is harmless) 

>     * For handling PDIV, vdsm need to accept dict with the uuids. Accepting xml with the uuids does no make sense.

xml makes it aligned with all the other VM APIs

>     * We need new APIs to make the flow easier to manage

there's no complexity in API on vdsm side, neither on engine side. Plus if we can easily keep using the same API I would prefer to keep compatibility and the code simpler

>     * For new APIs we need new cluster version to make it sane to manage on engine side

this is just a complexity introduced by new API. No new API == no need to do this.

>     * Enigne cannot predict block volume path, what it does today is wrong and unsupported by vdsm. Volume path is internal implementation detail and any vdsm client that "predict" this is going to fail sooner or later. I warned about this many times.

Engine is the only meaningful vdsm client, so for better or worse this is what is supported.
We can use placeholders to replace it with calculated path from metadata (like we use for serial number). If you'd like to close the xml refactoring finishing the storage subsystem transition this would be a good task to do...

mz-pdm: I agree, it looks like a good plan to me.

ahadas: +1

michalskrivanek: LGTM

source/develop/release-management/features/storage/cd-block-storage.md
ahadas: @vjuranek I'd suggest to give it a second thought - as @nirs  pointed out in his last comment, it seems to be possible to achieve that without this flag, right? sure, it may complicate the recovery flow but considering that execute the recovery flow significantly less times than run-vm, I'd say that if we can avoid the extra modification of the metadata on run vm (with cd-rom), it would be the better approach here

ahadas (commented):
@vjuranek that's not the new version you've mentioned in the comments, right?

source/develop/release-management/features/storage/cd-block-storage.md
nirs: > After some more offline discussion, can we agree on following @nirs @michalskrivanek @ahadas @mz-pdm ?
> 
> * engine will send XML

Maybe, we will discuss these details when we work on the implementation.

> * virt part of vdsm will accept XML
> * virt part of vdsm will create a dict with PDIV from XML and will pass it to storage part
> * if there is no XML in the request, path will be used (i.e. we will fall back to old behavior)
> * fall back to old behavior described in previous item means we don't need new vdsm API and can reuse the old one without breaking backward compatibility

No, we are not going to do these kinds of hacks. We need clean API, no hacks.

> If you agree on these items, I'll update the doc appropriately.

I strongly disagree with this plan.

nirs (approved):
Thanks for the feedback, we will go with this design.

source/develop/release-management/features/storage/cd-block-storage.md
michalskrivanek: if there is a disagreement even at this phase then it should be resolved, not dismissed...
it won't help to get to a patch stage and get a -2 there.

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+However, non of these functions worked properly with block based storage.
+Changing CD failed on block storage as the appropriate volume wasn't activated.
+Ejecting CD failed to deactivate the volume.
+For more details, see following bugs.
+
+#### CD-ROM related XML
+
+Currently each VM contains CD-ROM (without CD), e.g.
+
+```
```
sandrobonazzola, line 36: please use ```xml

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+      <address type='drive' controller='0' bus='0' target='0' unit='2'/>
+    </disk>
+```
+
+so that it contains path to the ISO.
+
+Besides this libvirt XML, oVirt adds its own XML for each disk to be able to identify
+the volume by PDIV (pool, domain, image volume) coordinates and includes other useful info like lease path. e.g.
+
+```
```
sandrobonazzola, line 68: please use ```xml

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+                <ovirt-vm:path>/rhev/data-center/mnt/blockSD/88252cf6-381e-48f0-8795-a294a32c7149/images/89f05c7d-b961-4935-993f-514499024515/626a493f-5214-4337-b580-96a1ce702c2a</ovirt-vm:path>
+                <ovirt-vm:volumeID>626a493f-5214-4337-b580-96a1ce702c2a</ovirt-vm:volumeID>
+            </ovirt-vm:volumeChainNode>
+        </ovirt-vm:volumeChain>
+    </ovirt-vm:device>
+```
+
+When CD is not loaded, this element is still present (as CD-ROM is always present), but is empty, e.g.:
+
+```
```
sandrobonazzola, line 89: please use ```xml

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+      1. fail the flow with the original error
+1. remove `state` attribute from VM metadata
+  - if failed: log the error and succeed
+
+
+#### VM with loaded CD, eject CD
+
+1. engine sends `changeCD` command with empty PDIV format.
+1. vdsm updates VM metadata with new attribute `state="ejecting"`:
+```
```
sandrobonazzola, line 144: please use ```xml

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+  - if error is anything else except device in use error, remove `state` attribute from metadata, re-throw the exception and fail
+1. remove drive spec from VM metadata
+  - if failed: log error and succeed
+
+
+#### VM with loaded CD, changeCD
+
+1. engine sends `changeCD`, specifying new ISO path by PDIV format and also path.
+1. update exiting drive spec for loaded CD with `state="ejecting"`:
+```
```
sandrobonazzola, line 162: please use ```xml

source/develop/release-management/features/storage/cd-block-storage.md
```diff
@@ -0,0 +1,309 @@
...
+    <ovirt-vm:device devtype="disk" name="sdc" state="ejecting">
+```
+1. add another drive spec to VM metadata for CD being loaded with attribute `state="loading"`
+```
```
sandrobonazzola, line 166: please use ```xml

sandrobonazzola:
Anything else to be reviewed or can we merge?

vjuranek:
> Anything else to be reviewed or can we merge?

this was re-worked based on our meeting and needs another round of reviews

michalskrivanek (approved):

PR #2401, opened 2020-12-01
vdsm: Update git url

nirs:
Update git clone example to the recommended URL.

### DIFF
--- a/source/develop/developer-guide/vdsm/developers.md
+++ b/source/develop/developer-guide/vdsm/developers.md
@@ -48,7 +48,7 @@ Our public git repository is located at: [oVirt.org](http://gerrit.ovirt.org/git
 
 You can clone this repository by running the following command:
 
-      git clone http://gerrit.ovirt.org/p/vdsm.git
+      git clone "https://gerrit.ovirt.org/vdsm"
 
 ## Installing the required packages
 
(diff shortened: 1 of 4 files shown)


source/develop/developer-guide/vdsm/developers.md
```diff
@@ -332,15 +332,15 @@ There are several ways to update your vds_bootstrap, for example:
...
+        # git clone "https://gerrit.ovirt.org/vdsm"
         # cd vdsm
         (if required increase the Version/Release from vdsm.spec.in)
         # ./autogen.sh --system && make && make rpm
         # rpm -Uvh ~/rpmbuild/RPMS/noarch/vdsm-bootstrap* 
 
 *   Clone the vdsm tree, execute autogen/make and copy the vds_bootstrap/vds_bootstrap.py to /usr/share/vds_bootstrap
 
-        # git clone git://gerrit.ovirt.org/vdsm
+        # git clone "https://gerrit.ovirt.org/vdsm"
```
didib, line 343: Did you deliberately not update lines 462, 508 below (not sure how to comment on them directly)?

I see you only update vdsm (as your commit message clearly says). Perhaps open an issue to update all git links, not just vdsm.

I think the reason you didn't notice the other lines is because 'git' and 'clone', in the .md files (but not generated html), are separated, instead of with a space, with NO-BREAK SPACE aka U+00A0. Perhaps open another issue about replacing all of these with spaces (I see tons of them all over the place).

nirs: No, I missed them, thanks for noticing. Good idea to open an issue about nbsp.

didib: Now filed #2412 .

PR #2429, opened 2021-01-25
Add feature page for sriov live migration

almusil:

- Add feature page for SR-IOV live migration

### DIFF
--- a/source/develop/release-management/features/network/sriov-live-migration.md
+++ b/source/develop/release-management/features/network/sriov-live-migration.md
@@ -1,0 +1,187 @@
+---
+title: SR-IOV Live Migration without downtime
+category: feature
+authors: amusil
+feature_name: SR-IOV Live Migration without downtime
+feature_modules: engine,network,vdsm
+feature_status: In Development
+---
+
+# SR-IOV Live Migration without downtime
+
+## Summary
+
+This feature allows you to migrate VM that is connected via SR-IOV with
+a minimal network downtime and supersedes the [Live Migration Support For SRIOV][1]
+that required additional configuration inside affected VM.
+
+The state of the feature is tracked in [[RFE] [SR-IOV] Migration should not require downtime as of today][2].
+Related patches can be found on [gerrit topic:sriov_live_migration][3].
+
+## Owner
+
+*   Name: [Ales Musil](https://github.com/almusil)
+
+*   Email: <amusil@redhat.com>
+
+
+## Detailed Description
+
+In order to migrate a VM that is connected to SR-IOV passthrough profile
+we need to first unplug the SR-IOV device from a VM on the source, migrate
+the VM and plug SR-IOV device on the destination. This causes longer downtime
+during which the VM is not connected to the network.
+
+The current implementation allowed to mitigate this downtime
+[Live Migration Support For SRIOV][1], however this approach required manual
+configuration to be done in the guest which is not convenient for large setups.
+
+In order to achieve minimal downtime the [NET_FAILOVER][4] support was added
+to Linux kernel. The driver provides automated mechanism which is transparent
+to the VM, and the user space. It is achieved by `failover` netdev which acts
+as master device and controls two slave devices, the `primary` device being
+SR-IOV passtrough and the virtio as `standby`. The VM user access network
+through the `failover` device.
+
+There is still need to provide two interfaces, the SR-IOV passtrough device
+and virtio device, to the affected VM, but the  overall configuration overhead
+has been reduced to selecting suitable vNIC profile pair.
+
+
+## Prerequisites
+
+The implementation requires a [Linux kernel][5], [libvirt][6] and [qemu][7]
+support with `net_failover`.
+
+The [Linux kernel][5] support has to be only on guest side.
+
+## Limitations
+
+* In addition to the SR-IOV NIC, a linux-bridge based has to be connected to the source, and the destination host.
+* Because of the required support in the guest, only RHEL8 (Linux kernel version >=4.18 or >=5.10) and Windows
+  (Windows VirtIO Drivers >=0.1.189) guests are supported.
+* Hosts with Cluster level version of >=4.6 are required.
+
+## Implementation
+
+### Entity Description
+
+vNIC profile will get a new attribute that will point to the failover virtio
+vNIC profile. This selection will be only available with `Passthrough` and `Migratable`
+selected.
+
+### User Experience
+
+Upon creation of vNIC profile, the user can enable failover and select
+the failover network vNIC profile. The `Failover` selection is available
+only for `Passthrough` and `Migratable` vNIC profiles.
+
+REST API takes an additional parameter for vNIC called `failover`
+containing id of the failover vNIC profile:
+
+```
+<vnic_profile id="123">
+  <name>vNIC_name</name>
+  <pass_through>
+    <mode>enabled</mode>
+  </pass_through>
+  <migratable>true</migratable>
+  <failover id="456"/>
+</vnic_profile>
+```
+
+VM interface then requires only single vNIC profile that has the `failover` attribute
+in order to utilize this behavior.
+
+### Engine backend
+
+As engine is responsible for creating XML for a new VM nic, the
+appropriate XML will be created and provided to libvirt.
+
+In order to ensure that the virtio network is available the migration policy needs to be updated.
+
+The `ActivateDeactivateVmNicCommand` will be updated to handle plug/unplug of the failover nic, when needed.
+During migration the command will ensure that only `VF` VM nic is unplugged and then plugged back, any other
+user interaction will affect both `VF` and `failover` VM nic.
+
+As mentioned in the [Limitations](#limitations), new cluster level is required to ensure correct package version
+of libvirt and qemu.
+
+### Libvirt XML
+
+If the feature is activated for a new VM nic, the libvirt XML definition
+contains an additional network with attribute [teaming element][8],
+as follows:
+
+```
+<devices>
+  <interface type='network'>
+    <source network='mybridge'/>
+    <mac address='00:11:22:33:44:55'/>
+    <model type='virtio'/>
+    <teaming type='persistent'/>
+    <alias name='ua-backup0'/>
+  </interface>
+  <interface type='network'>
+    <source network='hostdev-pool'/>
+    <mac address='00:11:22:33:44:55'/>
+    <model type='virtio'/>
+    <teaming type='transient' persistent='ua-backup0'/>
+  </interface>
+</devices>
+```
+
+### Vdsm
+
+Because of the current rule that `VF` and `failover` have same MAC address, vdsm code needs to adjusted
+to account for that. In order to achieve that, vdsm will require matching of alias and MAC address if the
+interface xml contains `teaming` attribute.
+
+## Installation/Upgrade
+
+The feature will not affect new or existing installations.
+
+## Testing
+
+To test this feature, the desired VM with SR-IOV failover
+configured, should be migrated to a different host.
+
+## Open qeustions
+
+* An allowed limit of packet drop during migration?
+* Can the virtio device be connected to the same PF as VF?
+
+
+## Documentation & External references
+
+[Live Migration Support For SRIOV][1]
+
+[1]: https://www.ovirt.org/develop/release-management/features/network/liveMigrationSupportForSRIOV.html
+
+[Bug 1688177 - [RFE] [SR-IOV] Migration should not require downtime as of today][2]
+
+[2]: https://bugzilla.redhat.com/1688177
+
+[gerrit topic:sriov_live_migration][3]
+
+[3]: https://gerrit.ovirt.org/#/q/topic:sriov_live_migration
+
+[Linux: NET_FAILOVER][4]
+
+[4]: https://www.kernel.org/doc/html/latest/networking/net_failover.html
+
+[Linux: net: Introduce generic failover module][5]
+
+[5]: https://github.com/torvalds/linux/commit/30c8bd5aa8b2c78546c3e52337101b9c85879320
+
+[qemu: net/virtio: add failover support][6]
+
+[6]: https://github.com/patchew-project/qemu/commit/9711cd0dfc3fa414f7f64935713c07134ae67971
+
+[Bug 1693587 - RFE: support for net failover devices in libvirt][7]
+
+[7]: https://bugzilla.redhat.com/1693587
+
+[Teaming a virtio/hostdev NIC pair][8]
+
+[8]: https://libvirt.org/formatdomain.html#teaming-a-virtio-hostdev-nic-pair


source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+
+Vm nic will get a new attribute that will point to the failover virtio 
+vnic profile. 
+
+### User Experience
+
+Upon creation of new VM nic, the user can enable failover and select 
+the failover network vnic profile. The `failover` selection is available 
+only for passthrough master vnic profiles. 
+
```
almusil, line 75: After this section there should be image how the menu looks.

almusil:
@dominikholler @mburman5 Please take a look

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+## Limitations
+
+* The virtio network has to be required to ensure connectivity on the source,
+  and the destination hosts.
+
+## Implementation
+
+### Entity Description
+
+Vm nic will get a new attribute that will point to the failover virtio
```
dominikholler, line 67: Is there a reason to have the attribute on the VM NIC, instead of having the new attribute on the vNIC profile?
If the attribute would be on the vNIC profile, it could be reused by multiple NICs.

dominikholler: What will be the relation to the "Migratable" attribute?

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+### Vdsm
+
+Handling of update and device removal is handled in vdsm. In both cases we 
+can actually apply changes that are relevant to both interface as those 
+interfaces are linked by the `teaming` attribute. 
+
+Relevant update changes are:
+* Nic plug/unplug
```
dominikholler, line 126: vNIC plug should not require a code change in VDSM.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+* Nic being linked
```
dominikholler, line 127: I think link/unlink is not supported for SR-IOV.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Installation/Upgrade
+
+The feature will not affect new or existing installations.
+
+## Testing
+
+To test this feature, the desired VM with SR-IOV failover 
+configured, should be migrated to a different host.
+
```
dominikholler, line 138: Can you please add more details, e.g. ping with a given interval and maximum allowed dropped ICMP messages?

dominikholler (changes requested):
Maybe I will have more comments in the future.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+In order to migrate a VM that is connected to SR-IOV passthrough profile
+we need to first unplug the SR-IOV device from a VM on the source, migrate
+the VM and plug SR-IOV device on the destination. This causes longer downtime
+during which the VM is not connected to the network.
+
+The current implementation allowed to mitigate this downtime
+[Live Migration Support For SRIOV][1], however this approach required manual
+configuration to be done on a VM which is not convenient for large setups. 
+
+In order to achieve minimal downtime the [NET_FAILOVER][4] support was added
```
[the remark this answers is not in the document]
almusil: Why it should point to the patch instead of the documentation?

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+
+The implementation requires a [Linux kernel][5], [libvirt][6] and [qemu][7] 
+support with `net_failover`. 
+
+The [Linux kernel][5] support has to be only on guest side. 
+
+## Limitations
+
+* The virtio network has to be required to ensure connectivity on the source,
+  and the destination hosts.
```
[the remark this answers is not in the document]
almusil: Isn't the second one said in the requirements? It does not have to be necessary RHEL 8 but any kernel with that supports it or am I missing something?

source/develop/release-management/features/network/sriov-live-migration.md
almusil: That might be actually a better approach.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Makes sense but it is confusing because technically it is edit of existing interface

source/develop/release-management/features/network/sriov-live-migration.md
almusil: That might be actually dependent on the hardware isn't it? I'm frankly not sure what are the requirements for "low downtime" and the RFE is not helping either, maybe we need to clarify that term first before we put any hard requirement to the feature page.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+title: SR-IOV Live Migration
+category: feature
+authors: amusil
+feature_name: SR-IOV Live Migration
+feature_modules: engine,network,vdsm
+feature_status: In Development
+---
+
+# SR-IOV Live Migration
+
```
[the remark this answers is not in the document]
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+## Detailed Description
+
+In order to migrate a VM that is connected to SR-IOV passthrough profile
+we need to first unplug the SR-IOV device from a VM on the source, migrate
+the VM and plug SR-IOV device on the destination. This causes longer downtime
+during which the VM is not connected to the network.
+
+The current implementation allowed to mitigate this downtime
+[Live Migration Support For SRIOV][1], however this approach required manual
+configuration to be done on a VM which is not convenient for large setups.
```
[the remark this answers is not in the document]
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,172 @@
...
+In order to achieve minimal downtime the [NET_FAILOVER][4] support was added 
+to Linux kernel. The driver provides automated mechanism which is transparent 
+to the VM, and the user space. It is achieved by `failover` netdev which acts
+as master device and controls two slave devices, the `primary` device being 
+SR-IOV passtrough and the virtio as `standby`. The VM user access network
+through the `failover` device. 
+
+There is still need to provide two interfaces, the SR-IOV passtrough device
+and virtio device, to the affected VM, but the  overall configuration overhead
+has been reduced to selecting suitable vnic profile pair.
```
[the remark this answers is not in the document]
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Right

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+feature_modules: engine,network,vdsm
+feature_status: In Development
+---
+
+# SR-IOV Live Migration without downtime
+
+## Summary
+
+This feature allows you to migrate VM that is connected via SR-IOV with
+minimal network downtime and supersedes the [Live Migration Support For SRIOV][1]
```
mburman5, line 15: with a minimal network

source/develop/release-management/features/network/sriov-live-migration.md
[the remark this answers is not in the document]
mburman5: For the first bullet Dominik wrote: If on the destination host the linux bridge network is not attached, then the VM migration shouldn't  even be scheduled by the engine. Like if it missing VF on the destination host, right? 
Also, does it matter on which NIC/port the linux bridged will be attached, on the parent PF or on any other physical interface that has connectivity?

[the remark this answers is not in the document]
almusil: That should be ensured by the network being required. I think the second point is on administrator. We should not enforce rules here IMO as the admin should know best the impact.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+* In addition to the SR-IOV NIC, a linux-bridge based (or an OVN physical networkin OVS clusters)
+  has to be connected to the the source and the destination host.
+
+## Implementation
+
+### Entity Description
+
+vNIC profile will get a new attribute that will point to the failover virtio 
+vNIC profile. This selection will be only available with `Passtrough` and `Migratable`
+selected.
```
mburman5, line 72: Sounds good and correct

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+</vnic_profile>
+```
+
+### Engine backend
+
+As engine is responsible for creating XML for a new VM nic, the 
+appropriate XML will be created and provided to Libvirt. 
+
+In order to ensure that the virtio network is available it has to be
+marked as required, and the migration policy needs to be updated.
```
mburman5, line 100: Yes, the migration shouldn't start if there is no network on the destination host

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: So user will need to add 2 vNICs to the guest VM right, one virtio vNIC and one passthrough vNIC.
Can we make it clear?

almusil: No quite the opposite. The vnic profile that will have the failover selected needs to be added to the VM as single interface and that is enough. But good point that this might be confusing I will make it clear in the feature page.

source/develop/release-management/features/network/sriov-live-migration.md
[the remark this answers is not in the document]
mburman5: Hmm, i think it should be like no free VF on destination host, the migration not started.
I think it's better than having the admin start the migration and fail with error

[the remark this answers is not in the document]
almusil: Right but this policy is implied by the migratable sriov and this feature won't modify it.

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: Yes please. 
So it means only one sr-iov with failover vNIC should be added to the VM, yes?

almusil: Exactly, to engine it is single interface.

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: I think that in our test we will need to limit and allow some packet loss, or not? these are good questions actually..

almusil: Yes I agree, but first we should define the requirements and then built those parameters around that.

source/develop/release-management/features/network/sriov-live-migration.md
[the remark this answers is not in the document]
mburman5: I prefer vNIC or VM's vNIC

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: On the vNIC profile is much better. It has already migratable and passthrogugh.

almusil: Agree, it makes much more sense

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+The implementation requires a [Linux kernel][5], [libvirt][6] and [qemu][7] 
+support with `net_failover`. 
+
+The [Linux kernel][5] support has to be only on guest side. 
+
+## Limitations
+
+* In addition to the SR-IOV NIC, a linux-bridge based (or an OVN physical network in OVS clusters)
+  has to be connected to the source, and the destination host, marked as required.
+
```
dominikholler, line 62: Because of the required support in the guest, only RHEL8 guests are supported.

- If we don't know yet which kernel version is required, we should highlight this here.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+in order to utilize this behavior.
+
+### Engine backend
+
+As engine is responsible for creating XML for a new VM nic, the 
+appropriate XML will be created and provided to Libvirt. 
+
+In order to ensure that the virtio network is available it has to be
+marked as required, and the migration policy needs to be updated.
+
```
dominikholler, line 101: Do we want to ensure that the guest supports the feature? What should or might happen, if the guest does not support the feature?

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+to Linux kernel. The driver provides automated mechanism which is transparent 
+to the VM, and the user space. It is achieved by `failover` netdev which acts
+as master device and controls two slave devices, the `primary` device being 
+SR-IOV passtrough and the virtio as `standby`. The VM user access network
+through the `failover` device. 
+
+There is still need to provide two interfaces, the SR-IOV passtrough device
+and virtio device, to the affected VM, but the  overall configuration overhead
+has been reduced to selecting suitable vNIC profile pair.
+
```
dominikholler, line 49: Do we already know, if the virtio device can be connected to the same physical NIC as the VF?

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+### Vdsm
+
+Handling of update and device removal is handled in vdsm. In both cases we 
+can actually apply changes that are relevant to both interface as those 
+interfaces are linked by the `teaming` attribute. 
+
+Relevant update changes are:
+* Nic unplug
+* MAC address change 
+
```
dominikholler, line 135: @erav Can you please have a look? You might be deep in the details here.

dominikholler (changes requested):

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: Good question :)

almusil: Engine knows kernel version of quest, but IMO I would not put any restriction there. Even if the kernel version has the module it might be disabled/compiled without it.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,179 @@
...
+* In addition to the SR-IOV NIC, a linux-bridge based (or an OVN physical network in OVS clusters)
+  has to be connected to the source, and the destination host, marked as required.
+* Because of the required support in the guest, only RHEL8 guests are supported (Linux kernel version >=4.18 or >=5.10).
+
+## Implementation
+
+### Entity Description
+
+vNIC profile will get a new attribute that will point to the failover virtio 
+vNIC profile. This selection will be only available with `Passtrough` and `Migratable`
```
erav, line 69: nit: Passthrough

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,179 @@
...
+selected. 
+
+### User Experience
+
+Upon creation of vNIC profile, the user can enable failover and select 
+the failover network vNIC profile. The `Failover` selection is available 
+only for `Passtroguh` and `Migratable` vNIC profiles.
```
erav, line 76: nit: Passthrough

source/develop/release-management/features/network/sriov-live-migration.md
dominikholler: Maybe the backup interface would be invisible or ignored by the guest? If we don't know this yet, we should note this as an open question which we have to check.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+  <migratable>true</migratable>
+  <failover id="456"/>
+</vnic_profile>
+```
+
+VM interface then requires only single vNIC profile that has the `failover` attribute 
+in order to utilize this behavior.
+
+### Engine backend
+
```
dominikholler, line 96: - I understood that the live-migration code has to be modified, to skip the unplug of the SR-IOV NIC. If this, this should be mentioned here.
- I also understood, that we should have a new cluster level to ensure that all hosts in the cluster have the required libvirt and qemu versions.

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+There is still need to provide two interfaces, the SR-IOV passtrough device
+and virtio device, to the affected VM, but the  overall configuration overhead
+has been reduced to selecting suitable vNIC profile pair.
+
+
+## Prerequisites
+
+The implementation requires a [Linux kernel][5], [libvirt][6] and [qemu][7] 
+support with `net_failover`. 
+
```
dominikholler, line 55: Which versions of these components will be required? Which RHEL version will contain them?

source/develop/release-management/features/network/sriov-live-migration.md
```diff
@@ -0,0 +1,178 @@
...
+The [Linux kernel][5] support has to be only on guest side. 
+
+## Limitations
+
+* In addition to the SR-IOV NIC, a linux-bridge based (or an OVN physical network in OVS clusters)
+  has to be connected to the source, and the destination host.
+* Because of the required support in the guest, only RHEL8 guests are supported (Linux kernel version >=4.18 or >=5.10).
+
```
dominikholler, line 63: Can you please check if there will be a new virtio driver version for Windows and Windows will be supported by Red Hat, too? 
https://github.com/virtio-win/kvm-guest-drivers-windows/pull/472

source/develop/release-management/features/network/sriov-live-migration.md
dominikholler: Maybe we can conserve this as an open question?

dominikholler (changes requested):

source/develop/release-management/features/network/sriov-live-migration.md
mburman5: Will this break migration implementation and logic that we currently have? will this mean that the Vf won't get hotunplugged from the source host?

mburman5: Also, about the required network. We agreed that making the fail-over network as required, is not the right direction and might introduce issues.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: This is more or less open question because we don't know that yet also wouldn't be enough to state that 4.6 cluster is needed?

source/develop/release-management/features/network/sriov-live-migration.md
almusil: The required statement is not there anymore.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: No guest will see both interface no matter if the failover module is there or not. But only in case of the kernel support it will create the teaming interface automatially.

source/develop/release-management/features/network/sriov-live-migration.md
dominikholler: You could write that RHEL 8.3 would be not enough with an explaination , and mention the new cluster level in the section about the Engine backend.

source/develop/release-management/features/network/sriov-live-migration.md
almusil: There is a new section called open questions, which can be updated as we find the answers

source/develop/release-management/features/network/sriov-live-migration.md
almusil: I have mentioned it to the limitations it seems more fitting

source/develop/release-management/features/network/sriov-live-migration.md
almusil: Done

source/develop/release-management/features/network/sriov-live-migration.md
almusil: I have mentioned the skip of unplug and the cluster level version

dominikholler (approved):
Looks good

source/develop/release-management/features/network/sriov-live-migration.md
erav: * Nic unplug:
Is the failover attribute immutable after the profile creation?
if not is toggling it via a vnic profile update applied immediately to the device or is a vnic replug required? this might have an implication to the out-of-sync indication.
* MAC update:
IIUC there is a single vNIC profile with a single MAC address for both devices. So from the point of view of engine's MAC pool infrastructure this setup should not be different than any other vnic profile. if there are duplicate addresses in the pool, the migration might run into collisions regardless of the failover feature.

almusil: >     * Nic unplug:
>       Is the failover attribute immutable after the profile creation?
>       If not, is toggling it via a vnic profile update applied immediately to the device or is a vnic replug required? this might have an implication to the out-of-sync indication.

It will be editable, and does not have immediate effect. So this should be also part of the out-of-sync flag. 

> 
>     * MAC update:
>       IIUC there is a single vNIC profile with a single MAC address for both devices. So from the point of view of engine's MAC pool infrastructure this setup should not be different than any other vnic profile. if there are duplicate addresses assigned from the pool, the migration might run into collisions regardless of the failover feature.

Yes, the nic behaves the same in regards of MAC addresses only under the hood there are two nics with same MAC.

almusil:
@dominikholler @mburman5 
I have updated the feature page with recent changes, if you could please take a look

dominikholler (approved):
Looks good, also in combination with https://gerrit.ovirt.org/c/ovirt-engine/+/113853/1 and below.

mwperina (approved):
Looks good to me

PR #2458, opened 2021-04-06
bundle update

sandrobonazzola:

- Updating the bundle.

- Added kramdown-parser-gfm as required by newer Jekyll version.


Gemfile.lock
```diff
@@ -172,9 +174,12 @@ DEPENDENCIES
...
   jemoji
   kramdown
+  kramdown-parser-gfm
   liquid
   listen
 
 RUBY VERSION
    ruby 2.5.5p157
 
+BUNDLED WITH
```
duck-rh, line 184: please remove these two lines, they only cause problems when preparing updates ona different system than the target one

duck-rh (changes requested):
The builder is currently using F30 and all such changes to deps need to be tested on this system and not on your personal system to avoid breaking production. I did the test and it works fine.

Once the small requested change is done you can merge.

Gemfile.lock
sandrobonazzola: Done

PR #2459, opened 2021-04-07
Switch to GH actions and test on prod system

duck-rh:
The goal is to be more integrated. The CI currently is associated with a
specific user account on travis-ci.org but it's better for the team to
have more control over it (or in case the person leaves the project).

I also made changes to test the website on the same OS used by the
production builder. This way changes to the Gemfile can be tested by the
team directly.


duck-rh:
@sandrobonazzola could you have a look please?

.github/workflows/website_tests.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo '/!\ nonbreaking spaces have been found within the markdown (see previous log entries)'
+            exit 2
+          fi
+
+      - name: "Install Podman"
+        run: sudo apt-get install -y podman
+
+      - name: "Create the Website Build Environment container"
+        # Duck: DO NOT change the target OS, it needs to be in sync with the production builder
+        run: podman run -dt --name ovirt-web-tests --volume .:/tmp/website registry.fedoraproject.org/fedora:30
```
sandrobonazzola, line 51: Fedora 30 reached EOL on 2020-05-26 ; any chances that we can move to more recent Fedora?

sandrobonazzola (approved):
Looks good to me to my limited knowledge fo how GH actions works. Just a comment on using fc30 but not blocking this patch on it.

janoszen:
@sandrobonazzola the GH actions part is sane too.

sandrobonazzola:
Looks like it still tries to run travis, that part should be disabled i guess

janoszen:
@sandrobonazzola at this time GH actions only builds the website, it doesn't push it I believe. @duck-rh ?

.github/workflows/website_tests.yml
duck-rh: Yes I wanted to upgrade too but did not get around to do it. Now we can bump the version in a PR and see if it would work before reinstalling the builder.

duck-rh:
> @sandrobonazzola at this time GH actions only builds the website, it doesn't push it I believe. @duck-rh ?

No it does not. Since Travis is not our infra we did not want it to touch production. There's a separate server doing the job. I think it could be a good fit for OpenShift in the future.

duck-rh:
> Looks like it still tries to run travis, that part should be disabled i guess

Yes, I did not want to break other PRs' workflow. I'm going to disable it right after merging.

duck-rh:
Thanks for the reviews.

PR #2472, opened 2021-04-28
handle deprecation notice for bundler options

sandrobonazzola:

- Solves:
```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across
bundler invocations, which bundler will no longer do in future versions.
Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
```

### DIFF
--- a/run-server.sh
+++ b/run-server.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -e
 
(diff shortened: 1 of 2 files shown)


duck-rh:
@sandrobonazzola sorry, I was away on the Golden Week, LGTM

PR #2507, opened 2021-05-26
Developers welcoming

sandrobonazzola:

- Making developer section more welcoming

This pull request needs review by: @oVirt/ovirt-site  @mwperina @michalskrivanek @ahadas @shenitzky @janoszen @stoobie


sandrobonazzola:
For reviewers, probably easier to just look at https://github.com/oVirt/ovirt-site/pull/2507/commits/abc9ca1040f1789e608a6b5b7ed037d2df29ccb0 ; the index.md file was moved to devdocs.md and the PR diff is not the best way to see the new doc.

ahadas:
Merged quickly I see :)
Looks good but do we really prefer bugs to be filed on an "empty" team? wouldn't it increase the chance it would be missed?

sandrobonazzola:
@ahadas we don't prefer but better to get the bug opened rather than not jut because of the ovirt team field :-) I set a bugzilla whine notifying me 3 times a day if a bug is left without a team assigned.Let's see how it goes for a while.

PR #2510, opened 2021-06-01
Add feature page for dedicated CPU policies

nyoxi:
Signed-off-by: Tomáš Golembiovský <tgolembi@redhat.com>

- This PR introduces a feature page for dedicated CPU policies

This pull request needs review by: @ahadas @mz-pdm @liranr23 @smelamud @ljelinkova

### DIFF
--- a/source/develop/release-management/features/virt/dedicated-cpu.md
+++ b/source/develop/release-management/features/virt/dedicated-cpu.md
@@ -1,0 +1,211 @@
+---
+title: Dedicated CPU Policies
+category: feature
+authors: tgolembi
+---
+
+# Dedicated CPU Policies
+
+## Summary
+
+This feature introduces policies for dynamically dedicating physical CPUs to
+the VM. While this can be (to some extent) done with static CPU pinning, the
+CPU pinning requires that the VM is pinned also to a particular host(s). It
+also ideally requires that no other VMs without pinning are running on the host
+to avoid sharing the pinned CPUs. Instead, this new feature allows the user to
+configure only the way how its physical CPUs should be assigned to the VM and oVirt
+will take care of dynamically allocating suitable physical CPUs as well as
+forbidding other VMs from running on dedicated CPUs.
+
+
+## Owner
+
+* Name: Tomáš Golembiovský
+* Email: tgolembi@redhat.com
+
+## Glossary
+
+  * [*CPU pinning*](/develop/sla/cpu-pinning.html):
+    Feature in oVirt that allows you to statically pin vCPUs to pCPUs by
+    defining a pinning string in VM configuration (e.g. `0#1_1#2-3`).
+  * [*NUMA auto pinning policy*](https://bugzilla.redhat.com/show_bug.cgi?id=1688186):
+    Feature in oVirt that defines pinning policies (*Pin* and *Resize and
+    Pin*) that allow CPU pinning to automatically adapt to the topology of
+    pinned host. Just like CPU pinning, this is performed once during VM
+    configuration.
+  * *pCPU*: Physical CPU of a host.
+  * *vCPU*: Virtual CPU of a VM.
+
+
+## Description
+
+CPU pinning and NUMA auto pinning policies don't take into account current
+utilization of a host. Instead, it is a responsibility of the administrator to
+make sure that VMs running on a host are not pinned to same pCPUs and ideally
+also to make sure that no VMs without CPU pinning are running on the host. This
+makes the pinning process rather difficult and inflexible.
+
+To be able to have more flexibility in configuring how vCPUs of VMs use pCPUs
+this feature introduces several policies for automatic dedicating of CPUs.
+This feature can be used instead of the manual CPU pinning methods (not in
+addition). Compared to CPU pinning the advantages are that:
+
+  * VM does not need to be pinned to a specific host(s),
+  * takes into consideration existing VMs which solves one of the issues with
+    *Pin* NUMA auto pinning policy where several VMs can end up to be pinned
+    to same CPUs,
+  * it makes sure that VMs without any dedicated CPUs don't share CPUs with
+    VMs that have dedicated CPUs
+
+The policies described below divide the set of available pCPUs into
+three sub-sets. There are the dedicated CPUs where each of those can be
+assigned only to a single VM. Then there are some restricted CPUs (threads)
+that cannot be used by any
+other VM because of requirements on the threads of the core. And
+finally there is the set of CPUs that are not strictly assigned to any
+particular VM and are shared between all the VMs with *shared* policy. This
+subset is hereby referred to as *shared pool*. Host CPUs that play a special
+role, e.g. CPU that VDSM is pinned to or CPUs reserved to host on SAP HANA
+systems are considered to be also part of the *shared pool*.
+
+The available CPU policies are:
+
+  * *shared*: *(default)* The guest vCPUs will be allowed to roam among all
+    pCPUs from shared pool.
+  * *dedicated*: The guest vCPUs will be strictly pinned to a set of host pCPUs
+    (similarly to static CPU pinning). The set of pCPUs will be chosen to match
+    the required guest CPU topology. If the host has SMT architecture, thread
+    siblings are preferred.
+  * *isolate-threads*: If the host does not have SMT architecture this is same
+    as *dedicated* policy. If the host has SMT architecture each vCPU is placed
+    on a different physical core and all thread siblings are excluded from use
+    by this and any other VM (rendering all but one thread from a core unusable).
+  * *siblings*: The host must have an SMT architecture. Each vCPU is
+    allocated on thread siblings and threads from single core can only be
+    assigned to this VM. If some threads remain unused these will be excluded
+    from use by any other VM.
+
+
+## Prerequisites
+
+There are no special prerequisites for this feature.
+
+
+## Security
+
+This feature is primarily designed for performance tuning and is not meant to
+be a security feature. It is possible for other VMs to share the dedicated CPUs
+for a short period of time when the VM is brought up (boot, after migration,
+restore from snapshots, etc.). This is not so much problem during boot because
+by the time OS is started the CPUs should be already evicted, but it could
+potentially provide a window of opportunity (for exploiting CPU
+vulnerabilities) after finished migration or when restoring VM from snapshot.
+
+
+## Limitations
+
+See the *Security* section.
+
+Currently, when a CPU is already dedicated to some VM, VDSM will not move the
+VM onto another CPU to make it free to another VM. This may be addressed in
+the future (see *Future Work* section), but for now this means:
+
+  * It may be impossible to start VMs with CPU pinning on that host because
+    the requested CPUs are already occupied. It is suggested to start VMs with
+    CPU pinning before any VMs with dedicated CPU policy.
+
+  * It may be impossible to start a VM with dedicated CPUs even though there
+    seems to be enough space in terms of number of used vs. free CPUs.
+
+
+## Changes to Engine and Vdsm
+
+The high level overview of the workflow in Engine and Vdsm is following:
+
+1.  When refreshing host capabilities Engine picks the information about which
+    CPU is dedicated to VDSM process. It will also read a detailed information
+    about CPU topology on the host.
+
+2.  On VM start/migration/etc. Engine consults the policy for dedicating CPUs
+    and schedules the VM on host with enough free CPUs. Engine determines the
+    mapping between physical and virtual CPUs.
+
+3.  Engine produces domain XML that contains properly populated `<vcpupin>`
+    element (similarly to CPU pinning) and stores the policy name in VM
+    metadata.
+
+4.  Vdsm starts the VM and analyzes which CPUs were dedicated. Vdsm updates the
+    pool of shared CPUs and re-configures VMs without any CPU pinning or
+    dedicated CPUs to run on the new set of CPUs from shared pool.
+
+5.  When the VM with CPU pinning or dedicated CPUs is being destroyed, Vdsm
+    updates the pool of shared CPUs. Vdsm then re-configures VMs without any
+    CPU pinning or dedicated CPUs to run on the new set of CPUs from shared
+    pool. Because we allow overprovisioning CPUs this needs to be done
+    immediately to better utilize the CPUs among VMs.
+
+Several changes to the internal API are required. To accomplish step 1. the
+verb `Host.getCapabilities` will be extended to provide keys:
+
+  * `vdsmCpuAffinity`: list of one or more CPU IDs to which VDSM process is
+    pinned to
+  * `cpuTopology`: detailed description of CPU topology in a form of a list
+    where each object contains:
+
+    * `cpu_id`: ID of the physical CPU
+    * `numa_cell_id`: to which NUMA node the CPU belongs
+    * `socket_id`: on which socket the CPU resides
+    * `die_id`: on which die of the socket this CPU resides; usually there is
+      only one die in the socket so in most cases this will be `0`.
+    * `core_id`: on which core of the die this CPU resides
+
+Further, Engine will add a new entry into VM metadata called `cpuPolicy` that
+will contain the requested CPU policy. The valid values for `cpuPolicy` are:
+
+  * `none`: no pinning or policy selected, this is regular VM with shared CPUs
+  * `pin`: VM has fixed CPU pinning (CPU pinning or NUMA auto pinning)
+  * `dedicated`: VM uses the `dedicated` policy
+  * `isolate-threads`: VM uses the `isolate-threads` policy
+  * `siblings`: VM uses the `siblings` policy
+
+And finally, migration parameters in
+`VM.migrate` will contain a field `cpus` that will contain the CPU mapping for
+the VM on the destination host. The content will be `<vcpupin>` element ready
+for insertion to domain XML. Because the libvirt domain on destination is
+created by VDSM on source it needs to know which CPUs are dedicated to the VM
+on destination so that it can provide a correct domain XML to the destination
+host.
+
+
+## Testing
+
+*TODO*
+
+There are many parts that need to be tested
+
+feature testing:
+- create and especially migration
+- `isolate-threads` is respected
+- `siblings` excludes unused threads
+
+backward compatibility:
+- cpu pinning
+- migration
+
+
+## Future Work
+
+In the beginning CPU hot-plugging will not be supported for VMs with policies
+other than *shared*. To figure out how the hot-plugging should behave for each
+particular policy will remain as a future work.
+
+Use of VMs with dedicated CPU policies may in time lead to fragmentation of
+cores and threads on the hosts. Leading not only to inefficient use,
+difficulties with scheduling but ultimately could prevent new VMs from
+starting on a host. As an example consider a situation where your VM needs to
+use a single socket, but all the sockets are already occupied by smaller VMs.
+So even though there may be enough free cores/threads in total on the host,
+you won't be able to start the VM without removing VMs from one of the sockets
+firsts. Techniques for re-assigning CPUs for consolidating resources need to
+be investigated. It may (or may not) turn out to be a prerequisite for useful
+CPU hot-plugging too.


source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+  * *shared*: *(default)* The guest vCPUs will be allowed to roam among all
+    pCPUs from shared pool.
+  * *dedicated*: The guest vCPUs will be strictly pinned to a set of host pCPUs
+    (similarly to static CPU pinning). The set of pCPUs will be chosen to match
+    the required guest CPU topology. If the host has SMT architecture, thread
+    siblings are preferred.
+  * *isolate-threads*: If the host does not have SMT architecture this is same
+    as *dedicated* policy. If the host has SMT architecture each vCPU is placed
+    on a different physical core and all thread siblings are excluded from use
+    by this and any other VM (rendering all but one thread from a core unusable).
```
mz-pdm, line 56: Does it mean there is no space for a policy that would isolate threads in the described sense, but without strictly pinning guest vCPUs to pCPUs?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+  * *require-threads*: The host must have an SMT architecture. Each vCPU is
```
mz-pdm, line 57: "require-threads" sounds a bit confusing. How about single-VM-threads or something else?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+  * it may be impossible to start a VMs with CPU pinning on that host because
+    the requested CPUs are already occupied;
+
+  * it may be impossible to start a VM with dedicated CPUs even though there
+    seems to be enough space in terms of number of used vs. free CPUs.
+
+To workaround this limitation one can start VMs with CPU pinning before VMs
+with dedicated CPU policy. And more generally, the VMs with dedicated CPUs can
+be manually migrated to another host to free the needed CPUs and then possibly
+migrated back.
```
mz-pdm, line 93: Does it mean it's not possible to move dedicated vCPUs to different pCPUs while a VM is running? And it's possible to do that when migrating (using libvirt hooks)?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+
+*TODO*: Tips how to identify candidates for migration?  virsh? Something in UI?
```
mz-pdm, line 95: I don't understand the question -- what's the problem and how virsh can help solve it?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+## User Experience
+
+## Changes to Engine and Vdsm
+
+The high level overview of the workflow in Engine and Vdsm is following:
+
+1.  When host switches to *UP* state engine queries it to find out which pCPUs
+    are dedicated through a policy or pinned. This is necessary for engine to
+    get a good picture about CPU usage so that it can schedule new VMs
+    properly.
```
mz-pdm, line 107: What does determine the pCPUs used? Engine, libvirt, Vdsm?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+
+2.  On VM start/migration/etc. Engine consults the policy for dedicating CPUs
+    and schedules the VM on host with enough free CPUs. Engine determines the
+    mapping between physical and virtual CPUs.
```
mz-pdm, line 111: If Engine does it, what additional information does Engine need from the host in the preceding point?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+    metadata.
+
+4.  Vdsm starts the VM and analyzes which CPUs were dedicated. Vdsm updates the
+    pool of shared CPUs and re-configures VMs without any CPU pinning or
+    dedicated CPUs to run on the new set of CPUs from shared pool.
+
+5.  When the VM with CPU pinning or dedicated CPUs is being destroyed, Vdsm
+    updates the pool of shared CPUs. Vdsm then re-configures VMs without any
+    CPU pinning or dedicated CPUs to run on the new set of CPUs from shared
+    pool.
```
mz-pdm, line 124: Why is this needed? To spread VMs on more pCPUs in case they shared them before?

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: Let's not also forget about the CPU which Vdsm is pinned to.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Future Work
+
+In the beginning CPU hot-plugging will not be supported for VMs with policies
+other than *shared*. To figure out how the hot-plugging should behave for each
+particular policy will remain as a future work .
+
+Use of VMs with dedicated CPU policies may in time lead to fragmentation of
+resources on the hosts. Leading not only to inefficient use, difficulties with
+scheduling but ultimately could prevent new VMs from starting on a host.
```
mz-pdm, line 167: It would be worth to elaborate a bit more. What kinds of problems, not already experienced without dedicated CPUs, can arise? Do you have NUMA on mind and/or something else?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+    organized
+
+Further, Engine will add a new entry into VM metadata called `cpuPolicy` that
+will contain the requested CPU policy. And finally, migration parameters in
+`VM.migrate` will contain a field `cpus` that will contain the CPU mapping for
+the VM on the destination host. This is necessary so that the source VDSM can
+provide a correct domain XML to the destination host.
+
+*TODO*: In what format? Can we use `<vcpupin>` element content or should we use
+something more concise? Alternatively, we can also let destination VDSM parse
```
mz-pdm, line 140: I'd say <vcpupin> format is OK -- metadata is unfortunately no small thing anyway and reusing <vcpupin> format would keep things easy to understand.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+`<vcpupin>` from XML in `VM.migrationCreate` and use libvirt API to pin the CPUs.
```
mz-pdm, line 141: I'm not sure I understand what you mean here.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Let's say you request a VM with 3 vCPUs and the host has 2 threads per core. Do you mean each vCPU should be able to move between 2 (threads in core) pCPUs or all 6 (threds times cores) pCPUs? Either way, I don't see what could be benefit of that (performance or otherwise).

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: It means this is not something VDSM will do for you. At least now now. It is also mentioned in Future work section.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Let's say user wants somehow to reconcile the resources on the host(s). How should he/she go about figuring out which CPUs are in use and VMs to move. Can we give any guide or tips here? (If not we can drop this point.)

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Engine when the VM is intialy started. But for recovering host engine may or may not remember a correct picture. This possibly depends if this will be stored in DB or not.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: True. I couldn't find it so I don't know if this is hardcoded or configurable. Does engine know which CPU it is?

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: So that the CPUs can be used more efficiently. We allow over-provisioning the CPUs on hosts. 

If we didn't over-provision then theoretically this could be omitted because It will be done on next VM start/migration anyway.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: I am afraid this is little bit hard to explain shortly and without pictures. :) But yes, one such example is when you request a whole NUMA node and you cannot because even though the host may appear to have enough free pCPUs the NUMA node(s) are already occupied by smaller VMs.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: OK. And how about the changes on migration? I'd suggest to clarify both here.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: See __set_cpu_affinity in vdsmd.py. It's configurable and I think Engine doesn't know about the selection (but I guess it should consider it with fixed pinning). Anyway, VMs with dedicated CPUs should avoid sharing their CPUs with Vdsm so perhaps the information should be reported to Engine.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: OK, it would be worth to clarify it in the text.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: I see. I think using ``<vcpuping>`` would still be OK, in accordance with the general approach of preferably providing XML snippets from Engine directly.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: So you mean the destination would decide itself based on the source domain XML, letting libvirt to decide? It doesn't look like a good idea if we want to implement the general policy processing etc. in Engine.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: At least one example explained briefly and without pictures :-), let's use it for clarification (I'm not sure whether there can be other kinds of fragmentation).

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: OK, let's mention these options in the next update of the text, unless we decide for the one or the other now.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: I don't think we want to instruct the users exploring CPU assignments using virsh. A GUI object with squares representing CPUs in NUMA nodes and their assignments to VMs would be nice and helpful for this purpose but maybe too much effort.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: No, I confused static pinning with dedicated pinning here, sorry. I feel like the document would benefit from a short glossary at its beginning.

liranr23: I had the same problem as Milan with confusion on a offline discussion. Pinning != Dedicated and it should be highlighted.

About the isolate-threads itself, I think 1 thread == 1 vCPU, if a host has 2 threads per core and the VM has 2 vCPUs it should take 1 core with the 2 threads in it. Is it talking about 3 vCPUs when one core will be split? Why not allowing VM with 1 vCPU to use that one free thread? Or, let it be shareable?
If this is limiting it, what is the reason for this policy? 
Same for the bottom threads limitation for `require-threads`.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+1.  When host switches to *UP* state engine queries it to find out which pCPUs
+    are dedicated through a policy or pinned. This is necessary for engine to
+    get a good picture about CPU usage so that it can schedule new VMs
+    properly.
+
+2.  On VM start/migration/etc. Engine consults the policy for dedicating CPUs
+    and schedules the VM on host with enough free CPUs. Engine determines the
+    mapping between physical and virtual CPUs.
+
+3.  Engine produces domain XML that contains properly populated `<vcpupin>`
```
liranr23, line 113: this is the same as cpu pinning. do you mean it's going to be provided and unseen by the user?
if we provide it like that from the engine, it might be possible to show the actual mapping of vCPU to pCPU.
Also, it is understandable from the workflow that on migrate we do the same for starting (destination host) and destroy(source host), right?

Is there an option that a running VM with dedicated CPUs and without migrating, will ever change the dedicated CPUs?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+
+5.  When the VM with CPU pinning or dedicated CPUs is being destroyed, Vdsm
+    updates the pool of shared CPUs. Vdsm then re-configures VMs without any
+    CPU pinning or dedicated CPUs to run on the new set of CPUs from shared
+    pool.
+
+Several changes to the internal API are required. To accomplish step 1. new API
+verb `Host.cpuInfo` needs to be defined that will provide:
+
+  * list of pCPUs dedicated to VMs through CPU pinning or policy
```
liranr23, line 129: or a dedicated CPU policy

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+- create and especially migration
+- isolate-threads is respected
+- require-threads excludes unused threads
+
+backward compatibility:
+- cpu pinning
+- migration
+
+
+## Future Work
```
liranr23, line 159: auto pinning "existing/pin" feature, to not clash on the same pCPUs

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Limitations
+
+See the *Security* section.
+
+Currently, when some CPUs are already dedicated to a VM they cannot be evicted
+and provided to another VM. Notably this means:
+
+  * it may be impossible to start a VMs with CPU pinning on that host because
+    the requested CPUs are already occupied;
```
ahadas, line 85: right, and it should be prevented on the engine side, right?
and "a VM" ?

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: if we assume the engine has the complete picture of the CPU assignments and restrictions, then it's just another scheduling-filter that filters out hosts that the vCPUs cannot fit in due to exclusiveness, no? we don't show the user which hosts cannot be scheduled for migration due to other restrictions like memory, host devices, etc, right?

if the user asked to migrate the VM without specifying a target host, we'll find one, if exists
and if the user asked to migrate the VM to a particular host and it doesn't fit, the scheduler would product a relevant message, saying why the host was filtered out

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: good point. for SAP HANA, we reserve the first core of each socket for VDSM IIRC (@liranr23 please keep me honest)

liranr23: Correct, SAP HANA reserves the first core of each socket to the host.
Also, I remember High Performance VM have some pinning (iothread/emulator threads?)

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: the more I think of it, the harder it is for me to find the similarity between what was asked for PIN and this feature (dedicated CPUs).

when you have 1 host with 2 threads and you run 2 VMs with 1 vCPU and policy=PIN then it's similar because you want each VM to be pinned to a different CPU (let's put aside things like VDSM and other system process for simplification)
but let's say that you run a third VM with 1 vCPU and you didn't set it with dedicated policy - its vCPU can be pinned to either of the host threads, right?

so it's similar in a way because in both cases we want the engine to automatically pin vCPUs but if the PIN policy doesn't mean that the pinned vCPU should be pinned exclusively, then it's different and then yeah, vCPUs that were pinned automatically should not be break dedicated CPUs as manual (explicit) pinning

liranr23: Yes, but since the new verb in VDSM - we know which CPUs are taken, so we can create a vCPU pinning according to that. You are right that it's not dedicated, only pinned.
This feature will allow us to PIN to the unused pCPUs. Even without dedicated CPUs. I think it might be also a combination, dedicated+PIN and then it will be also exclusive.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: > I had the same problem as Milan with confusion on a offline discussion. Pinning != Dedicated and it should be highlighted.

Ok, I am adding a short glossary in a next version.

> 
> About the isolate-threads itself, I think 1 thread == 1 vCPU, if a host has 2 threads per core and the VM has 2 vCPUs it should take 1 core with the 2 threads in it.

No, `isolate-threads` will place exactly 1 vCPU on a core, no matter how many threads are in the core. On the other hand `require-threads` does what you are describing. It will fit as many vCPUs on a core as there are threads.

> Is it talking about 3 vCPUs when one core will be split? Why not allowing VM with 1 vCPU to use that one free thread? Or, let it be shareable?
> If this is limiting it, what is the reason for this policy?
> Same for the bottom threads limitation for `require-threads`.

The point of `isolate-threads` and `require-threads` policy is to make sure no other VMs share the cores in use.

And although we could let the vCPUs float around threads in the core if there are unused threads, it brings up a question how to schedule the vCPUs. What if there are multiple vCPUs on the core? Let's say you have 4 threads per core and you only need to use 2, will one vCPU be pinned to 1 thread and the other vCPU pinned to the other 3 or will each vCPU have 2 threads? What if you have 3 vCPUs on such core? Which one of the three vCPUs will be the winner with 2 threads? 

All in all I believe that letting the vCPUs roam around the core defeats the purpose of dedicating CPUs. Of course, the counterargument is that the scheduler on host probably knows best what to do.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Of course I am open to suggestions here and we can rename any of the policies if anyone has a good idea. But for this one I don't think `single-vm-threads` is an improvement.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: I see that this paragraph has been completely misunderstood. I'll think if this can be phrased differently and try to rewrite it or (which is more likely) I will remove it.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: > so it's similar in a way because in both cases we want the engine to automatically pin vCPUs but if the PIN policy doesn't mean that the pinned vCPU should be pinned exclusively,

Well, I thought this is rather an unwanted caveat of PIN policy rather than expected behavior.

> then it's different and then yeah, vCPUs that were pinned automatically should not be break dedicated CPUs as manual (explicit) pinning

I don't think I understand this. You cannot expect pinned CPUs (manual or auto) to share with dedicated CPUs. This breaks the expectation of dedicated CPU. So if there is already VM with pinning started another VM with dedicated CPU policy will have to use different CPUs. On the other hand if there is already a VM with dedicated CPU policy already occupying some CPUs than you cannot start a VM with pinning on the same CPUs (which means you cannot start that VM at all).

> Thinking on it a bit more - what will happen if we combine CPU pinning(static) with dedicated?

On a single VM? You cannot, that does not make much sense.

If all of the above is confusing then we can decouple the expectation on sharing (shared/exclusive) from the CPU assignment (static CPU pinning, NUMA auto pinning or automatic assignment what we call *dedicated CPU policy* here). That way you can create all the various combinations of exclusivity and assignment. This would give the user most flexibility but may be little bit difficult to grasp by users. It still does not solve the scheduling problem though -- you cannot run exclusive VMs and shared VMs on same CPUs.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,170 @@
...
+## Prerequisites
+
+There are no special prerequisites for this feature.
+
+## Security
+
+This feature is primarily designed for performance tuning and is not meant to
+be a security feature. It is possible for other VMs to share the dedicated CPUs
+for a short period of time when the VM is brought up (boot, after migration,
+restore from snapshots, etc.). This is not so much problem during boot because
```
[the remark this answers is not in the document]
nyoxi: I assume this is same as starting a snapshot, no?

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: > Correct, SAP HANA reserves the first core of each socket to the host.

Do we expect that only VMs with dedicated CPUs will run on SAP HANA envs? If not VDSM needs to know about this to avoid placing shared VMs on the cores.

> Also, I remember High Performance VM have some pinning (iothread/emulator threads?)

Yes, looking at how this works it is kind of strange. It picks first two CPUs of NUMA node that a VM has most CPUs pinned on. But it does not necessarily mean that the CPUs are actually pinned by that VM. We will need to clarify how we want that to interact with shared VMs.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Paragraph removed.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Updated with an example.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: No, source needs to get the information from engine and pass it to destination. The above paragraph has been updated.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: > this is the same as cpu pinning. 

Technically yes, since it's going to be decided by engine we don't have to use libvirt API.

> do you mean it's going to be provided and unseen by the user?
> if we provide it like that from the engine, it might be possible to show the actual mapping of vCPU to pCPU.

We can think of some UI changes that would provide this information somehow. Maybe some host-level view of how all its CPUs are used.

> Also, it is understandable from the workflow that on migrate we do the same for starting (destination host) and destroy(source host), right?

For migration we need to pass this information to the source VDSM. It will then pass that to destination.

> 
> Is there an option that a running VM with dedicated CPUs and without migrating, will ever change the dedicated CPUs?

Presently no.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: yes

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+  * VM does not need to be pinned to a specific host(s),
+  * takes into consideration existing VMs which solves one of the issues with
+    *Pin* NUMA auto pinning policy where several VMs can end up to be pinned
+    to same CPUs,
+  * it makes sure that VMs without any dedicated CPUs don't share CPUs with
+    VMs that have dedicated CPUs
+
+The available policies described below divide the set of available pCPUs into
+three sub-sets, There are the dedicated CPUs that are assigned only to a
+single VM. Then there are some unassigned CPUs that cannot be used by any
```
mz-pdm, line 60: "... only to a single VM." -- each of the pCPUs is assigned to  a single VM, not the whole pCPU group, right?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+other VM because of restrictions requirements on the threads of the core. And
```
mz-pdm, line 61: CPU = thread here, right?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+finally there is the set of CPUs that are not strictly assigned to any
+particular VM and are shared between all the VMs with *shared* policy. This
+subset is hereby referred to as *shared pool*. When scheduling VMs engine also
+takes into consideration use of host CPUs that play a special role, e.g. CPU
+that VDSM is pinned to or CPUs reserved to host on SAP HANA systems.
```
mz-pdm, line 66: And those CPUs belong to which of the three groups?

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: How about "consolidate-threads"? I still don't think there are any "required" threads here.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+Currently, when a CPU is already dedicated to some VM, VDSM will not move the
+VM onto another CPU to make it free to another VM. This may be addressed in
+the future (see *Future Work* section), but for now this means:
+
+  * It may be impossible to start VMs with CPU pinning on that host because
+    the requested CPUs are already occupied. It is suggested to start VMs with
+    CPU pinning before any VMs with dedicated CPU policy.
+
+  * It may be impossible to start a VM with dedicated CPUs even though there
+    seems to be enough space in terms of number of used vs. free CPUs.
```
mz-pdm, line 117: Why CPU movement is needed here when dedicated vCPUs can be placed on any pCPUs? Is it because of isolate-threads policy, which could be used if threads of other VMs could be consolidated?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+The high level overview of the workflow in Engine and Vdsm is following:
+
+1.  When host switches to *UP* state engine queries it to find out which pCPUs
+    are dedicated through a policy or pinned. This is necessary for engine to
+    get a good picture about CPU usage so that it can schedule new VMs
+    properly. Although Engine is the one responsible for assigning the CPUs
+    and should know which CPUs are used by which VM, it may not remember the
+    state for recovering hosts or after Engine restart. Whether this is needed
+    at all probably depeonds on if the information about CPU allocations will
+    be stored in the database or not.
```
mz-pdm, line 131: Isn't it easier to store it in the database rather than query the hosts?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+    and schedules the VM on host with enough free CPUs. Engine determines the
+    mapping between physical and virtual CPUs.
+
+3.  Engine produces domain XML that contains properly populated `<vcpupin>`
+    element (similarly to CPU pinning) and stores the policy name in VM
+    metadata.
+
+4.  Vdsm starts the VM and analyzes which CPUs were dedicated. Vdsm updates the
+    pool of shared CPUs and re-configures VMs without any CPU pinning or
+    dedicated CPUs to run on the new set of CPUs from shared pool.
```
mz-pdm, line 143: What with older Vdsm versions, which don't support this?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+backward compatibility:
+- cpu pinning
+- migration
+
+
+## Future Work
+
+In the beginning CPU hot-plugging will not be supported for VMs with policies
+other than *shared*. To figure out how the hot-plugging should behave for each
+particular policy will remain as a future work .
```
mz-pdm, line 190: An extra space before the final dot.

source/develop/release-management/features/virt/dedicated-cpu.md
liranr23: > On a single VM? You cannot, that does not make much sense.
Why not? :)
If I want the VM to use specific pCPUs, and also to have them dedicated for my VM. As long they are freed on the host I can't see reason why it won't be possible.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: The feature will be tied to some future cluster level so old VDSM should not be a problem. As for the opposite case (old engine and newer VDSM) we will treat all VMs without policy in metadata as shared unless they have explicit CPU pinning in domain XML.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: That will be different for each case. I suppose the CPU where VDSM is pinned to will be in shared pool. For the SAP HANA I need to understand the situation little bit more.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: removed

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,202 @@
...
+    immediately to better utilize the CPUs among VMs.
+
+Several changes to the internal API are required. To accomplish step 1. new API
+verb `Host.cpuInfo` needs to be defined that will provide:
+
+  * CPU that VDSM is pinned to
+  * list of pCPUs dedicated to VMs through CPU pinning or auto pinning policy
+    or dedicated CPU policy
+  * CPU indices and topology information how are sockets, cores and threads
+    organized
```
nyoxi, line 159: I've been thinking about this a little and it may be redundant to add a new verb for this. If Engine will store the CPU assignments in DB and we don't have to report it (second bullet), the rest of the information (first and third bullet) can be added to `Host.getCapabilities`.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,202 @@
...
+
+  * It may be impossible to start a VM with dedicated CPUs even though there
+    seems to be enough space in terms of number of used vs. free CPUs.
+
+
+## Changes to Engine and Vdsm
+
+The high level overview of the workflow in Engine and Vdsm is following:
+
+1.  When host switches to *UP* state engine queries it to find out which pCPUs
```
ahadas, line 125: before the host reaches the UP state, there should be no VMs running on it (unless it was non-responding) - did you mean a different state?

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: yeah, this concept of reducing the amount of data in the database by getting the most up-to-date data on engine startup rather than restoring the "previously known state" from the database was always temping and we never really pushed for it

so let's think about it for sec - the engine needs the pinning and policies that it requested for each VM, right? if the engine doesn't need anything else from VDSM then maybe yes, it would be easier to store it in vm_dynamic - both the generated pinning and the dedicated policy..

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,202 @@
...
+5.  When the VM with CPU pinning or dedicated CPUs is being destroyed, Vdsm
+    updates the pool of shared CPUs. Vdsm then re-configures VMs without any
+    CPU pinning or dedicated CPUs to run on the new set of CPUs from shared
+    pool. Because we allow overprovisioning CPUs this needs to be done
+    immediately to better utilize the CPUs among VMs.
+
+Several changes to the internal API are required. To accomplish step 1. new API
+verb `Host.cpuInfo` needs to be defined that will provide:
+
+  * CPU that VDSM is pinned to
```
ahadas, line 155: the question is if we want to pin VDSM to a particular CPU - maybe it's better to keep it "floating" and just avoid dedicating the first CPU. so generally the OS will move VDSM as needed but will always have that first CPU (currently it's per-socket for SAP HANA I believe) to run on

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: Yeah

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: > > On a single VM? You cannot, that does not make much sense.
> 
> Why not? :)
> If I want the VM to use specific pCPUs, and also to have them dedicated for my VM. As long they are freed on the host I can't see reason why it won't be possible.

correct me if I'm wrong but I think you both are talking about different things:
@liranr23  meant to have certain vCPUs both pinned and defined as dedicated while @nyoxi  referred to having some of the VMs pinned without being dedicated and some pinned as dedicated
do I get this right?

>If all of the above is confusing then we can decouple the expectation on sharing (shared/exclusive) from the CPU assignment (static CPU pinning, NUMA auto pinning or automatic assignment what we call dedicated CPU policy here). That way you can create all the various combinations of exclusivity and assignment. This would give the user most flexibility but may be little bit difficult to grasp by users. It still does not solve the scheduling problem though -- you cannot run exclusive VMs and shared VMs on same CPUs.

this makes sense to me.
so users will be able to:
1. set pinning explicitly and then set whether they should be shared or exclusive
2. use auto pinning policy - either the 'resize and pin' we have for SAP HANA or 'pin' that doesn't resize or 'dedicated' is similar to 'pin' in the sense that it respects the specified CPU topology but pin it as described above or the other two policies

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: And what would be the practical difference between Vdsm pinned to a particular CPU and having the first CPU always available for Vdsm? Also note that:
- Vdsm cannot be moved among CPUs completely freely at OS discretion, its pinning is very important for Vdsm performance.
- The CPU which Vdsm is pinned to can be currently user-defined in Vdsm configuration.

ahadas: ok, I didn't take pinning that is not via VDSM configuration into consideration here - if it's possible then yeah, the engine should respect it

yeah, generally it should be similar I think besides use cases like running a different process on the host that is also pinned to the first CPU and if VDSM wouldn't have been pinned to it as well, it could have get more CPU shares elsewhere

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: If that's ok then lets assume these "reserved" are part of shared pool from VDSM POV to make it simple. If this turns out to be a problem we can fix that later.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: What about just "siblings"?

source/develop/release-management/features/virt/dedicated-cpu.md
liranr23: That seems right.
If we give `fixed` CPUs and we know they are available on the host, I think we can either make the exclusive or not.

Thinking a bit on problems - when you set it the CPU `fixed` and `shared` (let's call it VM1). Once a VM with dedicated CPU starts (VM2), it might cause VM1 to lose the used CPUs as we re-assign them, right? I think we should document it as it might happen in such cases. If we are going to support the above logic.
This is the only problem I can think of by allowing it.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: Better than "require-threads".

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: According to 4. above, VMs with pinned CPUs are not touched on re-assignment (but non-pinned/dedicated VMs can be additionally placed on their CPUs). 
I don't know what all use cases there are for pinned CPUs but I can imagine one may want to have exclusive pinned CPUs.

liranr23: I was wrong in the previous comment. The conflict shouldn't happen. if the fixed CPU pinning is running (and we know these CPUs were available), they will run on them. Once you run a dedicated CPU VM, if there are not enough freed CPUs, it will fail. Otherwise it will use what available in the host. So conflict shouldn't happen.

ahadas: I think we can rule out dedicated with shared and think of it as the case of no-method + exclusive, no?

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+## Owner
+
+* Name: Tomáš Golembiovský
+* Email: tgolembi@redhat.com
+
+## Glossary
+
+  * *CPU pinning*: Feature in oVirt that allows you to statically pin vCPUs
```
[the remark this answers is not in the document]
nyoxi: done

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: I have changed it to `siblings` but I am not quite satisfied with it yet. If anyone still has some other suggestions feel free to share them.

source/develop/release-management/features/virt/dedicated-cpu.md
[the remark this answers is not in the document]
nyoxi: I'll go with "restricted" then.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: This is fully configurable in `vdsm.conf`. User can use special string `auto` which tells VDSM to pick a CPU or can specify a comma separated list of CPUs.

source/develop/release-management/features/virt/dedicated-cpu.md
mz-pdm: I think my original comment was about something that was clarified in an updated version of the document, so it can be marked as resolved.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,211 @@
...
+
+  * `vdsmCpuAffinity`: list of one or more CPU IDs to which VDSM process is
+    pinned to
+  * `cpuTopology`: detailed description of CPU topology in a form of a list
+    where each object contains:
+
+    * `cpu_id`: ID of the physical CPU
+    * `numa_cell_id`: to which NUMA node the CPU belongs
+    * `socket_id`: on which socket the CPU resides
+    * `die_id`: on which die of the socket this CPU resides; usually there is
```
nyoxi, line 158: The information about dies is not interesting for us at the moment, but I have included it too for completeness. If anyone thinks it should be removed let me know.

nyoxi: Actually I just realized that cores are indexed per-die and not per-socket. So removing die_id would mean we need to reindex core_ids.

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: > I think we can rule out dedicated with shared and think of it as the case of no-method + exclusive, no?

what I meant here is that when no assignment is defined (no method) and yet exclusiveness is requested then the CPUs will practically be "dedicated" (as they are pinned exclusively) and assigned by the logic that was described for the `dedicated` policy

other than that, lgtm

ahadas (commented):
very few things that are not entirely clear to me but the overall design looks good and there are patches posted already so let's wrap up the discussion about these things and merge this PR

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,211 @@
...
+      only one die in the socket so in most cases this will be `0`.
+    * `core_id`: on which core of the die this CPU resides
+
+Further, Engine will add a new entry into VM metadata called `cpuPolicy` that
+will contain the requested CPU policy. The valid values for `cpuPolicy` are:
+
+  * `none`: no pinning or policy selected, this is regular VM with shared CPUs
+  * `pin`: VM has fixed CPU pinning (CPU pinning or NUMA auto pinning)
```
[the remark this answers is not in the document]
nyoxi: We need to differentiate for migration workflow:
* `none`: means shared pool is used, we need to remove pinning specification from domain XML before sending it to destination, destination VDSM will take care of assigning CPUs from shared pool
* `pin`: this is the only case when we want to keep pinning information in domain XML during migration
* *other policy*: we need to remove pinning information and replace it by the information provided by Engine in `VM.migrate`.

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: After ruling out the uncertain combinations the only valid combination seems to be for CPU pinning with exclusive CPUs (no sharing). In my opinion tracking the exclusivity separately is unnecessary complication. The same can be achieved (from VDSM perspective) by adding a new value (e.g. `pin-exclusive`) to `cpuPolicy` enum in metadata.

This is an easy change and can be done any time later once we decide to implement this feature. I am going to resolve this conversation now.

source/develop/release-management/features/virt/dedicated-cpu.md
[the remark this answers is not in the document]
ahadas: ack, makes sense

ahadas (approved):

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -0,0 +1,211 @@
...
+5.  When the VM with CPU pinning or dedicated CPUs is being destroyed, Vdsm
+    updates the pool of shared CPUs. Vdsm then re-configures VMs without any
+    CPU pinning or dedicated CPUs to run on the new set of CPUs from shared
+    pool. Because we allow overprovisioning CPUs this needs to be done
+    immediately to better utilize the CPUs among VMs.
+
+Several changes to the internal API are required. To accomplish step 1. the
+verb `Host.getCapabilities` will be extended to provide keys:
+
+  * `vdsmCpuAffinity`: list of one or more CPU IDs to which VDSM process is
```
mz-pdm, line 150: This is called `vdsmToCpusAffinity` now. But it's OK to update this and the following info later, based on the actual implementation.

mz-pdm (approved):

sandrobonazzola (approved):

PR #2582, opened 2021-10-12
Add VM screenshot feature page

saifabusaleh:

Add VM screenshot feature page

This pull request needs review by: @ahadas @liranr23 @mz-pdm  @nyoxi @ljelinkova  @smelamud

### DIFF
--- a/source/develop/release-management/features/virt/vm-screenshot.md
+++ b/source/develop/release-management/features/virt/vm-screenshot.md
@@ -1,0 +1,90 @@
+---
+Title: VM Screenshot
+Category: feature
+Authors: sabusale
+---
+
+# VM Screenshot
+
+## Summary
+
+Allow users to take screenshot of a running VM
+
+## Owner
+
+*   Name: Saif Abu Saleh (sabusale)
+*   Email: sabusale@redhat.com
+
+## Description
+
+Users need a way to see the current state of their VMs for debug purposes,
+when they don't have console access
+
+### UI
+
+Not supported
+
+### VDSM
+
+VDSM will call Libvirt screenshot API [[2]]
+
+Libvirt will call QEMU to capture the screenshot
+
+then VDSM will return `ppm` file content as base64 string
+
+
+
+### Backend
+
+Backend will call VDSM to capture screenshot and will receive base64 string screenshot
+and then base64 string will be converted to byte array and returned
+
+### REST API
+
+Rest API will call Backend to capture the screenshot
+
+Rest API call is synchronous operation
+
+Rest API implementation is intended for debugging only
+
+API call Example:
+
+`/vms/{vm:id}/screenshot`
+
+```
+      POST /vms/1234/screenshot
+      Accept: application/xml
+```
+
+Response is a byte array of the screenshot
+```
+iVBORw0K..
+```
+
+### Image format issues
+
+`ppm` is not a standard screenshot format, and users may have issues to see the screenshot, and it is not straight forward
+to display it in HTML[[4]], unlike e.g. `png base64`
+
+issue[[5]] opened to QEMU to add screen dump as PNG
+
+other option is to convert the `ppm` content to other standard format like `png`
+ways to convert it:
+*   Convert it in VDSM using PIL image library: [example](https://stackoverflow.com/questions/69675903/how-do-i-use-python-image-library-to-convert-a-libvirt-stream-to-a-different-ima)
+*   In Engine use java API to convert it using [ImageMagick](https://stackoverflow.com/a/2291493/5825468)
+
+### Future work
+*   Add the functionality to web UI
+*   Libvirt screenshot API provides option to capture screenshot by screen number (in case there are multiple screens).
+
+    for now by default 0 only used(first screen),
+    add the option to take screenshot by screen number
+    
+* Change generate screenshot API call to be asynchronous
+
+
+[1]: https://fileinfo.com/extension/ppm
+[2]: https://libvirt.org/docs/libvirt-appdev-guide-python/en-US/html/ch12s02.html
+[3]: https://github.com/qemu/qemu/blob/0289f62335b2af49f6c30296cc00d009995b35f6/ui/console.c#L420
+[4]: https://stackoverflow.com/a/41647242/5825468
+[5]: https://gitlab.com/qemu-project/qemu/-/issues/718


liranr23 (approved):
/LGTM

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
+---
+title: VM Screenshot
```
liranr23, line 2: Capital letter on title -> Title
Category, Authors.

liranr23 (commented):
dots at the end of sentences.

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
...
+Allow users to take screenshot of the VM
+
+## Owner
+
+*   Name: Saif Abu Saleh (sabusale)
+*   Email: sabusale@redhat.com
+
+## Description
+
+User can take a screenshot of running VM
```
ahadas, line 20: a running VM

and please elaborate on the use case - it's mainly intended for debugging, when you have no access (console) to the guest

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
...
+Backend will call VDSM to capture screenshot and will receive the `ppm`[[1]] file content
+and then `ppm` file content will be returned
+
+### VDSM
+
+VDSM will call Libvirt API[[2]] that capture the screenshot and saves it on given path on the host file system
+and then will return the `ppm` file content and remove the image on the host file system
+
+Saving the image:
+*   path will be temporary directory path
```
ahadas, line 37: I think we have a vm-specific directory we can place it in

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Backend
+
+Backend will call VDSM to capture screenshot and will receive the `ppm`[[1]] file content
+and then `ppm` file content will be returned
+
+### VDSM
+
+VDSM will call Libvirt API[[2]] that capture the screenshot and saves it on given path on the host file system
+and then will return the `ppm` file content and remove the image on the host file system
```
ahadas, line 34: it's a quick operation, right?

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Image format issues
+
+`ppm` image is not a standard image format, and users may have issues to see the image, and it is not straight forward
+to display it in HTML[[4]], unlike e.x `png base64`
+
+other option is to convert the `ppm` content to other standard format like `png`
+ways to convert it:
+*   Run command line command to convert it using ImageMagick
+*   Use java API to convert it using ImageMagick , e.x: https://stackoverflow.com/a/2291493/5825468
```
ahadas, line 69: e.g.?

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,83 @@
...
+and then encode with `base64`, similar to icons
```
ahadas, line 70: I think that as it is mainly intended for debugging, users can install GIMP or any other tool that opens ppm
but yeah, it would be nice to design this operation to be extensible so if, e.g., a UI client would like to present it, it can ask for receiving it in a given format (like png) and the engine will convert it

source/develop/release-management/features/virt/vm-screenshot.md
saifabusaleh: Done

source/develop/release-management/features/virt/vm-screenshot.md
saifabusaleh: actually no, since the file is not compressed, see: https://en.wikipedia.org/wiki/Netpbm#:~:text=The%20PPM%20format%20is%20not,file%20size%20is%2073%2C848%20bytes.

I took image and it was with size 2.3MB, so it will take time to read the file and return it

source/develop/release-management/features/virt/vm-screenshot.md
mz-pdm: Both libvirt (under /var/lib/libvirt/qemu) and Vdsm (under /run/vdsm) create VM specific subdirectories, not necessarily for general use though.

saifabusaleh: looks like definitely /var/lib/libvirt/ could be good place (Like TPM) or /var/lib/libvirt/qemu/

but again, depending what we want to do with this

If we don't want to persist the images in the host, I think the directory path shouldn't really matter and creating it in random place is enough, since it will be deleted

otherwise, if we want to keep the images in the host, then its really matter

source/develop/release-management/features/virt/vm-screenshot.md
saifabusaleh: maybe it can cause an issue to the API?

lets assume the image size is >5Mb
download speed is less then 1Mbit/s

the response from VDSM alone will take > 1 minute (only the download part)
not considering other things like running the command, parse the file, ....

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,86 @@
...
+Not supported
+
+### Backend
+
+Backend will call VDSM to capture screenshot and will receive the `ppm`[[1]] file content
+and then `ppm` file content will be returned
+
+### VDSM
+
+VDSM will call Libvirt API[[2]] that capture the screenshot and saves it on given path on the host file system
```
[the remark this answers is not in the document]
saifabusaleh: changed the sentence a little bit, what about now?

source/develop/release-management/features/virt/vm-screenshot.md
ahadas: we typically don't want clients to wait on long operations and so we implement them as asynchronous operations but it would really complicate the implementation here

so ok, if it can take relatively long time and we implement it as a synchronous operation, I think we'll need to make it crystal clear that the current implementation is intended for debugging also at the API level

source/develop/release-management/features/virt/vm-screenshot.md
ahadas: IIUC, we don't have an alternative but to persist it on the host and then read it and send it to the engine, right?
I think that by using an existing per-VM folder we are (1) more consistent (as you wrote, e.g., with TPM); and (2) assuming we already clear those folders when destroying VMs, it would simplify handling of negative flows

saifabusaleh: agree

the thing is that we need such a folder, such that it will get deleted when the VM is destroyed

for tpm and nvram from what I checked this folders is created and managed by Libvirt

this is not the case here, since not specifying path will generate the screenshot on the root directory

saifabusaleh:
Looks like Libvirt already have support for PNG: https://github.com/libvirt/libvirt/blob/master/tools/virsh-domain.c#L5526

But QEMU Driver still doesn't support this MIME type: https://github.com/libvirt/libvirt/blob/master/src/qemu/qemu_driver.c#L3474

Can we file an issue for them to add PNG support?

ahadas:
> Looks like Libvirt already have support for PNG: https://github.com/libvirt/libvirt/blob/master/tools/virsh-domain.c#L5526
> 
> But QEMU Driver still doesn't support this MIME type: https://github.com/libvirt/libvirt/blob/master/src/qemu/qemu_driver.c#L3474
> 
> QEMU Monitor screendump: https://qemu-project.gitlab.io/qemu/system/monitor.html
> 
> Can we file an issue for them to add PNG support?

sure, feel free to file a bug

source/develop/release-management/features/virt/vm-screenshot.md
mz-pdm: Using a /var/lib/libvirt subdirectory is probably not a good idea, this directory belongs to libvirt and we shouldn't mess with it. As for TPM and NVRAM, we use it only to read and restore data managed by libvirt, as already pointed out by Saif.
Using a /run subdirectory is probably also not a good idea, this directory servers AFAIK for files describing the state of the system, such as pid files, mount points, etc., not for data dumps.
/tmp or a similar location looks like the best place for such a temporary data that is needed only during the API call. Using Python's tempfile.TemporaryDirectory should do the right thing, including selecting the right and safe location.

ahadas: @mz-pdm ok, if we do that in a single call, i.e., call libvirt to dump the data and then immediately read the data, send it to the engine and remove the data, then yeah, a temporary folder sounds ok. but if we are implementing it in a way to prepares it for asynchronous operation, I'd prefer to avoid sending the path we dump the data back to the engine so the engine would later provide it back to VDSM or to cache it in VDSM, I think in that case it would be better to rely on some convention for where the data is stored

mz-pdm: Then a subdirectory of /var/lib/vdsm could be used for such a bit more persistent purpose.

saifabusaleh: few questions:
1- in /var/lib/vdsm Can we just do something like: /var/lib/vdsm/screenshots and then the screenshots names contain the VM name on them (see page Saving the image section)

2- what will happen when upgrading VDSM host to this directory?

mz-pdm: > 1- in /var/lib/vdsm Can we just do something like: /var/lib/vdsm/screenshots and then the screenshots names contain the VM name on them (see page Saving the image section)

Yes. There just needs to be some form of request-file mapping in case of asynchronous requests. And maybe using VM ids rather than names in the file names would be better -- less user friendly but more robust.

> 2- what will happen when upgrading VDSM host to this directory?

In the default/recommended setup, Vdsm cannot be upgraded when VMs are running on the host. The directory can simply stay there. Some content cleanup mechanism should be considered; maybe using systemd-tmpfiles would be the easiest way.

saifabusaleh: we can say we save to the static path  (e.g /var/lib/vdsm/screenshots) with file-name as generated in libvirt (<vm_name>-<timestamp>) for now

And the other parts (async request and all its requested work) will be in Future work

saifabusaleh:
> > Looks like Libvirt already have support for PNG: https://github.com/libvirt/libvirt/blob/master/tools/virsh-domain.c#L5526
> > But QEMU Driver still doesn't support this MIME type: https://github.com/libvirt/libvirt/blob/master/src/qemu/qemu_driver.c#L3474
> > QEMU Monitor screendump: https://qemu-project.gitlab.io/qemu/system/monitor.html
> > Can we file an issue for them to add PNG support?
> 
> sure, feel free to file a bug but it shouldn't block us from proceeding with this feature that is intended for improved debugging

opened: https://gitlab.com/qemu-project/qemu/-/issues/718

it shouldn't block us, but trying will not harm :)

source/develop/release-management/features/virt/vm-screenshot.md
```diff
@@ -0,0 +1,101 @@
...
+
+Libvirt will call QEMU to capture the screenshot and save it on given path on the host file system
+
+then VDSM will return the `ppm` file content and remove the screenshot on the host file system
+
+#### Saving the screenshot:
+*   screenshot name will be the same as generated by Libvirt, `<vm_name>-<host_date>-<host_time>.ppm`
+
+    e.g: `vm_centos-2021-11-10-12:00:47.ppm`
+*   path will be saved to `/var/lib/vdsm/screenshots` path
```
ahadas, line 39: > path will be saved to

the file will be saved to?

ahadas (changes requested):
minor comment, other than that lgtm

ahadas (approved):

liranr23 (approved):

mz-pdm (approved):

saifabusaleh:
@sandrobonazzola can we go ahead and merge this PR please?

Thanks

sandrobonazzola (approved):

PR #2609, opened 2021-10-25
Adding preview builds

janosdebugs:
This PR adds preview builds for the oVirt website using GitHub Actions and GitHub pages. It works by building the oVirt site in a container, then copying the data, and finally pushing it to the `gh-pages` branch. This branch will need to be created empty.

When building, the GitHub action is configured to only run one job in parallel, which avoids problems with multiple jobs overwriting each other.

It also comments on the PR so contributors get their preview link. (This may possibly need a personal access token from a bot account.)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,106 @@
...
+        tags: ovirt-site:latest
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
+    - name: Set up Jekyll cache
+      uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
```
duck-rh, line 38: IIUC this key does not exist and that would give the same result as not specifying `restore-keys`

.github/workflows/preview.yml
```diff
@@ -0,0 +1,106 @@
...
+    - name: "Change config"
+      run: |
+        set -exuo pipefail
+        PR_ID=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
+        PR_LINK=https://$(echo "${{github.repository}}" | sed -e 's#/#.github.io/#')/previews/${PR_ID}
```
duck-rh, line 43: I would has created the path and base first and concatenated to make PR_LINK; I found this a bit confusing.

.github/workflows/remove-preview.yml
```diff
@@ -0,0 +1,41 @@
...
+      cancel-in-progress: false
+    steps:
+    - name: Checkout gh-actions branch
+      uses: actions/checkout@v2
+      with:
+        ref: gh-pages
+    - name: Comment on PR
+      run: |
+        set -euo pipefail
+        PR_ID=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
```
duck-rh, line 21: PR-ID is recalculated so many times and sometimes even inlined which does not help reading the code.
Maybe we could consider using one of the methods highlighted in: https://stackoverflow.com/questions/57819539/github-actions-how-to-share-a-calculated-value-between-job-steps
(as you did for targetdir)

duck-rh (changes requested):
Looks pretty good, but added a few remarks.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    && echo "bundle install --quiet" >>/usr/local/bin/init.sh \
+    && echo 'exec bundle exec jekyll "$@"' >>/usr/local/bin/init.sh \
+    && chmod +x /usr/local/bin/init.sh
+
+WORKDIR /srv/site
+VOLUME /srv/site
+EXPOSE 4000
+
+ENTRYPOINT ["/usr/local/bin/init.sh"]
+CMD ["s","-t","-l","--host=0.0.0.0"]
```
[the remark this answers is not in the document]
janosdebugs: Exposed  that port, want people to be able to do edits in their local env. For the build process the `CMD` is overridden anyway.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM docker.io/library/ruby:2
+
+RUN DEBIAN_FRONTEND=noninteractive apt update \
```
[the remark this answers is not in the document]
janosdebugs: Done, used setup.sh for this purpose.

.github/workflows/preview.yml
janosdebugs: Done.

.github/workflows/remove-preview.yml
janosdebugs: Done.

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM docker.io/library/ruby:2
+
+# Copy and run the setup.
+COPY setup.sh /usr/local/bin/setup.sh
+RUN DEBIAN_FRONTEND=noninteractive apt update \
+    && apt install -y sudo \
+    && DEBIAN_FRONTEND=noninteractive SKIP_BUNDLE=1 /usr/local/bin/setup.sh \
+    && rm -rf /var/lib/apt/lists/*
```
duck-rh, line 8: missing `apt clean` to remove downloaded debs in /var/cache/apt/archives/.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# HTTP port.
+EXPOSE 4000
+
+# Live-reload port.
+EXPOSE 35729
+
+# All parameters passed to the container get routed through a script we create here.
+RUN echo "#!/bin/sh" >/usr/local/bin/init.sh \
+    && echo "set -e" >>/usr/local/bin/init.sh \
+    && echo "bundle config --local path 'vendor/bundle'" >>/usr/local/bin/init.sh \
```
duck-rh, line 24: bundle config/install are already done in `setup.sh`
(and btw we do not need to repeat them in `run-server.sh` either; I often forget about this script and run the exec manually and that's really not needed)

duck-rh (changes requested):
Looks really good.
The variables declaration makes things way easier to read.
I added minor comments and I think we're close to merging.

Dockerfile
janosdebugs: I made it conditional so that we can do it runtime rather than build time.

duck-rh:
About dropping the scripts: I liked the idea to have the choice to use a container or not but keep the steps in common with both methods. I'll let @sandrobonazzola decide what is best, just my opinion.

I also spotted another problem. We do not build with ubuntu-lastest and I would not suggest using a fast-moving target for such project since we do not need bleeding edge features. Whatever target we choose in the future I think we should use the same base distro that is currently used to generate the production website. Even if the CI will catch any problem that would make development easier.

duck-rh (approved):
LGTM
It's a good idea to have brought all the checks there.

janosdebugs:
@duck-rh I don't have rights to merge or retrigger the test run.

duck-rh:
@janosdebugs I restarted the action but it fails due to the same problem.

janosdebugs:
@duck-rh we'll need the machine user then.

janosdebugs:
@sandrobonazzola help please?

sandrobonazzola (approved):
lgtn

sandrobonazzola:
Checked with @janosdebugs ok to merge as the build is failing due to PR coming from a fork.

duck-rh:
@sandrobonazzola OK

PR #2622, opened 2021-11-03
Add parallel migration connections feature page

mz-pdm:

- Adding a feature page about parallel migration connections.

This pull request needs review by: @ahadas

### DIFF
--- a/source/develop/release-management/features/virt/parallel-migration-connections.md
+++ b/source/develop/release-management/features/virt/parallel-migration-connections.md
@@ -1,0 +1,92 @@
+---
+title: Parallel migration connections
+category: feature
+authors: mzamazal
+---
+
+# Parallel migration connections
+
+## Summary
+
+This feature adds support for multiple connections in migrations.
+
+## Owner
+
+*   Name: Milan Zamazal
+*   Email: mzamazal@redhat.com
+
+## Description
+
+libvirt and QEMU allow using multiple parallel connections for a single migration (QEMU calls this feature *multifd*).  With just a single migration connection and sufficient network bandwidth, the single migration thread can become a bottleneck due to a limited single-CPU capacity.  By using multiple connections fed with multiple threads run by multiple CPUs, a full network capacity can be utilized.  For instance, 8-16 connections can saturate 100 Gbps network bandwidth.
+
+Additionally, multifd implementation of data transfer is simpler and more efficient than the traditional migration mechanism and can be faster even with lower network bandwidths.  This concerns only the way memory data is transferred, the other migration concepts such as counting iterations or setting the maximum downtime are unaffected.
+
+Parallel migration connections can be enabled by passing ``VIR_MIGRATE_PARALLEL`` flag to [virDomainMigrateToURI3](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainMigrateToURI3) call.  The number of connections to use is specified using  [VIR_MIGRATE_PARAM_PARALLEL_CONNECTIONS](https://libvirt.org/html/libvirt-libvirt-domain.html#VIR_MIGRATE_PARAM_PARALLEL_CONNECTIONS) migration parameter.
+
+## Prerequisites
+
+Nothing special, just up-to-date libvirt and QEMU.
+
+## Limitations
+
+Current implementation of parallel migration connections in QEMU has some restrictions:
+
+- It doesn't support switching to post-copy.
+
+- RDMA doesn't work.  This is not a limitation for oVirt, because RDMA is not supported in oVirt.
+
+- XBZRLE compression is not supported.  However this kind of compression is generally not recommended to use in migrations because it adds little benefit and a significant overhead.  gzip/zstd is supported instead and migrations should actually benefit from it (but libvirt apparently doesn't currently provide means to select this compression kind).
+
+- Due to their better implementation, parallel migration connections can be faster than the traditional migration implementation even when the network bandwidth can be saturated with a single connection.  However, when single connection is specified, the parallel migration connection implementation uses a different code path, which may not be sufficiently robust, and it's recommended to always use at least two connections.  In the worst case, one of the threads will be idle while causing only a little overhead. 
+
+- Parallel migration connections haven't been tested extensively with connections faster than 100 Gbps.  Due to current multifd implementation, there is an increasing overhead when using many connections and it's not clear whether using more than 16 parallel connections adds a real benefit.
+
+- It still applies that large guests with terrabytes of RAM are difficult to migrate successfully.
+
+- Since it's easier to saturate network bandwidth with multiple connections, it's advisable to not migrate multiple VMs from a single host at the same time.
+
+- VM CPU pinning may restrict the number of CPUs available to the QEMU process and thus impose a limit on the meaningful number of parallel connections.
+
+- It's difficult to guess the right number of parallel connections to use.  A rule of thumb is to use one connection per 10 Gbps of available network bandwidth.  If the number of connections is a bit high, some threads may remain idle while the related overhead should be negligible.
+
+- There is very little documentation about parallel migration connections and the information above was obtained directly from platform developers.
+
+## User experience
+
+A new migration option to enable parallel migration connections is introduced in cluster and VM edit dialogs.  The cluster option specifies what to use in the given cluster.  VMs can override the cluster value individually.  The new option offers the following choices:
+
+- **Disabled** (the default in clusters): Parallel migration connections are not used.
+
+- **Auto**: Use or do not use parallel migration connections and set their number automatically based on the available resources (network bandwidth and CPU power).
+
+- **Parallel**: Always use parallel migration connections (for the supposed benefit of a more efficient implementation over non-parallel implementation) and set their number automatically based on the available resources (network bandwidth and CPU power).
+
+- **Custom**: Use parallel migration connections and use the number of connections specified by the user.  The minimum number of connections is 2.  There is no upper limit but it is recommended not to use more than 16 and the actual number of connections used for a particular migration is restricted to the number of threads on the source and destination hosts (whichever is lower).
+
+- **Cluster value** (only in the VM edit dialog; it's the default there): Use parallel connections as specified in the VM's cluster settings. 
+
+We use **Disabled** as the cluster default because it's a newly introduced feature in oVirt and there is not enough experience with it.  We may consider changing it to **Auto** or **Parallel** in next releases, especially if QEMU decides using parallel migration connections by default; this would be limited to a future cluster level.
+
+If **Parallel** is selected then the number of parallel connections is set to the lower number of the currently available migration network bandwidth divided by 10 Gbps and the number of VM vCPUs, but no less than 2 and no more than a configurable upper limit, which will be 16 by default.  **Auto** is the same except that parallel connection are not used if the computed number of connections is less than 2.
+
+Since parallel migration connections cannot be currently used with post-copy in QEMU, **Parallel** and **Custom** cannot be set if post-copy migration policy is set, or vice versa post-copy migration policy cannot be set if **Parallel** or **Custom** migration options are set (this must be checked for all the VMs in the cluster on cluster migration policy changes).
+
+If parallel migration connections are enabled, the following adjustments are made when a migration is initiated:
+
+- Compression is disabled regardless of the selected migration policy.  (Current libvirt apparently doesn't support using gzip/zstd compression, which could be helpful and it might be a good idea to always enable it.)
+
+- The maximum number of outgoing migrations is set to 1 regardless of the selected migration policy.  This may not be optimal if there is a non-parallel migration running that cannot saturate the available bandwidth but even in such a case it's generally simpler and safer not to start competing for the bandwidth.
+
+- If **Auto** is set and post-copy migration policy is selected, parallel migration connections are disabled. 
+
+## Testing
+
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+- There is very little documentation about parallel migration connections and the information above was obtained directly from platform developers.
+
+## User experience
+
+A new migration option to enable parallel migration connections is introduced in cluster and VM edit dialogs.  The cluster option specifies what to use in the given cluster.  VMs can override the cluster value individually.  The new option offers the following choices:
+
+- **Disabled** (the default in clusters): Parallel migration connections are not used.
+
+- **Auto**: Use parallel migration connections (TODO: Even when there is insufficient network bandwidth, for the supposed benefit of a more efficient implementation?) and set their number automatically based on the available resources (network bandwidth and CPU power).
```
ahadas, line 60: about the TODO - good question, probably yes but I think it should appear in the document in a more general way (that implies parallel migration may also not be set)

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+- **Custom**: Use parallel migration connections and use the number of connections specified by the user.  The minimum number of connection is 2.  (TODO: Should we put any real limit on the maximum number?)
```
ahadas, line 62: if it's debatable whether more than 16 connections adds a real benefit then we can add a configurable limit of 16 but anyway, even when the user sets a custom number it may make sense to reduce it based on the source/destination hosts, no?

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+- **Cluster value** (only in the VM edit dialog; it's the default there): Use parallel connections as specified in the VM's cluster settings. 
+
+We use **Disabled** as the cluster default because it's a newly introduced feature in oVirt and there is not enough experience with it.  We may consider changing it to **Auto** in next releases, especially if QEMU decides using parallel migration connections by default.
```
ahadas, line 66: yes and we should also limit that to the new cluster level (that will be added!)

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+If **Auto** is selected then the number of parallel connections is set to the lower number of the currently available migration network bandwidth divided by 10 Gbps and the number of VM vCPUs, but no less than 2.
+
+If parallel migration connections are enabled, the following adjustments are made when a migration is initiated:
+
+- Compression is disabled regardless of the selected migration policy.  (TODO: Check whether we can enable gzip/zstd compression.  If we can then it may be a good idea to always enable the compression.)
+
+- The maximum number of outgoing migrations is set to 1 regardless of the selected migration policy.  This may not be optimal if there is a non-parallel migration running that cannot saturate the available bandwidth but even in such a case it's generally simpler and safer not to start competing for the bandwidth.
```
ahadas, line 74: you mean when the cluster's setting changes, right? and what will happen - the engine will issue a call to VDSM in order to update the number of outgoing migrations to 1?

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+- If post-copy migration policy is selected, non-parallel migration is used regardless the settings.  (TODO: This is not the best user experience, it would be better to warn about this in the edit dialogs.  But how to handle the case when the cluster migration policy is non-post-copy, parallel connections are disabled in the cluster and parallel connections are enabled in some VM, which is valid so far, and then the cluster migration policy is changed to post-copy, which is not valid for the given VM anymore?)
```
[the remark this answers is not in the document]
mz-pdm: We can check settings for all the VMs in the cluster. It's still a question whether forcing the user to disable contingent parallel connections for all the affected VMs is the best user experience. It would most likely not be a typical use case so perhaps it can be acceptable.

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Testing
+
+It should be tested that migrations with parallel connection work as described above.  It requires at least about 40 Gbps network bandwidth.  The selected migration parameters can be checked in Vdsm logs.  Network bandwidth utilized by the migration should be measured to check that the data is indeed flowing in the specified amounts approximately, i.e. it works as expected on the side of Vdsm, libvirt, and QEMU.
+
+## Changes to Engine and Vdsm
+
+``VM.migrate`` Vdsm API call gets an additional parameter ``parallel``.  If it is unset, parallel migration connections are not used.  Otherwise the specified number of connections, which must be at least 1, is used.
```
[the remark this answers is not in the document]
mz-pdm: User should be prevented from setting the value to 1 in Engine validation.

source/develop/release-management/features/virt/parallel-migration-connections.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Parallel migration connections haven't been tested extensively with connections faster than 100 Gbps.  Due to current multifd implementation, there is an increasing overhead when using many connections and it's not clear whether using more than 16 parallel connections adds a real benefit.
+
+- It still applies that large guests with terrabytes of RAM are difficult to migrate successfully.
+
+- Since it's easier to saturate network bandwidth with multiple connections, it's advisable to not migrate multiple VMs from a single host at the same time.
+
+- VM CPU pinning may restrict the number of CPUs available to the QEMU process and thus impose a limit on the meaningful number of parallel connections.
```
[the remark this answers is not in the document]
mz-pdm: It's not necessarily bad if pinned/dedicated CPUs are used for migration threads. This helps to speed up the migration and, unlike autoconverge applied in this case, it doesn't waste CPU capacity. However I'm not sure whether/what means we have to influence the placement of the migration threads, I must look at it.

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: OK, I'll list both the options here (always parallel x only if there are sufficient resources).

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: Do you mean the custom number of connections specified by the user would effectively set the maximum rather than a fixed number of connections?

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: OK.

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: I supposed it could be done in a similar way as with migration policies where we have different outgoing migrations limits for different policies. It's a migration parameter in the Vdsm API so no extra API call should be needed.

source/develop/release-management/features/virt/parallel-migration-connections.md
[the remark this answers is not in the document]
ahadas: ack

source/develop/release-management/features/virt/parallel-migration-connections.md
ahadas: yeah, we have that heuristic of 10 Gbps per connection - I think that if the user specifies the number of connections, then we should not change it based on this heuristic but when it comes to CPU power, I think it never makes sense to use the specified number of connections when it exceeds the number of threads on the host(s), no?

source/develop/release-management/features/virt/parallel-migration-connections.md
ahadas: ah ok, ack

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: That's true. OK, I'll accommodate this observation here some way.

source/develop/release-management/features/virt/parallel-migration-connections.md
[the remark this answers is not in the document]
nyoxi: Just to clarify, will the intention be to use the VM CPUs instead of reserved or shared CPUs? The would certainly speed the convergence up but at the expense of degraded performance of the migrated VM during the process. Bear in mind that we are talking about VMs tuned for performance. Still, maybe that is something users are willing to accept during migration?

A consequence of this would be that you are also limited by number of CPUs of the VM in cases it has less than 16. An option would be to use combination of VMs CPUs and a shared pool. But using a shared pool complicates things quite a bit because we need to adapt the VM's configuration every time the shared pool changes. That being said...

Of course the above is not directly applicable to VMs without policy that run on a shared pool. And we probably should not allow migration of VM to affect performance of other VMs with dedicated CPUs (or pinning). That means VMs running in shared pool should use only CPUs in shared pool for migration. So it seems we wont be spared of adapting to shared pool changes during migration anyway.

It seems to me like a can of worms. It should be checked what is the real CPU consumption of the migration threads -- maybe all this is not needed and it would be best to leave the CPU management to the host instead of trying to manage this.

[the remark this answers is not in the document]
ahadas: > Just to clarify, will the intention be to use the VM CPUs instead of reserved or shared CPUs?

That's a good question, yeah, we need to see what means we have to influence the placement of the migration threads as Milan commented above and then see what we can do - reading my comment above, I thought about the latter (using reserved/shared CPUs) but that's not necessarily right (as you wrote) or possible.

> A consequence of this would be that you are also limited by number of CPUs of the VM in cases it has less than 16.

Yeah, this is taken into account in [calculateAutomaticParallelMigrations](https://gerrit.ovirt.org/c/ovirt-engine/+/116775/15/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java#563)

source/develop/release-management/features/virt/parallel-migration-connections.md
ahadas: yes, it blurs the difference between those options a bit but they will differ in case the number of vCPUs is greater than the amount of connections specified by the user or computed by the engine, and the amount of connections specified by the is different than the amount of connections that are computed by the engine based on the network bandwidth. it can go in both directions - that the user specifies a number that is greater than what the engine thinks is optimal (based on 10Gbps per connection) or lower than that

source/develop/release-management/features/virt/parallel-migration-connections.md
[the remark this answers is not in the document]
mz-pdm: My current intention is to leave the selection of CPUs to be used for the migration threads to the host. We don't want to pin the emulator threads (which I think are used for migrations) just because of migrations (a different thing would be if e.g. VMs with dedicated CPUs wanted to pin the emulator threads, but even in such a case this would just limit the maximum number of meaningful migration connections) and to complicate things without a good reason.

As for running migration threads on VM CPUs: If the VM wants to prevent that (e.g. because of dedicated CPUs), it should arrange the emulator thread pinning itself. It's true that migration threads, which can consume a lot of CPU power if the right number of threads is selected, can slow down dedicated CPUs if they land on them. But take into account that all our current migration policies use autoconverge, which slows down VM CPUs artificially if the migration doesn't converge. The difference is that migration threads can slow down CPUs since the beginning of the migration. But unless we want to disable autoconverge, which might prevent some VMs from finishing their migrations, VMs cannot expect having full CPU power during migrations even when they have dedicated or pinned CPUs.

Yes, another problem is if migration threads use CPUs dedicated to a different VM than the migrating one. We have two reasonable options here: to rely on the OS scheduling to do something reasonable or to manage emulator thread pinning. But again, if we want to do the latter then it should be a part of the dedicated CPUs functionality, since other QEMU stuff can possibly run on the dedicated CPUs, including current migrations.

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: Or to say it more simply to see the difference: If the user says "use 8 connections" then 8 connections will be used (unless it migrates between hosts that don't have that many threads, which is unlikely). With auto, it can be anything between 1 and the upper limit.

source/develop/release-management/features/virt/parallel-migration-connections.md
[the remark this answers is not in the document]
mz-pdm: OK, done.

source/develop/release-management/features/virt/parallel-migration-connections.md
mz-pdm: Done.

source/develop/release-management/features/virt/parallel-migration-connections.md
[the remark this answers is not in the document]
ahadas: With the addition to the dedicated CPU page, I think that's fair enough

ahadas (approved):

nyoxi (approved):

ahadas:
@sandrobonazzola can we get this one in please?

ahadas (approved):

sandrobonazzola (approved):

PR #2630, opened 2021-11-16
Dedicated cpu feature page udpate

nyoxi:

* CPU hotplugging
* storing pinned CPUs in metadata
* mention emulator/IO thread pinning in relation to parallel migrations

This pull request needs review by: @mz-pdm, @ahadas, @liranr23

### DIFF
--- a/source/develop/release-management/features/virt/dedicated-cpu.md
+++ b/source/develop/release-management/features/virt/dedicated-cpu.md
@@ -226,3 +226,12 @@ you won't be able to start the VM without removing VMs from one of the sockets
 firsts. Techniques for re-assigning CPUs for consolidating resources need to
 be investigated. It may (or may not) turn out to be a prerequisite for useful
 CPU hot-plugging too.
+
+At the moment we generally do not pin the emulator thread (`emulatorpin`) nor
+do we pin IO threads (`iothreadpin`) of the VMs. The only exception is high
+performance VMs where these threads are pinned to first core of the socket on
+which the VM runs. This however may change in the future in respect to another
+feature,
+[parallel migration connections](/develop/release-management/features/virt/parallel-migration-connections.html).
+In all cases Engine has to take care that CPUs used for emulator threads or IO
+threads are not dedicated to a VM. Such CPUs have to remain in shared pool.
--- a/source/develop/release-management/features/virt/dedicated-cpu.md
+++ b/source/develop/release-management/features/virt/dedicated-cpu.md
@@ -163,7 +163,7 @@ Further, Engine will add a new entry into VM metadata called `cpuPolicy` that
 will contain the requested CPU policy. The valid values for `cpuPolicy` are:
 
   * `none`: no pinning or policy selected, this is regular VM with shared CPUs
-  * `pin`: VM has fixed CPU pinning (CPU pinning or NUMA auto pinning)
+  * `manual`: VM has fixed CPU pinning (CPU pinning or NUMA auto pinning)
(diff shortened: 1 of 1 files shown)


nyoxi:
CI fails because #2622 is not merged yet.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -183,6 +194,15 @@ containing `<cputune>` (with no arguments) with `<vcpupin>` elements. E.g.:
 </cputune>
 ```
 
+API call `VM.setNumberOfCpus` will be extended with an optional field
+`cpusets`. It is a list CPU lists where each item of the list describes pinning
+for the vCPU at that index. For example `['1','3','5']` means that vCPU `1`
+should be pinned to to pCPU `1`, vCPU `2` should be pinned to pCPU `3` and vCPU
```
mz-pdm, line 200: ... pinned to pCPU ...

mz-pdm (commented):
Looks good, just one typo.

source/develop/release-management/features/virt/dedicated-cpu.md
```diff
@@ -183,6 +194,15 @@ containing `<cputune>` (with no arguments) with `<vcpupin>` elements. E.g.:
...
+`3` should be pinned to pCPU `5`. While the field itself is technically
+optional it is required for VMs with policy other than `shared` or `pin`. When
+specified, length of the list must match the new number of CPUs (i.e. must
+match `numberOfCpus` field).
```
ahadas, line 204: it this a new representation that is different than that we use when starting/migrating a vm?

source/develop/release-management/features/virt/dedicated-cpu.md
ahadas: yeah, double 'to'

source/develop/release-management/features/virt/dedicated-cpu.md
nyoxi: Unfortunately yes. We may reconsider how we pass information for migration and use the same approach as we do here (instead of passing XML).

mz-pdm: Yes, this would be better. It's a bit unfortunate to use a different representation on VM startup and then on migrations and CPU hot plugs. But using XML for hot plugs would be weird and using different representations for migrations and CPU hot plugs would be messy and confusing.

ahadas: we do that (using xml) for hot-plug disk and hot-plug NIC - it's true that we don't have a definition of 'device' in this case but if we now pass the same content we have in the domain xml, what prevents us from using the same representation we have in the domain xml?

mz-pdm: Formally nothing. But the difference is that in case of devices, we can (at least in theory) to pass the provided XML snippet directly to libvirt. In case of pinning and CPU hot plug, the provided XML cannot be used, it must be parsed and the CPUs are pinned in individual libvirt calls using numeric arguments. That means we create and parse XML just for the sake of having XML in the API. It may be good for consistency, but it wastes resources and doesn't match the idea of direct communication with libvirt using domain XML snippets. So what do you think considering that, should we use XML here or not?

nyoxi:
please review

liranr23 (approved):

ahadas (approved):

sandrobonazzola:
Merging based on @ahadas approval

PR #2642, opened 2021-11-29
Push into dashpanel branch

sanjacodes:
Fix Dashpanel Action to push into dashpanel branch instead of protected main branch


janosdebugs (approved):
LGTM

PR #2661, opened 2021-12-12
BZ#2025298 add note to DWH guide under Enabling Debug Mode

emarcusRH:
Fixes issue #2025298 

- add note to DWH guide under Enabling Debug Mode

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): (please @emarcusRH )

This pull request needs review by: (please @avlitman )

### DIFF
--- a/source/documentation/data_warehouse_guide/topics/Enabling_debug_mode.adoc
+++ b/source/documentation/data_warehouse_guide/topics/Enabling_debug_mode.adoc
@@ -23,9 +23,11 @@ DWH_AGGREGATION_DEBUG=true
 ====
 The move to daylight savings time/summer time in different time zones can cause the events log to report that the _ETL service aggregation to hourly tables has encountered an error_.
 If all the records are in the hourly aggregation views, there is no issue and this warning can be ignored.
-If this warning continues to appear, contact your system administrator
+If this warning continues to appear,
+ifdef::ovirt-doc[]
+contact your system administrator.
+endif::ovirt-doc[]
 ifdef::rhv-doc[]
- or the the Red Hat Support Team
-endif::[]
-.
+contact the the Red Hat Support Team.
+endif::rhv-doc[]
 ====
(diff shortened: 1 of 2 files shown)


avlitman:
Hi Eli,
sorry but we will need to change the note location.
after consulting with Shirly we decided to add the note here: https://access.redhat.com/solutions/3338001
under "Root Cause":

It can be caused by the move to daylight saving time/summer time in different time zones can cause the events log to report "ETL service aggregation to hourly tables has encountered an error".
If all the records are in the hourly aggregation views: there is no issue, and this warning can be ignored.
If this warning continue to appear you should contact us.

I think we can delete "It hasn't been determined yet. The progress could be tracked in Bug 1540154 var_datetime in history_configuration is in the future" since the bug is closed as WONTFIX.

We decided that because we have a note in the data warehouse guide that recommend to use UTC time zone:
we would like to emphasize this note and make it an important one:
https://www.ovirt.org/documentation/data_warehouse_guide/index.html#Overview_of_Configuring_Data_Warehouse,
https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html-single/data_warehouse_guide/index#Overview_of_Configuring_Data_Warehouse

emarcusRH:
- I removed the note from the DWH guide, and made the UTC comment in the intro into an IMPORTANT note
- I also updated the KCS article with the note, here is the draft: https://access.redhat.com/node/3338001/draft



> Hi Eli, sorry but we will need to change the note location. after consulting with Shirly we decided to add the note here: https://access.redhat.com/solutions/3338001 under "Root Cause":
> 
> It can be caused by the move to daylight saving time/summer time in different time zones can cause the events log to report "ETL service aggregation to hourly tables has encountered an error". If all the records are in the hourly aggregation views: there is no issue, and this warning can be ignored. If this warning continue to appear you should contact us.
> 
> I think we can delete "It hasn't been determined yet. The progress could be tracked in Bug 1540154 var_datetime in history_configuration is in the future" since the bug is closed as WONTFIX.
> 
> We decided that because we have a note in the data warehouse guide that recommend to use UTC time zone: we would like to emphasize this note and make it an important one: https://www.ovirt.org/documentation/data_warehouse_guide/index.html#Overview_of_Configuring_Data_Warehouse, https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html-single/data_warehouse_guide/index#Overview_of_Configuring_Data_Warehouse

dcdacosta (approved):

PR #2669, opened 2021-12-20
Add Engine dedicated cpus feature page

saifabusaleh:
Add engine dedicated cpus feature page

This pull request needs review by: @ahadas @liranr23 @ljelinkova @nyoxi

### DIFF
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -36,27 +36,22 @@ Engine:
 1. Add the new policies to CPU Pinning Policy enum
 2. Save host capabilities CPU topology in vds_dynamic
 3. Hold hostToAvailableCpu data structure:
-
-    Map Host id -> availableCpus object
-
-    availableCpus:
-
-    Map socket-id -> List of availableCores Object 
-
-    Socket: socket nr
-
-    availableCoresObject:
-    
-    Contains list of available cpus in the core
+    it will be on this form Map<Integer,Map<Integer,Map<Integer,List<String>>>>
+    Host id -> Map of Socket id to Map of Core id to list of cpus
+   (assuming host id, core id, socket id is unique)
 
     **Save/Update cached data structure:**    
 
     Option 1:
     * On updateVDSStatisticsData update the data structure with host capabilities
+      * On stop host, the host should be removed, and all the VM pinnings to that host should be pinned to other hosts
+      * On start host, the host should be added
     * Poll every X seconds all VMs and find which VM’s have cpu pinning and update the shared CPU pinning data structure based on VM pinned CPUs and host capabilities
 
     Option 2:
     * On updateVDSStatisticsData update the data structure with host capabilities
+      * On stop host, the host should be removed, and all the VM pinnings to that host should be pinned to other hosts
+      * On start host, the host should be added
     * On every start/stop/migrate/… VM update the shared CPU pinning data structure
     With used pinning
 
@@ -66,7 +61,7 @@ Engine:
     * Will be calculated after engine started after getting host caps and VMs cpu pinnings
     
     Option2:
-    * Create table in DB and save/read it from DB
+    * Create table in DB and serialize/deserialize it from DB
 
 4. Scheduling
     1. Start VM:
@@ -143,5 +138,28 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
 ## Limitations
 - Should only be supported with 4.7 cluster level
 
-## TBD
-- Siblings and isolate-threads
+
+
+### Isolate threads/Siblings
+
+Since we already know the available host capabilities and the available CPUs
+
+in case a VM is set with Isolate threads or sibilings, we can remove the whole core from the available cpu data structure
+
+Example:
+
+Host A with 1 Socket, 2 Cores in socket, 2 CPUs in each core
+
+hostToAvailableCpu data structure:
+
+{Host A id} -> {socket id} -> {core 1 id} -> {1,2}
+{Host A id} -> {socket id} -> {core 2 id} -> {3,4}
+
+Lets assume we have 1 VM with 2 vCPUs with isolate threads policy
+
+the VM will use from cpu ids 1 and 3, and cpu ids 2 and 4 will be blocked
+
+hostToAvailableCpu data structure after this change will be empty, since there is no available shared CPUs 
+current CPU pinning for the VM will be 1#1 2#3
+
+once the VM got stopped, we will need to restore the core 1 and 2 CPU's (which we will find from the host topology)
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -8,7 +8,7 @@ Authors: sabusale
 
 ## Summary
 
-Once choosing dedicated CPU policy, The vCPUs will be strictly pinned to a set of host pCPUs
+Once choosing dedicated CPU policy, the vCPUs will be strictly pinned to a set of host pCPUs
 (similarly to static CPU pinning). The set of pCPUs will be chosen to match the required guest CPU topology.
 
 Engine sets the pinning automatically when starting the VM as opposed to the manual pinning that is
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -137,7 +137,7 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
 
 
 
-### Isolate threads/Siblings
+## Isolate threads/Siblings
 
 Since we already know the available host capabilities and the available CPUs
 
@@ -166,7 +166,7 @@ current CPU pinning for the VM will be 1#1 2#3
 once the VM got stopped, we will need to restore core 1 and 2 CPU's (which we will find from the host topology)
 meaning, restoring also CPUs 2 and 4 to hostToAvailableCpu.
 
-### TBD
+## TBD
 
 - If VM CPU pinning policy is isolate threads or sibilings, and we have two hosts, one with SMT enabled and the other
   without SMT, do we need to count the host without SMT for scheduling(considering it same as dedicated)
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -33,19 +33,21 @@ Engine:
 
 ## Backend
 
-1. Add dedicated to CPU Pinning Policy enum
-2. Save host capabilities CPU topology in VM Dynamic
+1. Add the new policies to CPU Pinning Policy enum
+2. Save host capabilities CPU topology in vds_dynamic
 3. Hold hostToAvailableCpu data structure:
 
-    Map Host id -> availableCpu object
+    Map Host id -> availableCpus object
 
-    availableCpu:
+    availableCpus:
 
-    Map socket-id -> Physical-cpus
+    Map socket-id -> List of availableCores Object 
 
     Socket: socket nr
 
-    Physical-cpus(List of String) e.g: 1,2,3
+    availableCoresObject:
+    
+    Contains list of available cpus in the core
 
     **Save/Update cached data structure:**    
 
@@ -112,17 +114,23 @@ Engine:
 ### Example
 2 hosts are running and reporting the following available CPU topology
 
-Host A: 1 Socket, 4 CPUs
 
-Host B: 2 Sockets, 1 CPU on each socket
+Host A: 1 Socket, 2 Cores 2 CPUs
+
+Host B: 2 Sockets, 1 Core on each socket, one CPU in each core
+
+
+
 
 VM A with dedicated cpu pinning is already running on host A and have 2 vCPUs
 
-User wants to start VM B that requires 2 CPUs and has dedicated cpu policy
+User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
 
-- hostToAvailableCpu will be:
-  {host A id} -> { {socket 1 id} -> {1,2}}
-  {host B id} -> { {socket 1 id} -> {3}, {socket 2 id} -> {4}}
+- hostToAvailableCpu will be is:
+  
+  {host A id} -> { {socket 1 id} -> {core 1 id} -> {1,2}}
+  
+  {host B id} -> { {socket 1 id} -> {core 1 id } -> {1}, {socket 2 id} -> {core 2 id} -> {2}}
 - Scheduler ranks all hosts that have enough physical CPUs available based on hostToAvailableCpu. In our case, both
   hosts have enough pCPUs, however, host A has both available on one socket whereas the B would need to divide them on 
   2 different sockets. So A gets higher rank than B (B should be filtered)
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -62,10 +62,11 @@ Engine:
               - vcores that are part of the same vsocket should be on the same psocket
               - vthreads that are part of the same vcore should be on the same pcore
             - If no host left after the filtering phase throw validation error that cpu pinning can't be done 
-            - Score all hosts to find best matching host with enough pCPUs (two vCpus in different vCores can be
-              scheduled on same pCore)
-              - Host with less amount of sockets and cores left which
-                at least match the virtual topology will get higher score, since its better to use it
+            - Score all hosts to find best matching host in most efficient way so minimal
+              physical resources is used, host will gain higher score if:
+              - two vCpus in different vCores can be scheduled on same pCore
+              - two vCores in different vSockets can be scheduled on same pSocket
+              - two vSockets can be scheduled on one pSocket
         - Run phase
             - Based on Step 3 -Save/Update cached data structure
             - Update VM CurrentCpuPinning DB field with the mapping
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -56,20 +56,24 @@ Engine:
 4. Scheduling
     1. Start VM:
         - Scheduling phase
-            - Reads hostToAvailableCpu from step 3
-            - Filter hosts that can't match virtual CPU topology
-            - Hosts that doesn't match this two rules will be filtered:
-              - vcores that are part of the same vsocket should be on the same psocket
-              - vthreads that are part of the same vcore should be on the same pcore
-            - If no host left after the filtering phase throw validation error that cpu pinning can't be done 
-            - Score all hosts to find best matching host in most efficient way so minimal
-              physical resources is used, host will gain higher score if:
-              - two vCpus in different vCores can be scheduled on same pCore
-              - two vCores in different vSockets can be scheduled on same pSocket
-              - two vSockets can be scheduled on one pSocket
-        - Run phase
-            - Based on Step 3 -Save/Update cached data structure
-            - Update VM CurrentCpuPinning DB field with the mapping
+            - If VM have {Dedicated, Isolate-threads, Sibilings} policy:
+                - Reads hostToAvailableCpu from step 3
+                - Filter hosts that can't match virtual CPU topology
+                - Hosts that doesn't match this two rules will be filtered:
+                - vcores that are part of the same vsocket should be on the same psocket
+                - vthreads that are part of the same vcore should be on the same pcore
+                - If no host left after the filtering phase throw validation error that cpu pinning can't be done
+                - Score all hosts to find best matching host in most efficient way so minimal
+                  physical resources is used, host will gain higher score if:
+                - two vCpus in different vCores can be scheduled on same pCore
+                - two vCores in different vSockets can be scheduled on same pSocket
+                - two vSockets can be scheduled on one pSocket
+            - Run phase
+                - Based on Step 3 -Save/Update cached data structure
+                - Update VM CurrentCpuPinning DB field with the mapping
+           - If VM have other policies:
+                - Reads hostToAvailableCpu from step 3
+                - Filter hosts that have 0 pCpus ( all dedicated )
 
     2. Stop/Destroy VM:
         - Clear VM dynamic CPU pinning field
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -8,7 +8,7 @@ Authors: sabusale
 
 ## Summary
 
-Once choosing dedicated CPU policy, The guest vCPUs will be strictly pinned to a set of host pCPUs
+Once choosing dedicated CPU policy, The vCPUs will be strictly pinned to a set of host pCPUs
 (similarly to static CPU pinning). The set of pCPUs will be chosen to match the required guest CPU topology.
 
 Engine sets the pinning automatically when starting the VM as opposed to the manual pinning that is
@@ -43,30 +43,29 @@ Engine:
 
     **Save/Update cached data structure:**    
 
-    Option 1:
-    * On updateVDSStatisticsData update the data structure with host capabilities
-    * Poll every X seconds all VMs and find which VM’s have cpu pinning and update the shared CPU pinning data structure based on VM pinned CPUs and host capabilities
-
-    Option 2:
     * On updateVDSStatisticsData update the data structure with host capabilities
     * On every start/stop/migrate/… VM update the shared CPU pinning data structure
     with used pinning
+    * In case manual pinned cpus exist, they should be subtracted from hostToAvailableCpu 
 
    **Init cached data structure (on engine start/stop):**
-   
-    Option1:
+
     * Will be calculated after engine started after getting host caps and VMs cpu pinnings
     
-    Option2:
-    * Create table in DB and serialize/deserialize it from DB
 
 4. Scheduling
     1. Start VM:
         - Scheduling phase
             - Reads hostToAvailableCpu from step 3
-            - Filter hosts that can't match virtual CPU topology (pCpu pinning should
-              be close to the virtual topology, e.g: 2 vCPU on same vSocket should be placed on same pSocket)
-            - Score all hosts to find best matching host with enough pCPUs 
+            - Filter hosts that can't match virtual CPU topology
+            - Hosts that doesn't match this two rules will be filtered:
+              - vcores that are part of the same vsocket should be on the same psocket
+              - vthreads that are part of the same vcore should be on the same pcore
+            - If no host left after the filtering phase throw validation error that cpu pinning can't be done 
+            - Score all hosts to find best matching host with enough pCPUs (two vCpus in different vCores can be
+              scheduled on same pCore)
+              - Host with less amount of sockets and cores left which
+                at least match the virtual topology will get higher score, since its better to use it
         - Run phase
             - Based on Step 3 -Save/Update cached data structure
             - Update VM CurrentCpuPinning DB field with the mapping
@@ -90,7 +89,7 @@ Engine:
 ## Database
 
 * Save CPU mappings in field CurrentCpuPinning field in vm_dynamic table
-* Save CPU topology from host capabilities into vds_statistics
+* Save CPU topology from host capabilities into vds_dynamic
 * Save new pinning policies to DB using the existing CpuPinningPolicy field
 
 ## Rest API
@@ -123,9 +122,9 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
   {host A id} -> { {socket 1 id} -> {core 1 id} -> {1,2}}
   
   {host B id} -> { {socket 1 id} -> {core 1 id } -> {1}, {socket 2 id} -> {core 2 id} -> {2}}
-- Scheduler ranks all hosts that have enough physical CPUs available based on hostToAvailableCpu. In our case, both
-  hosts have enough pCPUs, however, host A has both available on one socket whereas the B would need to divide them on 
-  2 different sockets. So A gets higher rank than B (B should be filtered)
+- Scheduler checks all hosts that have enough physical CPUs available based on hostToAvailableCpu. In our case, both
+  hosts have enough pCPUs, however, host A has both available on one socket whereas host B would need to divide them on 
+  2 different sockets. So B get filtered
 - Host A is selected to run the VM
 - Engine creates the mapping between CPUs as: 1#1_2#2 and saves it to currentCpuPinning in vm_dynamic table for VM B entry
 - Engine sends the domain XML to VDSM with the required mapping
--- a/source/develop/release-management/features/virt/engine-dedicated-cpus.md
+++ b/source/develop/release-management/features/virt/engine-dedicated-cpus.md
@@ -1,0 +1,139 @@
+---
+Title: Engine Dedicated CPUs
+Category: feature
+Authors: sabusale
+---
+
+# Engine Dedicated CPUs
+
+## Summary
+
+Once choosing dedicated CPU policy, The guest vCPUs will be strictly pinned to a set of host pCPUs
+(similarly to static CPU pinning). The set of pCPUs will be chosen to match the required guest CPU topology.
+
+Engine sets the pinning automatically when starting the VM as opposed to the manual pinning that is
+done by the user and is static.
+
+## Owner
+
+-   Name: Saif Abu Saleh (sabusale)
+-   Email: sabusale@redhat.com
+
+## Responsibilities
+
+VDSM:
+
+- Makes sure that dedicated VMs vCPUs stay exclusively pinned to the given pCPU
+
+Engine:
+- Holds cached available cpus based on host capabilities and pinned CPUs
+- Schedule VM to run on best satisfying host based on the above cached data structure
+
+
+
+## Backend
+
+1. Add dedicated to CPU Pinning Policy enum
+2. Save host capabilities CPU topology in VM Dynamic
+3. Hold hostToAvailableCpu data structure:
+
+    Map Host id -> availableCpu object
+
+    availableCpu:
+
+    Map socket-id -> Physical-cpus
+
+    Socket: socket nr
+
+    Physical-cpus(List of String) e.g: 1,2,3
+
+    **Save/Update cached data structure:**    
+
+    Option 1:
+    * On updateVDSStatisticsData update the data structure with host capabilities
+    * Poll every X seconds all VMs and find which VM’s have cpu pinning and update the shared CPU pinning data structure based on VM pinned CPUs and host capabilities
+
+    Option 2:
+    * On updateVDSStatisticsData update the data structure with host capabilities
+    * On every start/stop/migrate/… VM update the shared CPU pinning data structure
+    With used pinning
+
+   **Init cached data structure (on engine start/stop):**
+   
+    Option1:
+    * Will be calculated after engine started after getting host caps and VMs cpu pinnings
+    
+    Option2:
+    * Create table in DB and save/read it from DB
+
+4. Scheduling
+    1. Start VM:
+        - Scheduling phase
+            - Reads hostToAvailableCpu from step 3
+            - Filter hosts that can't match virtual CPU topology (pCpu pinning should
+              be close to the virtual topology, e.g: 2 vCPU on same vSocket should be placed on same pSocket)
+            - Score all hosts to find best matching host with enough pCPUs 
+        - Run phase
+            - Based on Step 3 -Save/Update cached data structure
+            - Update VM CurrentCpuPinning DB field with the mapping
+
+    2. Stop/Destroy VM:
+        - Clear VM dynamic CPU pinning field
+        - Based on Step 3 -Save/Update cached data structure
+    
+3. Send vcpupin with cpusets and policy name in VM metadata
+   1. Send the policy name, available policy names: { none, pin, dedicated, siblings, isolate-threads }.
+   2. Send the cpusets (it will be stored in VM_Dynamic)
+
+4. Migrate VDSM call
+   1. Calculate vCPU to pCPU mapping based on the scheduled host
+   2. Send cpusets additional parameter
+   
+5. setNumberOfCpus VDSM call
+   1. Send cpusets additional parameter
+
+
+## Database
+
+* Save CPU mappings in field CurrentCpuPinning field in vm_dynamic table
+* Save CPU topology from host capabilities into vds_statistics
+* Save new pinning policies to DB using the existing CpuPinningPolicy field
+
+## Rest API
+
+1. Add new types of CPU pinning to api-model
+2. Update mapper with the new policies
+
+## Frontend
+
+- The added CPU pinning policy fields should be disabled on cluster level version < 4.7
+
+
+### Example
+2 hosts are running and reporting the following available CPU topology
+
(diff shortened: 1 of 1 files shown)


source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+1 Host with 8 pCPU
+And all VM’s scheduled to the same host
+
+* VM1 with dedicated CPU with 2 vCPU’s -> used_cpus: 2 current_pinning: 1#1-2#2  
+* VM2 with shared CPU with 2 vCPU’s -> used_cpus: 2 current_pinning null
+* VM1 destroyed -> used_cpus:  used_cpus:0
+
+
+#### Questions:
+* How to get to which actual pCPU number the vCPU’s got pinned?
```
liranr23, line 67: You need to decide based on the available pCPUs in the selected host. With the current description you made, you will need to pass on every VM that runs on the host with dedicated policy and start calculating which pCPU are free to use.

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+* What need to be checked in the whole virtual CPU topology?
```
liranr23, line 68: what do you mean?

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+* Add used_cpus field to the host to know how many CPUs is used
+
+
+
+#### Example:
+
+1 Host with 8 pCPU
+And all VM’s scheduled to the same host
+
+* VM1 with dedicated CPU with 2 vCPU’s -> used_cpus: 2 current_pinning: 1#1-2#2
```
liranr23, line 61: Note that dedicated has multiple options. Some might take "more" CPUs based on the chosen policy.

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Scheduling:
+
+* Find host that have the VM CPU topology on the scheduling phase
+* On run phase, schedule the VM to run on that host
+
+
+###Data Model:
+
+* Save CPU mappings in a field (maybe use CurrentCpuPinningField)?
```
liranr23, line 51: That's possible and once the VM goes down you will need to update the host(used_cpus).

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+* Add used_cpus field to the host to know how many CPUs is used
```
liranr23, line 52: To make it easier, used_cpus can be an array holding the pCPUs taken(to prevent later calculating of the cpu pinning string). Then you can filter easily by checking its length, while you can easily choose pCPU for your VM.

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+Frontend: Add Dedicated policy dropdown to the UI
+
+
+Steps 3-7: can be done in parallel by different person
+
+
+### Scheduling:
+
+* Find host that have the VM CPU topology on the scheduling phase
+* On run phase, schedule the VM to run on that host
```
liranr23, line 46: On run phase, create the VM CPU pinning on the selected host.

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+
+2- Scheduling
+
+3- Send vcpupin with cpusets and policy name in VM metadata(only when Policy is dedicated?)
+
+4- Send the policy name from vmStatic policy field
+
+5- Send the cpusets (convert from CurrentCpuPinningField form to array)
+
+6- API: Migrate
```
liranr23, line 32: On migration you will need to re-do steps 2-5. You need to find a host that has enough CPUs to use, and then you need to re-create the pinning (as it is possible that the new host freed CPUs are differ from the current used ones).

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+Send cpusets additional parameter (convert from CurrentCpuPinningField form to array)
+
+7- setNumberOfCpus VDSM call
+Send cpusets additional parameter (convert from CurrentCpuPinningField form to array)
+Frontend: Add Dedicated policy dropdown to the UI
```
liranr23, line 37: you suppose to add it into CpuPinningPolicy

liranr23 (commented):
Many details and answers can be seen in Tomas feature page, please check:
https://github.com/oVirt/ovirt-site/blob/f9199fc6f57fc8ac8c7172c6d25be70e06a43568/source/develop/release-management/features/virt/dedicated-cpu.md

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: Is the online cpus check is enough?

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: but how I know to which pCPU the vCPU got allocated on run phase?

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: what do you mean? dedicated CPU policy should not have options right?

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: why it need to be done in the migration? migration should call the scheduler to handle this right?

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: I changed the wording

source/develop/release-management/features/virt/dedicated-cpus.md
saifabusaleh: how it will make it easier?

I need mapping from vCPU to pCPU per VM anyway (to send the cpusets)

You mean I need it only to know which pCPU is currently available right?

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+### Tasks:
+
+
+1- Add dedicated to CPU Pinning Policy enum
+
+2- Scheduling
+
+3- Send vcpupin with cpusets and policy name in VM metadata(only when Policy is dedicated?)
+
+4- Send the policy name from vmStatic policy field
```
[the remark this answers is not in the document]
saifabusaleh: its not different, fixed the wording

source/develop/release-management/features/virt/dedicated-cpus.md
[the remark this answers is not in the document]
saifabusaleh: its not related, fixed the wording

source/develop/release-management/features/virt/dedicated-cpus.md
ljelinkova: You create the vCPU to pCPU mapping during the run phase based on the information sent from the host about available CPUs, its topology and the vm's cpu policy.

source/develop/release-management/features/virt/dedicated-cpus.md
ljelinkova: I think we should not keep the "used" CPUs but the "available" - as is done in the vdsm. In fact, I think the vdsm should report those available CPUs (called shared_pool I think) and we should work with that, not to keep our own copy of the data (except some kind of "pending" cpus as is used in scheduler even now)

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,69 @@
...
+Title: Engine Dedicated CPUs
+Category: feature
+Authors: sabusale
+---
+
+# Engine Dedicated CPUs
+
+## Summary
+
+Once choosing dedicated CPU policy, The guest vCPUs will be strictly pinned to a set of host pCPUs (similarly to static CPU pinning). The set of pCPUs will be chosen to match the required guest CPU topology.
```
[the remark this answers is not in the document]
saifabusaleh: Done

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,69 @@
...
+4. API: Migrate
+   1. Send cpusets additional parameter (convert from CurrentCpuPinning field form to array)
+   
+5. setNumberOfCpus VDSM call
+   1. Send cpusets additional parameter (convert from CurrentCpuPinning field form to array)
+
+
+### Data Model:
+
+* Save CPU mappings in a field (maybe use CurrentCpuPinning from vm_dynamic)?
```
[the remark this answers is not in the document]
saifabusaleh: I think I am missing few things:

If VDSM handles and have the mappings between vCpu and pCPU:
- why the engine need to send the mapping (vcpupin field in domain XML?)
- why the engine need to send the mappings in VM.migrate ?
- I may be missing something but I don't see in the VDSM that the mapping between vCPU and pCPU is exposed to the engine

The CPU topology doesn't really show the mapping
pinned_cpus property is not really exposed

So how the engine should get the mapping from VDSM?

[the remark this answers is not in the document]
ljelinkova: I think the flows goes like this:
1. 2 VDSMs are running and reporting the following CPUs are shared/available/free

- VDSM A:  CPUs 1, 2 on socket 1 and 3, 4 on socket 2
- VDSM B:  CPUs 1 on socket 1 and 4 on socket 2

2. User wants to start VM that requires 2 CPUs and has dedicated cpu policy
3. Scheduller ranks all hosts that have enough physical CPUs available using the selected cpu policy. In our case, both hosts have enough pCPUs, however, the A has both available on one socket whereas the B would need to divide them on 2 different sockets. So A gets higher rank than B
4. Host A is selected to run the VM
5. Engine creates the mapping between CPUs as: 1#1_2#2
6. Engine sends the domain XML to VDSM with the required mapping
7. VDSM updates its shared cpu pool and removes the mapped cpus (possibly with some more based on the selected policy)
8. VDSM runs the VM with pinned CPUs
9. VDSM A reports that only cpus 3,4 are in a shared pool/available

It is the engine that contains all the logic how the cpus should be pinned so that it can select the best host from all available hosts. VDSM only keeps the list of how its resources (CPUs) are utilized.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+## Owner
+
+-   Name: Saif Abu Saleh (sabusale)
+-   Email: sabusale@redhat.com
+
+## Responsibilities
+
+VDSM:
+
+- Manages the available pinned CPUs
```
ahadas, line 26: what does it mean `Manages the available pinned CPUs`?
I'd write it differently, when looking at the engine and vdsm parts in the process of running vms with dedicated cpus then vdsm makes sure that exclusive pinning is preserved by modifying other vms that are not set with dedicated cpus
for example, let's say that I have a VM running on the host with vCPU X that runs on pCPU 1 and the engine decides to pin vCPU Y of another VM with dedicated CPUs to pCPU 1, then the engine specifies that Y should be pinned exclusively to pCPU 1 and VDSM makes sure that X will not run on pCPU 1

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+
+Engine:
+- Schedule VM to run on best satisifing host
+- creates vCpu to pCpu mappings
```
ahadas, line 30: also here, `vCPU to pCPU mapping` is too general - it's more complicated than that, it's something like: engine can specify non-exclusive pinning specified by the user or exclusive pinning that is automatically computed
so it comes in the form of pinning rules (that mapping) and a pinning policy

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Backend
+
+1. Add dedicated to CPU Pinning Policy enum
+
+2.  Scheduling
+    1. Start VM:
+        - Scheduling phase
+            - Score all hosts to find best matching host with enough pCPUs
```
ahadas, line 40: here we need to be careful - as long as we relied on the OS scheduler to place the vCPUs, we didn't care much about the topology, however, here we are going to have some heuristic to pin the vCPUs so we really need to decide on the scoring

for instance, it's not good to place two vCores that are part of the same vSocket on two pCores that belong to different pSockets - it may be better to filter out such a host if that's the only option we have

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+        - Run phase
+            - Create vCpu to pCpu mapping based on the host free pCPUs reported
```
ahadas, line 42: `based on the host free pCPUs reported` - not sure what you mean by `free pCPUs`, whether it means that it is not exclusively pinned or that no vCPU runs on it - the latter is not useful at all since it may change right after the report
so if we're only interested in the former, we don't have to rely on what the host report - that's information the engine produces (besides the "reserved cpus" that need to be taken into account)

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+            - Update VM CurrentCpuPinning DB field with the mapping
+
+    2. Stop/Destroy VM:
+        - Clear VM dynamic CPU pinning field
```
ahadas, line 46: and the representation the engine should have per-host - we can't restore the pinning state on the host on run-vm, that would be too much of an overhead

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+3. Send vcpupin with cpusets and policy name in VM metadata
+   1. Send the policy name, available policy names: { none, pin, dedicated, siblings, isolate-threads }.
+   2. Send the cpusets (it will be stored in VM_Dynamic)
+
+4. Migrate VDSM call
+   1. Calculate vCPU to pCPU mapping based on the scheduled host
+   2. Send cpusets additional parameter
+   
+5. setNumberOfCpus VDSM call
+   1. Send cpusets additional parameter
```
ahadas, line 57: of all CPUs, not just the plugged ones
need to see what to do with unplugged ones - I guess we unplug the vCPUs with the highest indexes but I'm not sure

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Database
+
+* Save CPU mappings in field CurrentCpuPinning field in vm_dynamic table
```
ahadas, line 62: and the new pinning policies (using the existing field)

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Example
+2 hosts are running and reporting the following shared CPUs
+
+host A: CPUs 1, 2 on socket 1 and 3, 4 on socket 2
+
+host B: CPUs 1 on socket 1 and 4 on socket 2
```
ahadas, line 69: I don't see why we care about this information from the host
let's take host B for instance,
assuming pCPU 0 is reported as a `reserved cpu` and we know that pCPU 2, 3, 5-7 are exclusively pinned to some vCPUs, the engine can know that pCPUs 1 and 4 are not exclusively pinned - the engine would anyway need to know which vCPU is pinned to which pCPU so with that information, together with the reserved cpu, it has all the data to have a clear picture of what happens on the host, right?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+
+User wants to start VM that requires 2 CPUs and has dedicated cpu policy
+
+- Scheduler ranks all hosts that have enough physical CPUs available using the selected cpu policy. In our case, both 
+  hosts have enough pCPUs, however, the A has both available on one socket whereas the B would need to divide them on 
+  2 different sockets. So A gets higher rank than B
```
ahadas, line 75: ok, so that's the example I wrote before - in this case we should consider filtering out this host

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,102 @@
...
+1. Add new types of CPU pinning to api-model
+2. Update mapper with the new policies
+
+## Frontend
+
+- The added CPU pinning policy fields should be disabled on cluster level version < 4.7
+
+## Dependencies
+
+- VDSM should report current shared CPUs list to the engine(still not implemented)
```
ahadas, line 95: not sure why it's needed

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+Engine:
+- Holds cached available cpus based on host capabilities and pinned CPUs
+- Schedule VM to run on best satisfying host based on the above cached data structure
+
+
+
+## Backend
+
+1. Add dedicated to CPU Pinning Policy enum
+2. Save host capabilities CPU topology in VM Dynamic
```
liranr23, line 37: I don't get this one, you want to save host data in VM table?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+
+    Map Host id -> availableCpu object
+
+    availableCpu:
+
+    Map socket-id -> Physical-cpus
+
+    Socket: socket nr
+
+    Physical-cpus(List of String) e.g: 1,2,3
```
liranr23, line 48: Any additional information about cores?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+    2. Stop/Destroy VM:
+        - Clear VM dynamic CPU pinning field
+        - Based on Step 3 -Save/Update cached data structure
+    
+3. Send vcpupin with cpusets and policy name in VM metadata
+   1. Send the policy name, available policy names: { none, pin, dedicated, siblings, isolate-threads }.
+   2. Send the cpusets (it will be stored in VM_Dynamic)
+
+4. Migrate VDSM call
+   1. Calculate vCPU to pCPU mapping based on the scheduled host
```
liranr23, line 89: You also need to filter hosts here. The question will be what would be the indicator of can be scheduled on or not.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+        - Run phase
+            - Based on Step 3 -Save/Update cached data structure
+            - Update VM CurrentCpuPinning DB field with the mapping
+
+    2. Stop/Destroy VM:
+        - Clear VM dynamic CPU pinning field
+        - Based on Step 3 -Save/Update cached data structure
+    
+3. Send vcpupin with cpusets and policy name in VM metadata
+   1. Send the policy name, available policy names: { none, pin, dedicated, siblings, isolate-threads }.
```
liranr23, line 85: siblings and isolate-threads are core related, although nothing in the feature page touches it.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+### Example
+2 hosts are running and reporting the following available CPU topology
+
+Host A: 1 Socket, 4 CPUs
+
+Host B: 2 Sockets, 1 CPU on each socket
+
+VM A with dedicated cpu pinning is already running on host A and have 2 vCPUs
+
+User wants to start VM B that requires 2 CPUs and has dedicated cpu policy
```
liranr23, line 121: vCPUs

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,139 @@
...
+
+- hostToAvailableCpu will be:
+  {host A id} -> { {socket 1 id} -> {1,2}}
+  {host B id} -> { {socket 1 id} -> {3}, {socket 2 id} -> {4}}
+- Scheduler ranks all hosts that have enough physical CPUs available based on hostToAvailableCpu. In our case, both
+  hosts have enough pCPUs, however, host A has both available on one socket whereas the B would need to divide them on 
+  2 different sockets. So A gets higher rank than B (B should be filtered)
```
liranr23, line 128: what if the VM is set to 2 sockets as well? the host will be filtered out?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
liranr23: OK, I see it's TBD although mentioned here. A bit confusing.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Fixed

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: if the VM with cpu pinning policy
Hosts should be filtered/scored similar to Backend#Scheduling (see in code MigrateVmCommand#getVdsToRunOn )

 I think it should be similar to auto pinning/manual pinning policies, but based on other scheduling policy, right?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: we should try to see if there is a host that fit the virtual topology 


if there is host with >2 sockets but we can't manage to fit the virtual socket cpus to the physical one(meaning cpus will need to run on separate sockets), then it will be filtered

if  there is host that we can fit then it will not be filtered

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Done

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: This can be more complicated than it seems:

- what if the VM is dedicated and added CPU's doesn't fit based on the policy?

- un-plug with the new policies, we need to see what to do

optimally we would want to check where there is cores that is used by that VM that we can free, so for example we don't hold core with 1 CPU dedicated (in this case other VMs is blocked for siblings/isolate threads policy on that core)

(maybe same for sockets?)

source/develop/release-management/features/virt/engine-dedicated-cpus.md
liranr23: Yes.

source/develop/release-management/features/virt/dedicated-cpus.md
liranr23: Answered on later discussion.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Changed the wording

source/develop/release-management/features/virt/engine-dedicated-cpus.md
liranr23: My question was about this situation:
If i have a VM with 2 sockets, 1 core, 1 thread.
I have a host A with 1 socket, 1 core, 2 threads.
I have a host B with 2 sockets, 1 core, 1 thread.

Which one will you choose?
Or better asking, will you prefer to fit the virtual topology of the VM to the physical? Or rather, looking onto the physical to run on the same socket?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -36,27 +36,22 @@ Engine:
...
-    Map socket-id -> List of availableCores Object 
-
-    Socket: socket nr
-
-    availableCoresObject:
-    
-    Contains list of available cpus in the core
+    it will be on this form Map<Integer,Map<Integer,Map<Integer,List<String>>>>
+    Host id -> Map of Socket id to Map of Core id to list of cpus
+   (assuming host id, core id, socket id is unique)
```
liranr23, line 41: space

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -36,27 +36,22 @@ Engine:
...
 
     **Save/Update cached data structure:**    
 
     Option 1:
     * On updateVDSStatisticsData update the data structure with host capabilities
+      * On stop host, the host should be removed, and all the VM pinnings to that host should be pinned to other hosts
+      * On start host, the host should be added
```
liranr23, line 48: what does stop/start mean? stop- maintenance? removed from the cluster? start - host in UP status?

do we really need this information? If a host becomes irrelevant the VMs will migrate away, and if they have a CPU pinning policy they go to your point of migration in this doc.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -36,27 +36,22 @@ Engine:
...
     * Poll every X seconds all VMs and find which VM’s have cpu pinning and update the shared CPU pinning data structure based on VM pinned CPUs and host capabilities
 
     Option 2:
     * On updateVDSStatisticsData update the data structure with host capabilities
+      * On stop host, the host should be removed, and all the VM pinnings to that host should be pinned to other hosts
+      * On start host, the host should be added
```
liranr23, line 54: same as above

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -143,5 +138,28 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
...
+Host A with 1 Socket, 2 Cores in socket, 2 CPUs in each core
+
+hostToAvailableCpu data structure:
+
+{Host A id} -> {socket id} -> {core 1 id} -> {1,2}
+{Host A id} -> {socket id} -> {core 2 id} -> {3,4}
+
+Lets assume we have 1 VM with 2 vCPUs with isolate threads policy
+
+the VM will use from cpu ids 1 and 3, and cpu ids 2 and 4 will be blocked
```
liranr23, line 160: The VM will use CPU IDs..

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -143,5 +138,28 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
...
+
+### Isolate threads/Siblings
+
+Since we already know the available host capabilities and the available CPUs
+
+in case a VM is set with Isolate threads or sibilings, we can remove the whole core from the available cpu data structure
+
+Example:
+
+Host A with 1 Socket, 2 Cores in socket, 2 CPUs in each core
```
liranr23, line 151: You can add here, that having 2 CPUs in a core indicates of a host with 2 threads topology

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -143,5 +138,28 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
...
+
+{Host A id} -> {socket id} -> {core 1 id} -> {1,2}
+{Host A id} -> {socket id} -> {core 2 id} -> {3,4}
+
+Lets assume we have 1 VM with 2 vCPUs with isolate threads policy
+
+the VM will use from cpu ids 1 and 3, and cpu ids 2 and 4 will be blocked
+
+hostToAvailableCpu data structure after this change will be empty, since there is no available shared CPUs 
+current CPU pinning for the VM will be 1#1 2#3
```
liranr23, line 163: If you really wish to be on point: 0#1, 1#3.
We count from 0, but as it is only example and the host example starting from 1, that's OK as well.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -143,5 +138,28 @@ User wants to start VM B that requires 2 vCPUs and has dedicated cpu policy
...
+
+once the VM got stopped, we will need to restore the core 1 and 2 CPU's (which we will find from the host topology)
```
liranr23, line 165: Once... to restore cores 1 and 2 CPU's...
You may add: meaning, restoring CPUs 2 and 4 to hostToAvailableCpu.

source/develop/release-management/features/virt/dedicated-cpus.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### Tasks:
+
+
+1- Add dedicated to CPU Pinning Policy enum
+
+2- Scheduling
+
+3- Send vcpupin with cpusets and policy name in VM metadata(only when Policy is dedicated?)
```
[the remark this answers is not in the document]
saifabusaleh: Updated about sending the policy

Yes, about `manuallyPinnedCpus` I think it should be separate work

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: so you are saying this will be handled by the scheduler and this shouldn't matter for us anyway, right?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: I think in this case we should fit the virtual topology

if we can't fit the virtual topology, we can decide between two options:

1- Try to fit as much as we can from virtual topology, for example try to fit at least 1 socket from virtual topology and and if the second socket can't fit, than try to fit CPU's in separate sockets( if there is still sockets available)

2- Throw validation error, since probably the end-user will not gain too much performance from the CPU pinning that he is planning to do

source/develop/release-management/features/virt/engine-dedicated-cpus.md
liranr23: Yes

source/develop/release-management/features/virt/engine-dedicated-cpus.md
liranr23: this is an open discussion, TBH i'm not sure what is preferable.
1. we can be close as much we can to the virtual topology - best to fit the user's configuration.
2. we can do it by physical to have better performance.
but, without dedicated policy having 2 sockets could be on same physical socket or elsewhere.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: keeping discussion in above thread

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Done, removed those lines as it should be obvious that it is handled by the scheduler

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+## Backend
+
+1. Add the new policies to CPU Pinning Policy enum
+2. Save host capabilities CPU topology in vds_dynamic
+3. Hold hostToAvailableCpu data structure:
+    it will be on this form `Map<Integer,Map<Integer,Map<Integer,List<String>>>>`
+   
+    Host id -> Map of Socket id to Map of Core id to list of cpus
```
ahadas, line 41: if we store it in a per-host entity like VdsManager then we can save that first mapping

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+    **Save/Update cached data structure:**    
+
+    Option 1:
+    * On updateVDSStatisticsData update the data structure with host capabilities
+    * Poll every X seconds all VMs and find which VM’s have cpu pinning and update the shared CPU pinning data structure based on VM pinned CPUs and host capabilities
+
+    Option 2:
+    * On updateVDSStatisticsData update the data structure with host capabilities
+    * On every start/stop/migrate/… VM update the shared CPU pinning data structure
+    with used pinning
```
ahadas, line 53: we anyway need to set the about-to-be-applied-pinning somewhere between the scheduling and the time the vm starts (like pending cpu and memory) so the 2nd option makes more sense

I didn't get the updateVDSStatisticsData part - what do we need to retrieve on stats cycles?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+
+   **Init cached data structure (on engine start/stop):**
+   
+    Option1:
+    * Will be calculated after engine started after getting host caps and VMs cpu pinnings
```
ahadas, line 58: so if the host is up then we have the data from caps updated already and we "just" need to inspect the vms that run on the host (inspect their dynamic information in the database), no?

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    Option2:
+    * Create table in DB and serialize/deserialize it from DB
```
ahadas, line 61: no need for storing the same data multiple times, let's drop this option

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+
+4. Scheduling
+    1. Start VM:
+        - Scheduling phase
+            - Reads hostToAvailableCpu from step 3
+            - Filter hosts that can't match virtual CPU topology (pCpu pinning should
+              be close to the virtual topology, e.g: 2 vCPU on same vSocket should be placed on same pSocket)
+            - Score all hosts to find best matching host with enough pCPUs 
+        - Run phase
+            - Based on Step 3 -Save/Update cached data structure
```
ahadas, line 71: I wrote about this in a previous comment - this needs to be similar to setting the pending cpu/memory resources

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ahadas: yes but let's filter out hosts only if we break one of two rules:
1. pinning vcores that are part of the same vsocket to the same psocket
2. pinning vthreads that are part of the same vcore to the same pcore

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ahadas: > This can be more complicated than it seems:
> 
>     * what if the VM is dedicated and added CPU's doesn't fit based on the policy?

so we shouldn't plug I guess

> 
>     * un-plug with the new policies, we need to see what to do
> 
> 
> optimally we would want to check where there is cores that is used by that VM that we can free, for example if we have 1 CPU dedicated on the whole core (in this case other VMs is blocked for siblings/isolate threads policy on that core) so we should free it if possible

yeah, but that's introduces extra complexity (also need to change the engine<->vdsm API) - doesn't seem to worth it

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+   2. Send cpusets additional parameter
+   
+5. setNumberOfCpus VDSM call
+   1. Send cpusets additional parameter
+
+
+## Database
+
+* Save CPU mappings in field CurrentCpuPinning field in vm_dynamic table
+* Save CPU topology from host capabilities into vds_statistics
```
ahadas, line 93: ah ok, so that's what you meant before about the stats cycles - if we're talking about capabilities then it should go to vds_dynamic instead

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ahadas: I wouldn't filter the hosts based on this computation but rather score them differently (e.g., the lower the amount of sockets + cores, the higher the score will be)

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: yes, or manual pinning settings in vm_static

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Updated the feature page

source/develop/release-management/features/virt/engine-dedicated-cpus.md
```diff
@@ -0,0 +1,173 @@
...
+   
+    Host id -> Map of Socket id to Map of Core id to list of cpus
+    (assuming host id, core id, socket id is unique)
+
+    **Save/Update cached data structure:**    
+
+    * On updateVDSStatisticsData update the data structure with host capabilities
+    * On every start/stop/migrate/… VM update the shared CPU pinning data structure
+    with used pinning
+    * In case manual pinned cpus exist, they should be subtracted from hostToAvailableCpu
```
liranr23, line 49: why? i think we should only subtract dedicated, manual is shared pool and VDSM should handle it AFAIK.

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ljelinkova: I would also prefer to use java classes instead of this Map structure - for better readability + it would allow to add also some functionality to the classes. E.g. VdsSocket -> VdsCore -> VdsCpu. Possible methods could be VdsCore:hasAvailableCpus, VdsCore:getAvailableCpus, VdsCpu:isAvailable, maybe even VdsCpu:getVm (to return VM that this CPU is dedicated to).

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ahadas: @liranr23 yes but note that VDSM doesn't evacuate vCPUs that were manually pinned
example:
we manually pin vCPU-0 to pCPU-2
we cannot exclusively pin another vCPU exclusively to pCPU-2 because vCPU-0 won't move elsewhere

so if `available` refers here to "available for exclusive pinning" then pCPU-2 should be subtracted as well

source/develop/release-management/features/virt/engine-dedicated-cpus.md
ahadas: right

source/develop/release-management/features/virt/engine-dedicated-cpus.md
saifabusaleh: Done

liranr23:
> @liranr23 do we still need this page? if we have a clear picture of what we're going to do and this adds an overhead to the development, we can drop this..

We can. The only benefit would be for others to see how it's done from engine side. If we decide to have it, I will need to update it.

ahadas:
no need to maintain another page that is specific to the engine, not at this point

PR #2731, opened 2022-02-16
BZ2054905 - Updated Copyright date to 2022 and included legal notices…

dcdacosta:
… where missing from some docs

Bug Fix

Fixes issue https://bugzilla.redhat.com/show_bug.cgi?id=2054905

- Updated the copyright in the legal notices from 2021 to 2022
- Included the legal notices in the Planning and Prerequisites and Intro to VM Portal Guides because it was missing.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/master/CONTRIBUTING.md): @dcdacosta 
This pull request needs review by: @apinnick

### DIFF
--- a/source/documentation/administration_guide/index.adoc
+++ b/source/documentation/administration_guide/index.adoc
@@ -29,4 +29,4 @@ include::appe-System_Accounts.adoc[]
 [appendix]
 == Legal notice
 
-Certain portions of this text first appeared in link:{URL_downstream_virt_product_docs}administration_guide/[Red Hat Virtualization {vernum_rhv} Administration Guide]. Copyright © 2021 Red Hat, Inc. Licensed under a link:https://creativecommons.org/licenses/by-sa/4.0/[Creative Commons Attribution-ShareAlike 4.0 Unported License].
+Certain portions of this text first appeared in link:{URL_downstream_virt_product_docs}administration_guide/[Red Hat Virtualization {vernum_rhv} Administration Guide]. Copyright © 2022 Red Hat, Inc. Licensed under a link:https://creativecommons.org/licenses/by-sa/4.0/[Creative Commons Attribution-ShareAlike 4.0 Unported License].
--- a/source/documentation/administration_guide/index.adoc
+++ b/source/documentation/administration_guide/index.adoc
@@ -2,6 +2,9 @@
 include::common/collateral_files/attributes.adoc[]
 = Administration Guide
 
+:context: admin_guide
+:admin_guide:
+
 // Make sure Jekyll displays a guide title
 [discrete]
 = Administration Guide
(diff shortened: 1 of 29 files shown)


sandrobonazzola (approved):
LGTM

apinnick:
@dcdacosta Before I merge, I want to check these legal notices against what appears in the customer portal. I thought most docs were not using their own notices because Legal needs to keep them up to date. I will get back to you about this.

apinnick:
@dcdacosta Hate to add this to your workload but there are some odd things happening with the downstream docs. Fortunately, now that all the docs are in GitHub, you can include them in a single PR.

First, the legal notice is OK. I see that OpenShift is using their own legal notice, which is very close to RHV's. 

However, the Upgrade guide had the legal notice before an appendix for "Additional resources", which looks very odd.

Several guides, checked at random downstream, do not have the legal notice at all. One of them was an installation guide. Could you please look at all the master.adoc files to make sure that each one has a legal notice included as the last element? Thanks.

dcdacosta:
> @dcdacosta
> 
> I hate to add this to your workload but there are some odd things happening with the downstream docs. Fortunately, now that all the docs are in GitHub, you can include them in a single PR.
> 
> First, the legal notice is OK. I see that OpenShift is using their own legal notice, which is very close to RHV's.
> 
> However, the Upgrade guide had the legal notice before an appendix for "Additional resources", which looks very odd.
> 
> Several guides, checked at random downstream, do not have the legal notice at all. One of them was an installation guide. Could you please look at all the master.adoc files to make sure that each one has a legal notice included as the last element?
> 
> Finally, each master.adoc should link to the single legal notice module. I see some assemblies have the legal notice hard-coded, which makes them difficult to maintain.
> 
> Thanks!

@apinnick I added the legal notices to the master files where it was missing and corrected the location of the Legal notices in the Upgrade Guide. 

The instances where the legal notice is hardcoded is in the index.adoc files, which I believe are used for upstream only. I was told not to change anything for those except the copyright date last year when I did this work for the migration to the upstream repo. So, the notice for upstream [index.adoc] is slightly different than for downstream [master.adoc].

apinnick:
Hi Donna, if they were upstream only, they should have their own module and be tagged with ifdefs for ovirt. Or else the existing legal module should be tagged for oVirt and RHV. I will take a look next week and get back to you. Thanks!

source/documentation/administration_guide/index.adoc
```diff
@@ -29,4 +29,4 @@ include::appe-System_Accounts.adoc[]
 [appendix]
 == Legal notice
```
apinnick, line 31: If you don't want to create a module, at the very least I would create a 'year' attribute so that no one has to hunt down and update all these occurrences.

PR #2771, opened 2022-03-10
redirect page for console-client-resources

sandrobonazzola:
[Bug fix]

console-client-resources page is linked from the engine and from several
BZs and sites. Providing a redirect to the dropped page.
Using 302 instead of 301 as there's not yet a final decision on where
this content should be located within the site.


didib (approved):
Is the "/dropped" prefix going to be permanent? Sounds like a rather temporary name to me... 301 reply implies it's intended to be permanent.

Perhaps add somewhere a note clarifying what it's used for etc.

Right now, I think its contents should be moved elsewhere, scattered around the site as applicable.

Perhaps rename to "/dropped-features" or "/dropped-from-main-documentation" or whatever, to better clarify its intention. Better decide now then later, considering 301/caching/search engines/etc.

PR #3025, opened 2022-07-16
Delete dashpanel workflow

sanjacodes:
The dashpanel workflow uses the [dashpanel](https://github.com/marketplace/actions/dashpanel) Action to get an [overview](https://github.com/oVirt/ovirt-site/blob/dashpanel/dashpanel-ovirt.md) of our repositories.

I don't think anyone is monitoring this, though, so let's remove the redundant code that uses computing resources without a clear use.

---

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/main/CONTRIBUTING.md): @sanjacodes 

This pull request needs review by: @sandrobonazzola


sandrobonazzola (changes requested):
If we drop the dashpanel we need to drop also links to it within the website. I think it's linked at least in a couple of places

sanjacodes:
Ah, you're right - I'll add that to the PR! Or do you think we should keep it?

sandrobonazzola (approved):
LGTM

Issue #3098, opened 2022-12-05
Install guide needs an update for oVirt Engine on EL9


sandrobonazzola:
#### Type of problem?
Documentation issue

Installation guide at https://ovirt.org/documentation/installing_ovirt_as_a_standalone_manager_with_local_databases/index.html#Installing_RHEL_for_RHVM_SM_localDB_deploy says:

> The oVirt Engine must run on Enterprise Linux 8. For detailed installation instructions, see ...

This is not true anymore, oVirt Engine can now run on Enterprise Linux 9 as well.

sandrobonazzola:
@emarcusRH , @apinnick any suggestion on how to phrase the change?
CC @mwperina

apinnick:
In theory, the version should be described in only one section, [Operating system requirements](https://ovirt.org/documentation/installing_ovirt_as_a_standalone_manager_with_local_databases/index.html#operating-system-requirements_SM_localDB_deploy). 

The other parts of the doc should not specify a specific version.

sandrobonazzola:
> In theory, the version should be described in only one section, [Operating system requirements](https://ovirt.org/documentation/installing_ovirt_as_a_standalone_manager_with_local_databases/index.html#operating-system-requirements_SM_localDB_deploy).
> 
> The other parts of the doc should not specify a specific version.

This section needs update as well as it requires EL 8.6 but oVirt 4.5.4 is now requiring EL 8.7 as minimum while RHV requires 8.6 EUS.

apinnick:
> > In theory, the version should be described in only one section, [Operating system requirements](https://ovirt.org/documentation/installing_ovirt_as_a_standalone_manager_with_local_databases/index.html#operating-system-requirements_SM_localDB_deploy).
> > The other parts of the doc should not specify a specific version.
> 
> This section needs update as well as it requires EL 8.6 but oVirt 4.5.4 is now requiring EL 8.7 as minimum while RHV requires 8.6 EUS.

I created a supported RHEL version attribute. It looks like I will need to create min/max supported version attributes.

sandrobonazzola:
> I created a supported RHEL version attribute. It looks like I will need to create min/max supported version attributes.

can't attribute contain a string? If it can I would use "8.6" for RHV and "8.7 or newer" for oVirt

apinnick:
> > I created a supported RHEL version attribute. It looks like I will need to create min/max supported version attributes.
> 
[...]

PR #3099, opened 2022-12-06
GH3098: Update supported RHEL version

apinnick:
Bug fix

Fixes issue # [3098](https://github.com/oVirt/ovirt-site/issues/3098)

- Add attribute for latest supported RHEL version so that it is not hard-coded into the docs

-

-

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/main/CONTRIBUTING.md): @apinnick 

This pull request needs review by: @sandrobonazzola

### DIFF
--- a/source/documentation/common/collateral_files/_attributes.adoc
+++ b/source/documentation/common/collateral_files/_attributes.adoc
@@ -90,6 +90,7 @@ ifdef::rhv-doc[]
 :org-fullname: Red Hat
 :container-platform: OpenShift Container Platform
 :container-platform-virt: OpenShift Virtualization
+:supported-rhel-version: 8.6
 endif::rhv-doc[]
 
 ifdef::ovirt-doc[]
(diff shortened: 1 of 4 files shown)


source/documentation/common/collateral_files/_attributes.adoc
```diff
@@ -108,4 +109,5 @@ ifdef::ovirt-doc[]
 :org-fullname: oVirt
 :container-platform: OKD
 :container-platform-virt: KubeVirt
+:supported-rhel-version: 9
```
sandrobonazzola, line 112: I would say 8.7 or newer

sandrobonazzola (changes requested):
See inline comment

emarcusRH:
For RHV 4.4 SP1, there will never be another version of RHEL supported - the last and latest is supposed to be RHEL 8.6 EUS, so introducing an attribute for the latest RHEL version here is not a good idea for any RHV downstream docs

apinnick:
> For RHV 4.4 SP1, there will never be another version of RHEL supported - the last and latest is supposed to be RHEL 8.6 EUS, so introducing an attribute for the latest RHEL version here is not a good idea for any RHV downstream docs

@emarcusRH See the GitHub issue. The problem is that Ovirt supports 8.7 and later. So we need an attribute.

sandrobonazzola (approved):
LGTM

PR #3102, opened 2022-12-12
Update CONTRIBUTING.md

sandrobonazzola:

- Add comment about using podman with selinux enforcing.

This pull request needs review by: @oVirt/developers

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,6 +102,11 @@ docker build -t ovirt-site .
 docker run -v $(pwd):/srv/site -p 4000:4000 -p 35729:35729 -ti ovirt-site
 ```
 
+Please note that on SELinux enforcing systems the last command should be:
+```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -102,6 +102,11 @@ docker build -t ovirt-site .
 docker run -v $(pwd):/srv/site -p 4000:4000 -p 35729:35729 -ti ovirt-site
 ```
 
+Please note that on SELinux enforcing systems the last command should be:
```
barpavel, line 105: LGTM, just 1 suggestion:
If I understand correctly the below command comes **instead** of the `docker run -v $(pwd):/srv/site -p 4000:4000 -p 35729:35729 -ti ovirt-site` above, so maybe better to emphasize that with something like:
`...the last command should be:`
-->
`...the last command above should change to:`

barpavel (approved):

CONTRIBUTING.md
sandrobonazzola: thanks, updated.

Issue #3081, opened 2022-11-10
Issue: /download/ Authors tag blocks the view


HubKing:
[ Thank you for submitting an issue! Please read these instructions and then delete the instructions sections before submitting the issue.

Is this Issue something you can help fix? If you'd like to contribute to making our site better, submit a Pull Request along with your Issue report! ]

### Description of the problem
Why is the big authors tag there? It blocks the page.

![image](https://user-images.githubusercontent.com/10900489/201160997-9b4a191f-339c-4587-a3c4-6e763791518a.png)

#### Type of problem?
Formatting issue 

[ Enter a detailed description of the problem. If you have suggestions for improvements, please include them (and open a Pull Request too, if you'd like!). Please include links to users list discussions, external documentation, or anything else that might be helpful. Delete this section before submitting your Issue. ]

At least add an [X] button to close it.

PR #3263, opened 2026-04-08
Avoid to hide content with authors box

sandrobonazzola:
Bug fix

Fixes issue #3081

Just a workaround using style to avoid authors box hiding content when not running full screen with recent monitors.

This pull request needs review by: @JasperB-TeamBlue

### DIFF
--- a/source/_sass/lib/site.sass
+++ b/source/_sass/lib/site.sass
@@ -140,6 +140,13 @@ header.post-header
   float: left
   clear: none
   width: min-content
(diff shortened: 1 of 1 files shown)


JasperB-TeamBlue (changes requested):
I don't feel like expanding the author box to fullscreen when it would layer over something is the right solution. I think either the text should bend more around it or the author box should be closable but having it be this large takes up too much space and does not look good in my opinion.

Image

sandrobonazzola:
Closing for now, I'll see if I'll have some free time over the weekend to look into a more structured solution.

PR #3264, opened 2026-04-21
Import ovirt-design documentation

sandrobonazzola:
Feature

Fixes issue https://github.com/oVirt/ovirt-design/issues/19

- Imported https://github.com/oVirt/ovirt-design as development documentation.
- Marked the imported files as licensed under Apache-2.0 keeping the original license as the oVirt site defaults to CC BY-SA 4.0.

This pull request needs review by: @JasperB-TeamBlue


JasperB-TeamBlue (changes requested):
Looks good, only issue i currently have with this PR is the way the breadcrumbs of the site pick up on the current structuring of the files. When for example you go to the upgrade cluster tab you see the following:

Image

The issue is with this that you then can return to Admin UI which is not really a tab and that is due to how the files are structured. In my opinion it would be better if these weird breadcrumbs to generated sub pages were not present.

JasperB-TeamBlue (changes requested):
It still seems to generate some weird ghost pages inbetween the actually created pages based on the breadcrumbs for example https://ovirt.github.io/ovirt-site/previews/3264/develop/design/user-portal/ shows a weird melt together of what is listed under `User Portal` but also `Patternfly 4/User Portal` and `User Research/User Portal` this feels like unwanted behaviour. Either there are more inbetween breadcrumbs added to correlate to the title dividers or the breadcrumbs stop at `Ovirt UI Design`

Issue #3114, opened 2023-01-25
gpg key list is no longer accurate


ykuksenko:
The gpg key list shown below is no longer accurate.
https://github.com/oVirt/ovirt-site/blob/eff20922e2b435157fb3f57cfab2140669b156d6/source/download/rpms_and_gpg.md?plain=1#L80-L86
Per https://github.com/oVirt/ovirt-release/issues/298 there is a new key out there. Please list this key. I think the key id is `44E475CB24901D0C`.

ykuksenko:
The gpg instructions are also out of date for AlmaLinux9 (rhel9 based distros?) now.

ctodea:
This issue is also preventing us from installing ovirt-engine-appliance-4.5-20221206125848.1.el9.x86_64.rpm

Nihvel:
I'm having this issue too, this is my temporary fix. 

Edit `/etc/yum.conf` and append:
`exclude=ovirt-engine-appliance-4.5-20221206125848.1.el9`

The playbook (and the user too with dnf/yum) can now install the previous engine appliance version.

Alternatively, one can pre-install the engine appliance with
`dnf install <http-link-to-engine-appliance.rpm>` and let the playbook skip that task.

Detailed steps:
* Do a `dnf install ovirt-engine-appliance`
* Notice the (as of 21st March, latest) version `4.5-20221206125848.1.el9`
* Try to install it. It fails due to GPG key
* Use a different version: `dnf install ovirt-engine-appliance-4.5-20221026100609.1.el9`
* The ovirt engine is installed

MaoliWolf:
With #3278 merged, both halves of this should now be addressed:

  - The new `24901D0C` key was added to the *Currently used keys* table in [`9163c1e2`](https://github.com/oVirt/ovirt-site/commit/9163c1e212) (2024-02-26), but the commit wasn't linked to this issue, which is likely why it stayed open.
  - The *Importing Keys Manually* section, which still pointed at the legacy `FE590CB7`, now uses the current oVirt 4.5 key via `centos-release-ovirt45` (#3278).

  Looks closable to me — @dupondje, any objection?

dupondje:
Agree. To make it close automatically, you need 'Fixes #id' in the commit msg :)

PR #3278, opened 2026-05-29
Update GPG key import docs for oVirt 4.5

MaoliWolf:
[Bug fix]

  Fixes issue #3114

  - The *Importing Keys Manually* section pointed at the legacy `FE590CB7`
    key (2048-bit, oVirt 4.4 and older). oVirt 4.5 packages are signed with
    `24901D0C` (4096-bit), so the example now imports that key from
    `https://resources.ovirt.org/pub/keys/RPM-GPG-ovirt-v4`.
  - Replaced the EOL `ovirt-release44.rpm` automatic-import step with
    `centos-release-ovirt45`, which ships the key at
    `/etc/pki/rpm-gpg/RPM-GPG-KEY-oVirt-4.5`.
  - Annotated the *Currently used keys* table to mark `FE590CB7` as legacy
    (oVirt 4.4 and older) and `24901D0C` as current (oVirt 4.5).

  Notes:

  - `24901D0C` is not published to public keyservers (checked the default
    keyserver and keys.openpgp.org), so the example downloads it from
    `resources.ovirt.org` instead of using `gpg --recv-keys`.
  - The `FE590CB7` expiry (2028-04-06) in the table is left unchanged — it
    is correct; the key's expiry was extended over time
    (`RPM-GPG-ovirt` 2016 → `-v2` 2021 → `-v3` 2028).

  I confirm that this pull request was submitted according to the [contribution
  guidelines](https://github.com/oVirt/ovirt-site/blob/main/CONTRIBUTING.md): @MaoliWolf

  This pull request needs review by: @dupondje @JasperB-TeamBlue

### DIFF
--- a/source/download/rpms_and_gpg.md
+++ b/source/download/rpms_and_gpg.md
@@ -49,24 +49,23 @@ by any mirror or website that is providing the packages. Nightly repositories wo
 For some repositories, such as repositories with stable in default configuration, yum is able to find a proper key for the repository and asks the user
 for confirmation before importing the key if the key is not already imported into the rpm database.
 
-To get the public key:
+To get the public key for oVirt 4.5:
 
-    $ gpg --recv-keys FE590CB7
-    $ gpg --list-keys --with-fingerprint FE590CB7
+    $ curl -O https://resources.ovirt.org/pub/keys/RPM-GPG-ovirt-v4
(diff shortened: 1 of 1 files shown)


dupondje (approved):

dupondje:
Thanks!

Issue #3279, opened 2026-06-09
Clarify disk removal behavior during custom snapshot restore


benyraja:
The current oVirt documentation for Previewing and Restoring Snapshots does not clearly describe the behavior when using the Custom Preview option and selecting only a subset of disks.
Specifically, it does not mention what happens to disks that are not selected during the restore/commit operation.

When performing a custom snapshot preview and commit:

If only specific disks (e.g., root/C drive) are selected for restoration,
Any additional disks that were attached to the VM but not selected in the custom restore
May be removed from the VM configuration after the commit

This can lead to unintended disk detachment and potential data loss if users are unaware of this behavior.

PR #3281, opened 2026-06-16
docs: add warning for disk removal in custom snapshot restore

benyraja:
Documentation Update

Fixes issue # [3279](https://github.com/oVirt/ovirt-site/issues/3279)

This PR improves the documentation for Using a Snapshot to Restore a Virtual Machine by adding a warning related to the Custom preview/restore workflow.

Added a [WARNING] block under the Custom preview step
Clarifies behavior when selecting only a subset of disks during snapshot restore

Problem Addressed
The current documentation does not clearly explain what happens to disks that are not selected when performing a custom snapshot restore.
In practice:

Only the selected disks are included in the final VM state after Commit
Unselected disks may be removed from the VM configuration

This can lead to unexpected disk detachment and potential data loss.

I confirm that this pull request was submitted according to the [contribution guidelines](https://github.com/oVirt/ovirt-site/blob/main/CONTRIBUTING.md): @benyraja 

This pull request needs review by: any available maintainers. @dupondje @sandrobonazzola


source/documentation/virtual_machine_management_guide/topics/Using_a_snapshot_to_restore_a_virtual_machine.adoc
```diff
@@ -22,6 +22,15 @@ The snapshot is loaded.
 . Click the btn:[Preview drop-down menu] button and select *Custom*.
 . Use the check boxes to select the *VM Configuration*, *Memory*, and disk(s) you want to restore, then click btn:[OK]. This allows you to create and restore from a customized snapshot using the configuration and disk(s) from multiple snapshots.
 +
+[WARNING]
+====
+When using the *Custom* preview option, only the selected disks are included in the restored virtual machine state.
+
+Any disks that are *not selected* during this operation may be removed from the virtual macshine configuration after clicking *Commit*. This can result in unintended data loss if additional disks are not included in the selection.
```
JasperB-TeamBlue: Typo, Virtual machine not "macshine"

benyraja: Fixed the typo

JasperB-TeamBlue (changes requested):
Image
as you now can see, the warning is in the middle of the 7th step which breaks it up. My suggestion is to drop this warning down to between the 7th and 8th step.

benyraja:
Step 7 is preview mode and the warning in the middle helps the user to follow the instructions before selecting the preview mode. if I put it between 7 and 8, then the user may have to undo the snap from preview and retry the custom option based on the warning. 

Please advise.

JasperB-TeamBlue:
> Step 7 is preview mode and the warning in the middle helps the user to follow the instructions before selecting the preview mode. if I put it between 7 and 8, then the user may have to undo the snap from preview and retry the custom option based on the warning.
> 
> Please advise.

Visually it is just best practice to not break up blocks of text. If you want to warn people before they do the action then maybe put it before step 7. 

Also your whole warning text is quite confusing and longer then it needs to be. This warning itself is already present in the UI when executing the actions you prescribe.
image
The presence of this warning in the engine itself begs the question if an extra warning on the site is actually needed or that such a lengthy explanation outweighs the quite clear 'Excluded disks will be deleted when the snapshot is committed.' warning.

benyraja:
> Visually it is just best practice to not break up blocks of text. If you want to warn people before they do the action then maybe put it before step 7.

I agree that placing the warning before Step 7 is a better option and would improve the flow of the procedure without interrupting the restore steps.

Regarding the warning itself, the message displayed in the UI is somewhat ambiguous. It does not clearly indicate whether the excluded disks are removed from the VM configuration or from the snapshot itself. The intention of the documentation warning was to provide that additional context and clarify the outcome of committing a custom restore when only a subset of disks is selected.

That said, I can simplify the warning text and move it before Step 7 so it remains concise while still highlighting the potential impact on unselected disks.

benyraja:
image

JasperB-TeamBlue (changes requested):
Looking better, would it be possible to rebase it so you have one clean commit instead of the 4 you have now?

dupondje:
@benyraja: Should be 1 commit, so both commits (no merge commit needed), needs to be squashed into 1 commit.
Also the commit is missing a 'Signed-off-by' footer. Use git commit -s to add it automatically.

